PR #2, opened 2021-07-20
Example for  Reference Coordinates Test

MarkusSteinbrecher:


pablo-mayrgundter:
Probably better as a top-level dir outside of src, and also needs a .ifc on the file, but will add myself.

PR #16, opened 2021-08-05
More on IFC props and push to prod

pablo-mayrgundter:
More on IFC props, also allow selection of root tree elts since they have interesting props.  Push to prod so we can test speed of loading with new tree props.


pablo-mayrgundter:
Forcing this in with broken merge to fix later.

PR #85, opened 2022-01-28
Improve 404 bounce logic for GHPages serving.

pablo-mayrgundter:
Permalinks now work in our forked pages:

https://pablo-mayrgundter.github.io/Share/84/103/138/160/93485

And also happened to fix Firefox, which wasn't working at all with previous approach.  Tested in Safari too.

Firefox local <-> esbuild server doesn't work.. it's mystifying.  But I found out it wasn't working before either, so I think this is a strict improvement overall.

I'm not sure until we put it up at bldrs.ai whether it will work there.


pablo-mayrgundter:
I made a mistake.. this was the merge of upstream into my fork to prepare for the bounce changes.  They're coming next.

PR #86, opened 2022-01-30
Bounce, the real one

pablo-mayrgundter:
Same demo link, but now right PR..

https://pablo-mayrgundter.github.io/Share/84/103/138/160/93485

Recap: I sent #85, but it was actually just the merge of upstream main into my fork main.. no actual changes.  I thought I was in this branch but I wasn't.  All good now I hope!


public/404.html
```diff
@@ -1 +1,39 @@
-<html><head><meta http-equiv="refresh" content="0; URL=/"></head><body></body></html>
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Single Page Apps for GitHub Pages</title>
+    <script type="text/javascript">
+      // Single Page Apps for GitHub Pages
```
OlegMoshkovich, line 7: Do we need to include this description? @pablo-mayrgundter

public/index.html
```diff
@@ -38,10 +38,38 @@
...
+          }).join('?');
+          var finalPath = l.pathname.slice(0, -1) + decoded + l.hash
+          window.history.replaceState(null, null, finalPath);
+        }
+      }(window.location))
+    </script>
+    <!-- End Single Page Apps for GitHub Pages -->
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
```
OlegMoshkovich, line 66: minor: maybe we delete this?

public/index.html
```diff
@@ -38,10 +38,38 @@
...
     <div id="root"></div>
-    <script src="./index.js"></script>
+    <script>
+      const loader = document.createElement('script');
+      const pathPrefix = window.location.pathname.startsWith('/Share') ? '/Share' : '';
```
OlegMoshkovich, line 70: nice

public/index.html
pablo-mayrgundter: I think we leave in case some env doesn't have JS enabled, it'll otherwise fail silently.

OlegMoshkovich: ok

public/404.html
pablo-mayrgundter: Yeah, it was meant as a copy/paste block from source, so keeping it intact is a nice nod to the original dev.  Also, makes it easy to compare with original later.

pablo-mayrgundter:
Ok, trimmed some of the 404 code based on your comments. Still includes the link back to the proj.

public/404.html
pablo-mayrgundter: Ok, did some trimming.  Please take another look (first PTAL!).

PR #97, opened 2022-02-09
About

OlegMoshkovich:
https://olegmoshkovich.github.io/Share/share/v/p/index.ifc

Update:
- Icons
- About Panel
- Icon Group
- Shortcut panel


pablo-mayrgundter (commented):
Main comment, there's 2 changes here.. 1) theme change, 2) large file reorg.  Should be 2 PRs, and unfortunately I'm in the middle of the eslint PR which touches a lot of these too and does similar reorgs.  I think we need to sit down and sort these together

PR #113, opened 2022-02-19
Bug project

OlegMoshkovich:


src/Components/ItemProperties.jsx
```diff
@@ -257,9 +255,10 @@ function row(d1, d2, serial) {
   if (d2 === null) {
     return (<tr key={serial}><td key={serial + '-double-data'} colSpan="2">{d1}</td></tr>)
   }
-  return (
-    <Row d1={d1} d2={d2} serial={serial} />
-  )
+  if (typeof d1 === 'object' || d2 === 'object') {
```
pablo-mayrgundter, line 258: I get that this is a fix, but it will throw away data from the props view.  Let's take another look together to see if we can fix.

PR #137, opened 2022-03-07
Issues: fix zIndex fighting with NavPanel on mobile; fit onto mobile screen as well

pablo-mayrgundter:
https://pablo-mayrgundter.github.io/Share/share/v/p/index.ifc#i:8
image


pablo-mayrgundter:
PTAL!  It's pulling comments from bldrs.ai now, so clicking through all of them won't work.. so just check the location of the dialog on desktop and mobile and assume the content will change correctly

pablo-mayrgundter:
Just added a fix for Camera on load!

PR #143, opened 2022-03-14
theme: move theme control to pop-up from bldrs logo.

pablo-mayrgundter:
Moves theme control to pop-up from bldrs logo.  Try with single-click to toggle pop-up.  Considering using a Mui Transition too, but not sure how yet.

Also moves all theme setup code to src/Theme.jsx.

Also add double-click of logo to return to home.  **NOTE** doesn't work on mobile.  Hmmm

Leaving this as draft for the moment as I have a question for you.. I'm not seeing the default theme color change for the primary text in SearchBar and NavPanel.. any ideas?

https://pablo-mayrgundter.github.io/Share


OlegMoshkovich:
the reason the tree items don't toggle the color is because those are links and seem to follow their own styling
image

OlegMoshkovich:
@pablo-mayrgundter not sure what is going with the search input base

PR #174, opened 2022-04-06
Extend search bar functionality to open IFCs hosted on GitHub.

OlegMoshkovich:
- Extend search bar functionality to open GitHub model URLs.
- Add the ability to clear search bar input
- Add a link icon to the search bar to explain the URL feature
- Dynamically change the width of the search bar in the input string is longer then the defined min width
- Add small size to buttons
- minor: adjust style of snack_bar

https://olegmoshkovich.github.io/Share/share/v/gh/Swiss-Property-AG/Portfolio/main/EISVOGEL.ifc


src/Components/Buttons.jsx
```diff
@@ -89,11 +90,12 @@ export function ControlButton({
   setIsDialogDisplayed,
   icon,
   placement='left',
+  size = 'medium',
   dialog,
 }) {
   assertDefined(title, isDialogDisplayed, setIsDialogDisplayed, icon, dialog)
   const toggleIsDialogDisplayed = () => setIsDialogDisplayed(!isDialogDisplayed)
-  const classes = useStyles(useTheme())
+  const classes = useStyles(size === 'small' ? {buttonWidth: '40px'} : {buttonWidth: '50px'})
```
pablo-mayrgundter, line 98: Shouldn't this be in a @media selector down in the styles defs?

src/Components/Buttons.jsx
```diff
@@ -120,7 +122,7 @@ export function ControlButton({
  */
 export function FormButton({title, icon, type='submit', placement='left', size='medium'}) {
   assertDefined(title, icon)
-  const classes = useStyles(useTheme())
+  const classes = useStyles(size === 'small' ? {buttonWidth: '40px'} : {buttonWidth: '50px'})
```
pablo-mayrgundter, line 125: same

src/ShareRoutes.jsx
```diff
@@ -32,6 +32,43 @@ function Forward({appPrefix}) {
 }
 
 
+/**
+ * Check if input is a valid url
+ * @param {Object} input
+ * @return {boolean} return true if url is found
+ */
+export function isValidModelURL(input) {
```
pablo-mayrgundter, line 40: Caller should be required to provide string

src/ShareRoutes.jsx
```diff
@@ -32,6 +32,43 @@ function Forward({appPrefix}) {
...
+  {/* eslint-disable-next-line */}
+  const urlRegex = '((http|https)://)(www.)?' + '[a-zA-Z0-9@:%._\\+~#?&//=]{2,256}\\.[a-z]' + '{2,6}\\b([-a-zA-Z0-9@:%._\\+~#?&//=]*)'
```
pablo-mayrgundter, line 42: Hehe, that's a crazy regex!  I would avoid using one of this complexity if there's a simpler way.  Either way please add tests for this method.

How about do the new URL parse and then access the .pathname field of it.. still a string and need a regex, but it will be much simpler.

src/Components/SearchBar.jsx
```diff
@@ -73,7 +80,26 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
...
+        size = 'small'
+        placement = 'bottom'
+        icon={<SearchIcon/>}/>
+      <TooltipToggleButton
+        title={`Type GitHUB URL to access IFCs hosted on GitHUB.
+                Click on the link icon to learn more.`}
+        size = 'small'
+        placement = 'right'
+        onClick={()=>{
+          window.open('https://github.com/bldrs-ai/Share/wiki/GitHub-model-hosting')
```
pablo-mayrgundter, line 100: What did you make of the idea to pop-up a url-entry dialog.. demo link isn't working so not sure if you did.

I'd have assumed it would be here?

src/ShareRoutes.jsx
```diff
@@ -32,6 +32,43 @@ function Forward({appPrefix}) {
...
+  }
+}
+
+
+/**
+ * construct a valid path to the GitHUB model
+ * @param {Object} input
+ * @return {string} model URL
+ */
+export function constructModelURL(input) {
```
pablo-mayrgundter, line 57: Should also be string input.

src/ShareRoutes.jsx
```diff
@@ -32,6 +32,43 @@ function Forward({appPrefix}) {
...
+  const url = new URL(input)
+  const URLParamtersArr = url.pathname.split('/')
+  const URLParameters = {
+    org: URLParamtersArr[1],
+    repo: URLParamtersArr[2],
+    branch: URLParamtersArr[4],
+    fileName: URLParamtersArr[5],
+  }
+  {/* eslint-disable-next-line */}
```
pablo-mayrgundter, line 66: please narrow to line-length if that's all it's here for

src/ShareRoutes.jsx
```diff
@@ -32,6 +32,43 @@ function Forward({appPrefix}) {
...
+  const modelPath = `/share/v/gh/${URLParameters.org}/${URLParameters.repo}/${URLParameters.branch}/${URLParameters.fileName}`
```
pablo-mayrgundter, line 67: so this was where I was wondering if moving the function to this class might help.  Can't this string prefix be shared with some of the other code here?

pablo-mayrgundter (changes requested):
The demo link isn't working for me.

Please add tests for the new code in ShareRoutes, and also the instructions page you point to still looks in progress, that right?  I was noting that I think it needs more introductory text to help orient the user.

https://github.com/bldrs-ai/Share/wiki/GitHub-model-hosting

src/ShareRoutes.jsx
OlegMoshkovich: narrowing does not work.

src/ShareRoutes.jsx
OlegMoshkovich: fixed

src/Components/SearchBar.jsx
OlegMoshkovich: check the link. 
Dialog is not so straight forward since the link icon is within search input component.

src/ShareRoutes.jsx
OlegMoshkovich: I am not sure what you mean, can you please clarify?

src/Components/Buttons.jsx
OlegMoshkovich: Since we are styling the button explicitly, MUI small prop does not work. 
I added a size prop to all of the button components, the default is medium for all of the buttons. 
If the size is small it passes a bottonWidth Prop to the styles.
image

src/ShareRoutes.jsx
OlegMoshkovich: Added some tests. PTAL

OlegMoshkovich: @pablo-mayrgundter 
Actually the new URL() give us this object:
image

so i would need to check the hostname to be valid

src/ShareRoutes.test.jsx
```diff
@@ -1,13 +1,40 @@
...
+import {isValidModelURL, constructModelPath} from './ShareRoutes'
 
 
-test('ShareRoutes', () => {
-  const {getByText} = render(mockRoutes(
-      <ShareRoutes
-        installPrefix={'/'}
-        appPrefix={'share'} /> ))
-  expect(getByText(/BLDRS/i)).toBeInTheDocument()
+// valid url format
```
pablo-mayrgundter, line 4: The first arg for test is the description, so just move the comment into the first argument.  Sometimes I'll do the name of the method if it's just one test.  If it's multiple, like here, you can move your comment in after the method name e.g. "isValidModelURL: valid format"

src/Components/SearchBar.jsx
```diff
@@ -26,8 +29,8 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
   const [inputText, setInputText] = useState('')
   const onInputChange = (event) => setInputText(event.target.value)
   const searchInputRef = useRef(null)
-  const classes = useStyles()
-
+  const inputWidth = Number(inputText.length)*11
+  const classes = useStyles({width: inputWidth})
```
pablo-mayrgundter, line 33: I feel like the 11 should be derived from the font size, as an 'em' property.  Either way, more readable to have the 11 close to the usage location, so I think 'width' here should instead be inputTextLength and then the calculation in the useStyles func.

src/Components/SearchBar.jsx
```diff
@@ -44,11 +47,14 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [searchParams])
 
-
   const onSubmit = (event) => {
     // Prevent form event bubbling and causing page reload.
     event.preventDefault()
-
+    if (isValidModelURL(inputText)) {
```
pablo-mayrgundter, line 53: Per discussion, I think we need to add more here for users who aren't quite getting it right.  I suggest one more check here:

```
if (looksLikeURL(inputText)) {
  if (isValidModelURL(inputText)) {
     ...
  } else {
     // help user
  }
```
and looksLikeURL can be something simple like inputText.startsWith('http://')

src/ShareRoutes.jsx
pablo-mayrgundter: Agreed, but I think it's still preferable to have URL do the first parsing and then separate regexes for origin and pathname.  They'll be much simpler.  Regexes in general are hard so they should be short if possible.

src/ShareRoutes.jsx
pablo-mayrgundter: Sorry, I mean using the installPrefix and appPrefix instead of hardcoding them

OlegMoshkovich: oh the share and /v/gh ? but they are passed in sharedRoutes which is a separate function

src/ShareRoutes.jsx
OlegMoshkovich: yes but this long regex is to validate the origin unless you can think of a shorter version, i goggled around and they all seemed long.
at the moment the pathname is not being validated, agree it should be.

src/Components/SearchBar.jsx
OlegMoshkovich: yes that makes sense

OlegMoshkovich:
@pablo-mayrgundter PTAL!

src/Components/SearchBar.jsx
```diff
@@ -44,10 +47,20 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [searchParams])
 
-
   const onSubmit = (event) => {
     // Prevent form event bubbling and causing page reload.
     event.preventDefault()
+    setError('')
+    // if url is typed into the search bar open the model
+    if (isURL(inputText)) {
```
OlegMoshkovich, line 55: @pablo-mayrgundter look here.

src/ShareRoutes.jsx
```diff
@@ -32,6 +33,79 @@ function Forward({appPrefix}) {
...
+
+/**
+ * ParseIfcURL parses Github repository URLs or raw content URLs
+ * @param {string} url
+ * @return {string} Structured path to the model repositore
+ */
+export const parseIfcURL = (url) => {
+  assertDefined(url)
+  const parsed = new URL(url)
+  const parts = parsed.pathname.split('/')
```
pablo-mayrgundter, line 77: So I don't think this approach is very safe.  Remember that paths can be deep.. after the repo/branch etc. the file can be simple like "file.ifc" or "a/b/c/d/e/f/g/h/i/file.ifc"

Below then, you do path: parts[5], which I think currently is a bug.  You could join the rest together.

I also proposed a regex with named capture groups in chat that looks significantly simpler than this.

src/ShareRoutes.jsx
```diff
@@ -32,6 +33,79 @@ function Forward({appPrefix}) {
...
+  }
+}
+
+
+/**
+ * ParseIfcURL parses Github repository URLs or raw content URLs
+ * @param {string} url
+ * @return {string} Structured path to the model repositore
+ */
+export const parseIfcURL = (url) => {
```
pablo-mayrgundter, line 74: I think this should probably be limited to parseIfcRepoPath or smth, not the whole URL path.  This is because L88 below has hardcoded prefix for the app installation path, which should be avoided.  Caller should pull that off if possible.

src/ShareRoutes.test.jsx
```diff
@@ -1,13 +1,93 @@
...
+
+test('parseIfcURL: test raw path', () => {
+  try {
+    const retValue = parseIfcURL('https://raw.githubusercontent.com/Swiss-Property-AG/Portfolio/main/EISVOGEL.ifc')
+    expect(retValue).toBe('/share/v/gh/Swiss-Property-AG/Portfolio/main/EISVOGEL.ifc')
+  } catch (e) {
+    expect(e.message).toBe('Invalid URL: https://raw.githubusercontent.com/Swiss-Property-AG/Portfolio/main/EISVOGEL.ifc')
+  }
+})
+
```
pablo-mayrgundter, line 93: Please add a unit test for a deep path like "https://raw.githubusercontent.com/Swiss-Property-AG/Portfolio/main/a/b/c/d/e/f/g/h/EISVOGEL.ifc"

pablo-mayrgundter (changes requested):

pablo-mayrgundter (changes requested):
Triggering another "Submit review" event to see if it will post to chat.

src/Components/SearchBar.jsx
```diff
@@ -24,10 +27,10 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
   const navigate = useNavigate()
   const [searchParams, setSearchParams] = useSearchParams()
   const [inputText, setInputText] = useState('')
+  const [error, setError] = useState('')
   const onInputChange = (event) => setInputText(event.target.value)
   const searchInputRef = useRef(null)
-  const classes = useStyles()
-
+  const classes = useStyles({inputTextLength: Number(inputText.length) * 11})
```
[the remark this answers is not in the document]
OlegMoshkovich: yes, 11 is a number of pixels per letter that gives us the right length. 10 is too short, 11 is too long.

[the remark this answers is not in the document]
OlegMoshkovich: Well,  it was clearer before, I thought, but you suggested moving it closer to the input length. 

You probably have a better idea of how to implement it in a clean way, using 'em' or whatever.
I suggest you just do it in the clearest way for you.

src/Components/SearchBar.jsx
```diff
@@ -44,10 +47,20 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [searchParams])
 
-
   const onSubmit = (event) => {
     // Prevent form event bubbling and causing page reload.
     event.preventDefault()
+    setError('')
```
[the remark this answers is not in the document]
OlegMoshkovich: I can wrap it in the if statement,
if(error.length>0){
  setError('')
}

src/Components/SearchBar.jsx
```diff
@@ -63,18 +76,50 @@ export default function SearchBar({onClickMenuCb, showNavPanel}) {
...
+            }}
+            icon={<ClearIcon/>}/>:null
+        }
+        <FormButton
+          title='search'
+          size = 'small'
+          placement = 'bottom'
+          icon={<SearchIcon/>}/>
+        <TooltipToggleButton
+          title={`Type GitHUB URL to access IFCs hosted on GitHUB.
```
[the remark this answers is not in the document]
OlegMoshkovich: yes, I agree. 
Will fix.

src/ShareRoutes.jsx
```diff
@@ -32,6 +33,79 @@ function Forward({appPrefix}) {
...
+/**
+ * Check if input is a valid url
+ * @param {Object} input
+ * @return {boolean} return true if url is found
+ */
+export function isValidModelURL(input) {
+  assertDefined(input)
+  try {
+    const url = new URL(input)
+    const parsed = new URL(url)
```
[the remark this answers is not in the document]
OlegMoshkovich: Probably.
This is the method Ogali suggested, and I was under the impression you approved of it.

src/ShareRoutes.jsx
```diff
@@ -32,6 +33,79 @@ function Forward({appPrefix}) {
...
+    if (parsed.hostname !== 'github.com' &&
+        parsed.hostname !== 'www.github.com' &&
+        parsed.hostname !== 'raw.githubusercontent.com') {
+      return false
+    }
+    return true
```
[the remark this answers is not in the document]
OlegMoshkovich: That's good. will fix.

src/ShareRoutes.jsx
```diff
@@ -32,6 +33,79 @@ function Forward({appPrefix}) {
...
+  } catch (e) {
+    return false
+  }
+}
+
+
+/**
+ * ParseIfcURL parses Github repository URLs or raw content URLs
+ * @param {string} url
+ * @return {string} Structured path to the model repositore
```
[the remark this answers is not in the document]
OlegMoshkovich: will fix

src/Components/AboutControl.jsx
```diff
@@ -119,7 +119,6 @@ function AboutContent({installPrefix}) {
       </ul>
       <Typography variant='h5' color='info'>Highlighted Projects:</Typography>
       <div className = {classes.demoContainer}>
-        {/* eslint-disable-next-line */}
```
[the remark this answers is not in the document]
OlegMoshkovich: this is ok I increased the max line length to 140.
since it is happening a lot.

package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "buildrs",
-  "version": "1.0.0-r270",
+  "name": "bldrs-ai",
```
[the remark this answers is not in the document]
OlegMoshkovich: sure

src/ShareRoutes.jsx
OlegMoshkovich: right - that makes sense

pablo-mayrgundter:
I accidentally pushed my changes to your fork&branch (a lot of that lately! 🫣) but I think we like it?  Will wait to hear back your comments

src/ShareRoutes.jsx
```diff
@@ -94,3 +73,95 @@ export default function ShareRoutes({installPrefix, appPrefix}) {
...
+
+
+/**
+ * Check if input is a url
+ * @param {input} input
+ * @return {boolean} return true if url is found
+ */
+export function looksLikeLink(input) {
+  assertDefined(input)
+  return input.includes('/') || input.endsWith('.ifc')
```
OlegMoshkovich, line 107: do you think we should look out for any other signs of URL?

src/ShareRoutes.jsx
```diff
@@ -94,3 +73,95 @@ export default function ShareRoutes({installPrefix, appPrefix}) {
...
+ * @param {Object} urlStr
+ * @return {boolean} return true if url is found
+ */
+export function trimToPath(urlStr) {
+  assertDefined(urlStr)
+  urlStr = urlStr.trim()
+  if (urlStr.startsWith('http://')) {
+    urlStr = urlStr.substring('http://'.length)
+  } else if (urlStr.startsWith('https://')) {
+    urlStr = urlStr.substring('https://'.length)
```
OlegMoshkovich, line 122: is this the same case twice?

src/ShareRoutes.jsx
```diff
@@ -94,3 +73,95 @@ export default function ShareRoutes({installPrefix, appPrefix}) {
...
+  '(?<org>[^/]+)',
+  '(?<repo>[^/]+)',
+  '(?:(?<isBlob>blob)/)?(?<branch>[^/]+)',
+  '(?<file>.+)',
+]
+
+
+/** Matches strings like '/org/repo/branch/dir1/dir2/file.ifc' */
+const re = new RegExp(`^(?:.*com)?/${pathParts.join('/')}$`)
+
```
OlegMoshkovich, line 143: this is cool!

src/ShareRoutes.test.jsx
```diff
@@ -1,13 +1,59 @@
...
+  {s: '/org/repo/branch/a/b/c/file.ifc', out: pathAbc},
+  {s: 'www.google.com', err: 'Not a url with path: www.google.com'},
+]
+
+
+describe('ShareRoutes', () => {
+  test('looksLikeLink', () => {
+    tests.forEach((pair) => {
+      expect(looksLikeLink(pair.s)).toBe(pair.out !== undefined)
+    })
```
OlegMoshkovich, line 32: who is this test cathing the invalid URLS?

src/ShareRoutes.jsx
pablo-mayrgundter: I thought through a bunch of cases and came up with this, along with test cases.

src/ShareRoutes.jsx
pablo-mayrgundter: I even went simpler!  Now the caller needs to trim the string to just the path first, and then call the regex.  So in the end it was possible to split the work into a few methods.

src/ShareRoutes.test.jsx
pablo-mayrgundter: the function under test here just returns a boolean and doesn't throw exception.

pablo-mayrgundter:
Heya, I'm leaving for the weekend tomorrow so in the interest of time going to merge.

You said you were good and just had those comments.  I did update to handle more types of links in looksLikeLink and then answered the other questions.  So I think everything is addressed.

PR #178, opened 2022-04-13
Add Husky-powered pre-commit hook

oogali:
PR to introduce Husky as facility for installing and executing pre-commit hooks within the project.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "serve": "yarn build && node config/serve.js",
     "lint": "yarn eslint `find src -name '*.jsx'` `find src -name '*.js'`",
     "test": "jest",
-    "precommit": "yarn test && yarn lint"
+    "precommit": "yarn test && yarn lint",
+    "prepare": "husky install"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -14,7 +14,8 @@
     "serve": "yarn build && node config/serve.js",
     "lint": "yarn eslint `find src -name '*.jsx'` `find src -name '*.js'`",
     "test": "jest",
-    "precommit": "yarn test && yarn lint"
+    "precommit": "yarn test && yarn lint",
+    "prepare": "husky install"
```
pablo-mayrgundter, line 18: Can we call this smth like husky-init

pablo-mayrgundter (approved):
Just a little name change.  Great!

package.json
oogali: The lifecycle script must be called `prepare` -- that's a special part of the Yarn lifecycle that is automatically triggered when certain things are run.

https://docs.npmjs.com/cli/v8/using-npm/scripts

PR #179, opened 2022-04-13
<TooltipIconButton> Tests

oogali:
This PR adds the supporting infrastructure for manual mocking of components, and two tests for <TooltipIconButton>.

Small, optional bonus of converting from ES5 function to ES6 fat arrow function.


src/Components/Buttons.test.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+import {fireEvent, render} from '@testing-library/react'
+import {TooltipIconButton} from './Buttons'
+import QuestionIcon from '../assets/2D_Icons/Question.svg'
+import {MockComponent} from '../__mocks__/MockComponent'
+
+describe('<TooltipIconButton />', () => {
```
pablo-mayrgundter, line 7: Please 2 spaces between import and first code block.

src/__mocks__/MockComponent.jsx
```diff
@@ -0,0 +1,23 @@
+import useTheme from '../Theme'
+import {ThemeProvider} from '@mui/material/styles'
+import React, {createContext} from 'react'
+
+const ColorModeContext = createContext({
```
pablo-mayrgundter, line 5: Also here 2 empty lines.

Prefer MockComponent to be first code block since it matches filename, with ColorModeContext below.

src/__mocks__/MockComponent.jsx
```diff
@@ -0,0 +1,23 @@
...
+  toggleColorMode: () => {},
+})
+
+/**
+ * @param {Object} children React component(s)
+ * @return {Object} React component
+ */
+export const MockComponent = ({children}) => {
```
pablo-mayrgundter, line 13: I'm unfamiliar with this semantics.  Isn't it @param props and it has props.children property?

src/Components/Buttons.jsx
```diff
@@ -12,25 +12,27 @@ import {assertDefined} from '../utils/assert'
  * @param {Object} icon
  * @param {string} placement
  * @param {string} size
+ * @param {string} dataTestId Internal attribute for component testing
  * @return {Object} React component
  */
-export function TooltipIconButton({
+export const TooltipIconButton = ({
```
pablo-mayrgundter, line 18: So we were using more modern syntax but there was something about function that worked better.  I forget what.  Is there a reason to prefer const here besides shorter?

src/__mocks__/MockComponent.jsx
oogali: In ES6, you can destructure object properties into function parameters: among other things, it allows the scope of the function to get at the properties of the argument without the chore of individually accessing them.

Here's a better summary of my use case and other helpful use cases: https://simonsmith.io/destructuring-objects-as-function-parameters-in-es6

src/Components/Buttons.test.jsx
oogali: Corrected in latest push.

src/__mocks__/MockComponent.jsx
oogali: Corrected in ea796559bdadd91c4cfa3575fbccde1ef09c17ac.

src/__mocks__/MockComponent.jsx
pablo-mayrgundter: Ah of course!  I'm still blind to this syntax.  Sorry, thanks for the explanation!

src/Components/Buttons.jsx
pablo-mayrgundter: I suggest reverting this to function as it's more consistent with the rest of our code.  I think the rule we're using is "top-level exported functions should use function keyword".  I'll add that to styleguide.

pablo-mayrgundter: https://github.com/bldrs-ai/Share/wiki/Style-Guide

PR #183, opened 2022-04-22
Auth0 integration

pablo-mayrgundter:
Hi O,

Can you review this one?  I have it mostly working, but still having trouble with the staging environment.


shafy:
Hey @pablo-mayrgundter I had a look at this today. The problem is that we need to call GitHub with the GitHub token, not with the Auth0 access token. Since this is a SPA, [using the Auth0 Management API is a pain in the ass](https://auth0.com/docs/authenticate/identity-providers/calling-an-external-idp-api) because the token required for that shouldn't be exposed to the client.

I'm doing some more research if there's a way to get the GitHub token (in Auth0 lingo called Identity Provider Token). Keep you posted.

shafy:
@pablo-mayrgundter I've now tested this with getting my GitHub Identity Token manually using the [Auth0 Management API](https://auth0.com/docs/api/management/v2#!/Users/get_users_by_id) Explorer, and providing that token in the code like so I can successfully load an IFC file from my private repo:

```js
const token = 'gho_abc1abc2abc3xyz234'

// we need to set the header like so
viewer.IFC.loader.requestHeader = {
  'Authorization': `token ${token}`,
  'Accept': 'application/vnd.github.v3.raw',
}

// we need to call the github api, URL is different from getting raw file directly from a public repo
const filepath = 'https://api.github.com/repos/shafy/ifc-test/contents/haus.ifc'

// load model
const model = await viewer.IFC.loadIfcUrl(filepath, /* further args */)

```

Unfortunately, as I indicated in my earlier comment, to get the GitHub Identity Token from Auth0 programmatically, we need to call the Auth0 Management API. We can only do this from a service running in the backend, because to call the Management API we need to provide a special access token that is sensitive and can't be exposed to the client.

You can read more here:
- https://auth0.com/docs/secure/tokens/access-tokens/management-api-access-tokens
- https://auth0.com/docs/secure/tokens/access-tokens/identity-provider-access-tokens
- https://auth0.com/docs/secure/tokens/access-tokens/get-management-api-tokens-for-single-page-applications

As I see it, we have two possibilities which also depend a bit on the further strategy of BLDRS:
1. If we are going to go with a full-stack app (not only static page like today), it's probably a good idea to migrate over sooner than later (which framework / libraries / hosting we want to use needs to be decided).
2. If you think almost a static page will cover all our needs (except this one) in the foreseeable future, then it might make sense to use a service like Firebase (or better, [Supabase](https://supabase.com), which is open-source and based on Postgres, not a weird document database like Firebase). Supabase also provides an auth service, so it would probably make sense to use that instead of Auth0 in this case [0].

If the backend is going to get gradually more imporant (which I assume it will), I would go with a proper backend architecture and not a cloud function based one (so, not Firebase or Supabase). If you want to stay within the JS/Node ecosystem, there are great frameworks or libraries we could use.


0: I've used Auth0 for a couple of projects, and was never entirely happy. Their docs are lacking, and their SDKs are not intuitive.

---

If you want to reproduce getting your GitHub Access Token manually to load the model, here are the steps:
1. Get our test token in Auth0: Applications -> APIs -> Auth0 Management API -> API Explorer
2. Copy and paste it in the [Auth0 Management API Explorer](https://auth0.com/docs/api/management/v2), top left "Set API token"
3. Make sure you have a user account that's signed up with GitHub, and from the Auth0 dashboard grab your user id (it will be something like `github|12345678`
4. Back in the Managemenet API Explorer, run the ["Get User" endpoint](https://auth0.com/docs/api/management/v2#!/Users/get_users_by_id) and find your GitHub token in the `identities` array of the response
5. Adapt your code like shown in the snippet above, pasting your GitHub token instead of the example one
6. Adapt the `filepath` variable (make sure it's a private GitHub owned by you)

Adrian62D:
@pablo-mayrgundter Is it possible to access private github repo's with this new feature?

pablo-mayrgundter:
@Adrian62D Sorry, I missed this.  Yeah, the goal is to access private repos from bldrs.ai using third-party auth.  We're having trouble getting this to work though.

The workaround is to install Bldrs in the Private repo.  Serving from the private repo's GitHub Pages instance does allow viewing private models.

shafy:
@pablo-mayrgundter let me know if there's anything I can help with regarding this, having already put it some thought a while ago =)

pablo-mayrgundter:
Ah wow, thanks Can!  Will keep that in mind :)

PR #191, opened 2022-04-29
Add Gitpod and small fixes

shafy:
This PR adds a simple [Gitpod](gitpod.io) (cloud based ephemeral dev environments) config that enables us to spin this project up on Gitpod with one click. It simplifies getting up and running, especially for new contributors, and makes PR reviews more efficient. You can give it a try by opening this PR in Gitpod (prepend `https://gitpod.io/#`, [like this](https://gitpod.io/#https://github.com/bldrs-ai/Share/pull/191)).

While trying to get the project running, it threw two errors which I fixed. I'm not sure why, maybe related to my node version (v16)? The first error was due to a missing `assert {type: 'json'}` when importing `package.json` in the `version.mjs` file. Apparently, the `assert` statement is mandatory with newer version of Node. I also added the Babel ESLint parser, because the standard ESLint parser doesn't support the `assert` statement yet. When/if adding TypeScript later, we'll need the Babel ESLint parser anyways.
The second error was due to the `server.listen()` method trying to start the server twice when passing a `port` argument. I have no idea why this happens on my setup and not on yours. But removing the `port` argument solves the issue and it still works.

Let me know if I missed something regarding these errors?

Another question: Why do we commit the output of the build. Is it because of GitHub pages?


pablo-mayrgundter:
Wow, I love this idea.

Ok, it's been a year or so since I tried gitpod.. I clicked your link and have a session running with your PR.  TOO COOL!

I see it builds and runs!  TOO COOL!

However!  I click on port 8080 and it does bring me to the running instance, but there's no content. I suspect this is because we're using react router to load content based on URL.  But i'm not seeing any errors in console, which I would have expected. 🤔

What's a good way to debug this?  It'll probably take some playing with public/index.html and src/{Base,Share}Routes.jsx

image

shafy:
@pablo-mayrgundter I just tried to reproduce it and you're right. Good catch. It works if you reload it, though. See this screencast:

https://user-images.githubusercontent.com/1881676/166064569-cb2a9078-037b-4466-9dee-ff8507d828ca.mov

I'm also thinking it has something to do with the router. It's late here in Berlin, but I'll have a go at it over the weekend. Keep you posted.

shafy:
@pablo-mayrgundter the issue was that when Gitpod opened the URL it had a `document.referrer` and this led to a incorrect navigation/forward flow in this part of `BaseRoutes.jsx`:


```javascript
if (referrer) {
  const ref = new URL(referrer + window.location.hash)
  if (ref.pathname.length > 1) {
    navigate(ref)
   }
} else if (location.pathname === installPrefix ||
          location.pathname === (installPrefix + '/')) {
  debug().log('BaseRoutes#useEffect[], forwarding to: ', installPrefix + '/share')
  navigate(installPrefix + '/share')
}
```

The call to `navigate()` isn't triggered, because `ref.pathname.length` is equal to `1`. But even changing that condition and trigering `navigate()` doesn't solve the issue, because the navigation doesn't change the path and therefore `useEffect()` is not called again.

Solution:
I've now removed the `referrer` condition (see [97c1b0d](https://github.com/bldrs-ai/Share/pull/191/commits/97c1b0de50afac9f4059152e72d4ce24b0e86944)), because it wasn't clear to me what the intention of that logic is. Let me know if this breaks something (or why we have that in the first place). We could solve it alternatively by leaving the code as is, but wrapping the `BaseRoutes` component in `withRouter()` so that `useEffect()` is triggered, even if there are no prop changes ([see here for the explanation](https://stackoverflow.com/questions/57986395/react-useeffect-is-not-triggering-on-route-change)). We would also need to change `ref.pathname.length > 1` to `ref.pathname.length > 0`.

Let me know how you want to proceed.

pablo-mayrgundter:
Hi Can, sorry I'm slow on this.. going to be more delayed while I'm on vacation.  I'll try to get to it over the next week.

pablo-mayrgundter:
Hi Can, could you merge from main again? I'll take another look

shafy:
Done.

As a side not, we need to do something about the current GitHub Pages build. Checking in build outputs to version control is a pain in the ass, and the missing parity on how routing is handled on GitHub Pages vs. local environment is not great. My feeling is that we will need to move away from a SPA architecture anyways, so that would of course then resolve these issues also. My advice is to do this sooner rather than later, since it keeps building up tech debt. Happy to discuss if you want my input :-)

pablo-mayrgundter:
w00t!  Works for me 😍

PR #194, opened 2022-05-04
Specify type assertion for JSON import

oogali:


oogali:
I think it makes sense to keep this separate -- we can get this fix in without requiring us to take an immediate position on Gitpod yet.

oogali:
@shafy In my branch, I did not add that module so I'm curious why I don't receive any errors regarding `assert`...

image

oogali:
@shafy Never mind, I see that it's not being tested... I'll add it to the lint inputs and incorporate your module changes too.

shafy:
Looks great to me!

PR #211, opened 2022-05-30
Use cookies import

oogali:
PR to use the upstream cookies.js library rather than repeating code via copy-and-paste.


oogali:
The unit tests I was referring to are in the upstream project:
* https://github.com/pablo-mayrgundter/cookies.js

package.json
```diff
@@ -72,7 +72,7 @@
       "^.+\\.svg$": "<rootDir>/svgTransform.js"
     },
     "transformIgnorePatterns": [
-      "/node_modules/(?!(web-ifc|web-ifc-three|web-ifc-viewer|three)/)"
+      "/node_modules/(?!(@pablo-mayrgundter\/cookies.js|web-ifc|web-ifc-three|web-ifc-viewer|three)/)"
```
pablo-mayrgundter, line 75: Why is this needed?

src/privacy/Privacy.js
```diff
@@ -1,7 +1,8 @@
 import {
   getCookie as getCookiePrivate,
-  getCookieBoolean as getCookieBooleanPrivate,
```
pablo-mayrgundter, line 3: I moved these over to the library and checked.. yeah, the getCookieBoolean has the type check (as well as the value check), so I think we should use.

src/privacy/Privacy.js
```diff
@@ -14,7 +15,7 @@ export {recordEvent, isAnalyticsAllowed, setIsAnalyticsAllowed}
  * @param {string} component
  * @param {string} name
  * @param {string} defaultValue
- * @return {boolean} value of the setting
+ * @return {string} value of the setting
```
pablo-mayrgundter, line 18: Good catch.

pablo-mayrgundter (changes requested):
All looks good, but for me to run with node18 needed to remove bldrs/package.json#canvas.  Also removed gl.  See how goes for you

package.json
oogali: Without this, Jest expects the module to be already transformed into a CommonJS artifact. Since the packages listed in this regex are ES6 modules, Jest errors out on the ES6 import/export syntax.

This directive tells Jest to transpile these matching modules prior to executing the tests.

https://jestjs.io/docs/configuration#transformignorepatterns-arraystring

OlegMoshkovich:
@oogali @pablo-mayrgundter 

Is this ready to merge?

pablo-mayrgundter:
Hi O, could you take another pass at this or revert?

Issue #198, opened 2022-05-12
Setup: Failure to build & serve on Win11


Adrian62D:
**Platform:**
- Windows 11, 22000.675
- node v16.15.0
- npm v8.5.5
- Firefox v100.0

**Steps to reproduce:**
1. Follow the [setup instructions](https://github.com/bldrs-ai/Share) i.e.:
2. Open a command prompt
3. `yarn install`
4. `yarn serve`

**Expected result:**
- Browser opens http://localhost:8080/
- bldrs is shown

**Actual result:**
```
>yarn serve
yarn run v1.22.18
warning package.json: No license field
$ yarn build && node config/serve.js
warning package.json: No license field
$ yarn copy-web-asm && yarn clean && yarn new-version && node config/build.js
warning package.json: No license field
$ cp node_modules/web-ifc/*.wasm public/static/js
'cp' is not recognized as an internal or external command,
operable program or batch file.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**Details:**
- package.json appears to contain platform-specific shell commands
- When patching those commands (workaround) the `yarn serve` command succeeds. However, http://localhost:8080/ shows a directory listing with 4 files (docs/) instead of loading bldrs:

![image](https://user-images.githubusercontent.com/104829208/168153009-fd1a56ea-60d8-44c2-8e7a-84046852064c.png)

Thanks for your support.

BR,
Adrian

Adrian62D:
Hi all

After applying further changes to package.json I got it to launch properly in the browser. Great.

For sake of completeness, the current state of the windows-workaround below:

```
  "scripts": {
    "clean": "rd /s /q docs & xcopy public\\ docs\\ /E /H /Y",
    "new-version": "node --experimental-json-modules src/utils/version.mjs > package.json.new && move package.json.new package.json",
    "copy-web-asm": "xcopy node_modules\\web-ifc\\*.wasm public\\static\\js /E /H /Y",
    "build": "yarn copy-web-asm && yarn clean && yarn new-version && node config/build.js",
    "serve": "yarn build && node config/serve.js",
    "lint": "yarn eslint `find src -name '*.jsx'` `find src -name '*.[m]js'`",
[...]

PR #213, opened 2022-06-01
Use shx for issues#198.

pablo-mayrgundter:
Here's a mostly fix for #198.  I've tested it for the affected commands.  No solution for lint yet.

Ogali, if you could add your ideas for future improvement to the issue we can leave it open.

Adrian, if you're reading along, I'd appreciate testing on your side!


oogali:
@pablo-mayrgundter Lint fix is in #214.

pablo-mayrgundter:
@oogali ptal on your windows machine

PR #216, opened 2022-06-04
Draft: Notes

OlegMoshkovich:
Draft - requires review + discussion.

- Add zustand Store
- Refactor SideDrawer and connected it to the Store
- Refactor Item Panel - pull mode and selectedElement from the Store
- Add Issue Card
- Add CommentsPanel
- Add logic to pull all comments and issues from GitHub

https://olegmoshkovich.github.io/Share/share/v/p/index.ifc#c:-111.37,14.94,90.63,-43.48,15.73,-4.34


src/Components/IssuesControl.jsx
```diff
@@ -1,23 +1,274 @@
...
+ */
+export function CommentPanelAll() {
+  const classes = useStyles()
+  const selectedCommentId = useStore((state) => state.selectedCommentId)
+  const issuesStore = useStore((state) => state.issues)
+  const setIssuesStore = useStore((state) => state.setIssues)
+  const replies = useStore((state) => state.replies)
+  const setReplies = useStore((state) => state.setReplies)
+  const filteredComment = selectedCommentId ? issuesStore.filter((issue) => issue.id ===selectedCommentId)[0] :null
+
```
OlegMoshkovich, line 35: Logic For fetching issues.
@pablo-mayrgundter please take a look

src/Components/IssuesControl.jsx
```diff
@@ -1,23 +1,274 @@
...
+              imageURL: imageURL,
+            },
+        )
+      })
+      setIssuesStore(issuesArr)
+    }
+    fetchIssues()
+  }, [setIssuesStore])
+
+
```
OlegMoshkovich, line 80: Logic For fetching comments.
@pablo-mayrgundter please take a look

src/Components/IssueCard.jsx
```diff
@@ -0,0 +1,274 @@
...
+          <div className = {classes.username}>{date.split('T')[0]}</div>
+        </div>
+        <div className = {classes.titleRightContainer}>
+          {!selected &&
+          <div className = {classes.select}>
+            <TooltipIconButton
+              title={selected ? 'Back to the list':'Select Comment'}
+              size = 'small'
+              placement = 'bottom'
+              onClick = {() => {
```
OlegMoshkovich, line 64: TODO: When Note is selected need to change the URL.

OlegMoshkovich (commented):
@pablo-mayrgundter please take a look at the draft of the notes.

pablo-mayrgundter:
Looking cool!

1) UX feels good and a major improvement from current comments!  Once this code is ready, I think it'll be good for prod.  I have some nits below to fix before deploy.

Usage It's still a little obscure to me what I'm doing if I'm looking through top-level Topics or opening a Topic to go into its comments.  I'm not even sure those are the right terms.  Curious to discuss with you and feel out what might make it clearer.  But this is fine for now, to launch and iterate.

UX nits:
- Properties is too small (10%?) when it's shared with Notes.  Maybe 50/50?
- Buttons seem to have lost their open/close visual state.  In prod, when I open properties, its button changes to a big X.  Here I'm not seeing any change.  I think X is probably not right anymore since multiple buttons can be active.  How about using the surrounding background disc again?  That leaves the icon visible to remind you what it's for, but also it's clear that it's active and can be made inactive.
- Paging buttons for left/right ("<" and ">") should grey out when there's no more left or right to go to.


2) Please add tests for the new components.

3) I'll respond to code issues inline...

pablo-mayrgundter (changes requested):
Ok, going to stop mid-stream here.

Main comment for code is that the style code seems to be large and getting complex.  Please take a refactoring pass where you push all style into makeStyles (no inline styles in components unless there's an exceptional circumstance) and also push as much as possible into the theme so we don't replicate all of this in every component:

- There should be only a handful of colors in our palette, all defined in the theme and referenced elsewhere.
- Same for Typography with a half-dozen types (h1, h2, ..., p, i, code, etc.)
- Margin and padding are taking up a ton of space in these files.  I have a note about how to condense these inline above.  I think most of this can be moved to padding around Typography, so we can reuse heavily.  For the rest, maybe start defining common use gutters, like ListGutter or something for long lists.
- Shouldn't cursor types be global for different button types?  I feel like there's only a handful of these.
-

OlegMoshkovich:
Yes agree. 
Notes/Topics is too general, Issues/Comments seems too specific.

OlegMoshkovich:
Yes, agree. 
Notes/Topics is too general, Issues/Comments seems too specific.

package.json
```diff
@@ -64,7 +65,8 @@
     "verbose": false,
     "testEnvironment": "jsdom",
     "testPathIgnorePatterns": [
-      "src/Share.test.js"
+      "src/Share.test.js",
+      "ItemProperties.test.js"
```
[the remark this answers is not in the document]
pablo-mayrgundter: Found this on their discord: https://github.com/pmndrs/zustand/issues/271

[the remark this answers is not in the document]
pablo-mayrgundter: I think the first comment in the first link should be fine for us for now. the whole thread is about reducing the boilerplate of that approach, which is not yet a problem for us.

pablo-mayrgundter:
Ping.. how goes?

PR #218, opened 2022-06-14
WIP: Github LFS Support

oogali:


src/ShareRoutes.test.jsx
```diff
@@ -15,10 +15,10 @@ const tests = [
   {s: 'http://github.com/org/repo/blob/branch/a/b/c/file.ifc', out: pathBlobAbc},
   {s: 'https://www.github.com/org/repo/blob/branch/file.ifc', out: pathBlob},
   {s: 'https://github.com/org/repo/blob/branch/file.ifc', out: pathBlob},
-  {s: 'github.com/org/repo/blob/branch/file.ifc', out: pathBlob},
```
pablo-mayrgundter, line 18: The use case here is copying text from apps/sites that hide the protocol.  It's a nice-to-have

pablo-mayrgundter: I don't know.  I feel like server will upgrade to ssl if it wants and we don't have any private data per se, so http for me

src/utils/urls.js
```diff
@@ -0,0 +1,45 @@
+/**
+ * Gets the final URL needed to load an IFC model
```
pablo-mayrgundter, line 2: So am I understanding correctly that large files are hosted on a separate server, and this code detects the size difference and redirects the query to that server?

oogali: This code grabs the first 64 bytes of the file.

If that initial content contains a LFS signature, then it'll download from the Github LFS servers.

pablo-mayrgundter (approved):
Ok, LGTM!  Please extend the comment to describe the issue and mechanism a bit more in the util comment.

PR #231, opened 2022-06-19
Zustand ifc slice 

OlegMoshkovich:
- Break up store into UISlice and IFCSlice
- Store Viewer in the store IFCSlice
- Store Model in the store IFCSlice
- Store Selected Element in the store IFCSlice


src/Containers/CadView.jsx
```diff
@@ -67,8 +66,12 @@ export default function CadView({
   const isItemPanelOpenState = {value: isItemPanelOpen, set: setIsItemPanelOpen}
   const [isLoading, setIsLoading] = useState(false)
   const [model, setModel] = useState(null)
-  const snackMessage = useStore((state) => state.snackMessage)
+  const setViewerStore = useStore((state) => state.setViewerStore)
+  const setModelStore = useStore((state) => state.setModelStore)
```
pablo-mayrgundter, line 70: Seems like we'll have a bunch of variable like this that have react + zustand refs.  The current name reads a little backwards to me.. does setX and storeX sound good to you?

pablo-mayrgundter (approved):
LGTM.  One optional idea on naming...

PR #241, opened 2022-06-25
Use React Context API to avoid prop drilling

oogali:
Prop drilling[1] is the act of passing a property through multiple layers of components when the intermediate layers don't necessarily need said property.

Rather than passing the install and application prefix properties through each layer repeatedly because something downstream might need it, this has been refactored to use the React Context API which allows consumers to select which property they need without the use of prop drilling.

There's probably room for improvement here by removing the `appPrefix` member and replacing it with something that is calculated from `installPrefix` when needed, but that can be evaluated at a separate time.

1: https://www.geeksforgeeks.org/what-is-prop-drilling-and-how-to-avoid-it/

### DIFF
--- a/src/BaseRoutes.jsx
+++ b/src/BaseRoutes.jsx
@@ -28,13 +28,13 @@ import {usePaths} from './Paths'
 export default function BaseRoutes({testElt = null}) {
   const location = useLocation()
   const navigate = useNavigate()
-  const {installPrefix} = usePaths()
+  const {installPrefix, appPrefix} = usePaths()
 
   useEffect(() => {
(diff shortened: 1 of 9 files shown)


OlegMoshkovich:
I just connected zustand for the same purpose.

https://github.com/pmndrs/zustand

pablo-mayrgundter:
Hey Ogali, I thought you decided against this since we're using Zustand.  Why do you think we still need it?

OlegMoshkovich:
@oogali this branch is conflicting with the notes code - I assume.

PR #255, opened 2022-07-10
Stronger syntax and style enforce via ESLint

oogali:
PR to enable additional ESLint rules to improve code quality (and maintain it at its increased level).

The new rules are being enabled as warnings during this introductory period to avoid scratching the itch of doing one blast of `eslint --fix` thereby polluting Git history and decreasing the value provided by `git blame`.

Over time, as the existing code is refactored, replaced, or otherwise modified, the style and syntax violations should be addressed if they fall into the scope of what a developer is modifying.


pablo-mayrgundter (approved):
Nice idea to start as warning for new rules.

PR #292, opened 2022-07-28
storybook: add SB to serving and use esbuild for it.

pablo-mayrgundter:
This includes new build rules and slight change in how we build.

Storybook can now be accessed at our Netlify and bldrs.ai by appending /sb to the host. I'll add a link when this PR builds.

Usually yarn serve will build, but it wiped out the existing serving directory and rebuilt it. I tried this with storybook, but SB takes a long time to build and it doesn't hot serve.

Documented the updated package scripts here:
https://github.com/bldrs-ai/Share/wiki/Dev:-Guide#package-scripts


pablo-mayrgundter:
Not sure why the netlify link goes to /sb now.. i'll try to fix that up.

pablo-mayrgundter:
Also, a follow-on for this is maybe add /docs to our .gitignore so we are sure we stop checking it in.

OlegMoshkovich (approved):

PR #301, opened 2022-08-02
Design sprint 1

OlegMoshkovich:
Draft pull request to test the redesigned layout +. components.


pablo-mayrgundter:
Looks really interesting!

A few things aren't working.. .  I was expecting more iterations on the components in SB too.  Not doing that?  I suggest leaving as a draft and integrating things as you get them working in SB.

Are unit tests passing?  These are some major breakages:

- Properties appears to be broken.
- Looks like the new open dialog isn't functional at all?

Design:
- Did you consider the proposal for Open file icon in the Search bar?
- Did you consider the proposal to use just the "b" for the bottom-left logo?

PR #312, opened 2022-08-03
Fix #307: Notes should match current model.

### LINKED ISSUE #307: Notes should match current model
In notes, we currently show all issues in the repo, independent of what model you're viewing.

I propose we show only issues somehow related or tagged to the currently displayed model.

This also suggests that if we could display a whole repo's set of models (i.e. for Portfolio, or directories of models per Greg's feedback) then we'd display all issues for the directory and its subdirectories.

pablo-mayrgundter:
Fix for #307.


pablo-mayrgundter:
Ok, finally starting to work.  Search queries return results differently than collection fetching.

https://deploy-preview-312--bldrs-share.netlify.app/share/v/gh/Swiss-Property-AG/Portfolio/main/KNIK.ifc

The query is currently for filepath, which here is '/KNIK.ifc', but seems like it should be more like '/Swiss-Property-AG/Portfolio/main/KNIK.ifc' or maybe more.  Dunno, need to think on it.

Issue #325, opened 2022-08-10
upgrade: @mui/styled engine deprecated/broken with react 18


pablo-mayrgundter:
upgrade of "@mui/material": "^5.9.2" to "@mui/material": "^5.10.0" breaks with theme passing in makeStyles.  Digging into it a bit it looks like styled is deprecated but better yet "not compatible with React 18".

OlegMoshkovich:
@pablo-mayrgundter since we moved all of the styles into sx - can we upgrade MUI now?

pablo-mayrgundter:
https://github.com/bldrs-ai/Share/pull/338

PR #327, opened 2022-08-10
Upgrade outdated npms.  Found new break in @mui/material for version 5.9.2+ (#325).

pablo-mayrgundter:
Based off #324.


OlegMoshkovich (approved):
Has Conflicts - I think it is due to the order of PRs we discussed yesterday.

Issue #318, opened 2022-08-05
Bug: Element selection only works if tree view is open


MarkusSteinbrecher:
I created notes for this project: https://bldrs.ai/share/v/gh/Swiss-Property-AG/Seestrasse-Public/main/SEESTRASSE.ifc/86/109/136/28708/41077

On note Nr3, the slab on the ground floor is selected and included in the permalink.

- if the tree view is selected, the element is selected already on the first note
- if the tree view is hidden, the element is not selected at all

Does element selection require the tree view to be open?

MarkusSteinbrecher:
also: if I select any element (e.g. the roof) while I have the notes open; then clear selection and skip to the next note, the roof is selected again.

OlegMoshkovich:
Hello.

OlegMoshkovich:
This is a general bug that has been there since the beginning, not related to Notes.

pablo-mayrgundter:
I'm wondering if this is still happening after #328?  I can now click around on different comments' cameras and not reset what element is selected.  This is probably a problem itself, but not seeing any interaction with tree being open or not.

MarkusSteinbrecher:
it seems to be working fine now, I can select elements without the tree

pablo-mayrgundter:
Cool.  Ok, opening different issue for that.. it's simpler too.

pablo-mayrgundter:
Hey @MarkusSteinbrecher could you review and close if fixed now.

MarkusSteinbrecher:
it is still the same behavior as I commented above: "it seems to be working fine now, I can select elements without the tree. What is still there is this: Open notes view and select an element; deselect the element with the "clear selection" item; go to the next note and the element is selected again."

Selection works if tree view is closed. But if you select and deselect an element and skip through notes, the element is selected again. If this issue was just about the tree view, I guess it can be closed and we can create another issue.

PR #329, opened 2022-08-11
Hide nav panel by toggling css display property.

OlegMoshkovich:
This PR resolves [#318](https://github.com/bldrs-ai/Share/issues/318).
For the Property panel to display the properties of an element, the tree needs to be loaded into the scene.
Changed the tree toggle logic to control the CSS display property as opposed to subtracting the navpanel component from the scene.

### DIFF
--- a/src/Components/NavPanel.jsx
+++ b/src/Components/NavPanel.jsx
@@ -53,6 +53,7 @@ export default function NavPanel({
   onElementSelect,
   setExpandedElements,
   pathPrefix,
+  displayNavPanel,
 }) {
   assertDefined(...arguments)
 
@@ -73,7 +74,7 @@ export default function NavPanel({
   // eslint-disable-next-line
   }, [location])
 
-  const classes = useStyles()
+  const classes = useStyles({displayNavPanel: displayNavPanel})
   // TODO(pablo): the defaultExpanded array can contain bogus IDs with
   // no error.  Not sure of a better way to pre-open the first few
   // nodes besides hardcoding.
@@ -111,6 +112,7 @@ export default function NavPanel({
 
 const useStyles = makeStyles({
   root: {
+    'display': (props) => props.displayNavPanel ? 'block' : 'none',
     'position': 'absolute',
     'top': '94px',
     'left': '20px',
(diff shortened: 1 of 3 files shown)


pablo-mayrgundter:
One minor regression seems to be that selecting an element by double-clicking it no longer opens the tree.  Can you fix that while still fixing the bug?

OlegMoshkovich:
Yes, I can address it.
However, from the UX perspective sometimes is it annoying when the tree opens when the element is selected, especially on Mobile. Most of the time I don't care about where the element is in the hierarchy, I am only concerned with the properties.
What are your thoughts?

OlegMoshkovich:
@pablo-mayrgundter left a comment earlier. Please reply.

pablo-mayrgundter:
Agreed on mobile.  Maybe open automatically on desktop, but leave alone on mobile (i.e. don't change or clear)?

OlegMoshkovich:
ok put it back the way it was. upon load the tree is expanded.

pablo-mayrgundter:
Confused now.. looks like #318 was resolved.  Maybe take a look at this after merge and see if it still makes sense?  I think show/hide with CSS is preferable sometimes, but also tree can be heavy to load so worth keeping it hidden and not part of DOM e.g. on mobile if that's going to be frequent.

OlegMoshkovich:
this is what happens when the tree is hidden:
image
the IFC property is specified as an object


when the tree is shown
image
the right ifc element type is read, along with other properties

that is addressed by this PR

OlegMoshkovich:
in this pr even if the tree is hidden the right properties are read
image

OlegMoshkovich:
@pablo-mayrgundter does it make sense?

pablo-mayrgundter:
I'm not seeing the bug in production anymore.

OlegMoshkovich:
step to reproduce the big in producetion:
- go to https://bldrs.ai/share/v/gh/OlegMoshkovich/Logo/main/IFC_STUDY.ifc/103/25873/113/122/9585/10485
- hide the tree
- open the properties panel
- you will see the 
image
- show the tree
and you see
image

pablo-mayrgundter:
to recap discussion, this bug is a little more complex.  Clicking a link with a selected element doesn't repro this, but starting from scratch on bldrs.ai logo does.  I think this is actually a bug in the handling of selectedElement react vars

PR #333, opened 2022-08-16
Fix property set scrolling

OlegMoshkovich:
- This pr addresses scrolling for the psets, previously the psets in the expanded view did not scroll all the way.
- Change mobile drawer to full height when it is initially opened.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.0.0-r389",
+  "version": "1.0.0-r384",
   "main": "src/index.jsx",
   "homepage": "https://github.com/bldrs-ai/Share",
   "bugs": {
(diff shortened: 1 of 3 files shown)


src/utils/cutPlane.js
```diff
@@ -0,0 +1,203 @@
+import {Box3, BufferAttribute, BufferGeometry, Mesh, Vector3} from 'three'
```
pablo-mayrgundter, line 1: I think remove this file from the PR?

pablo-mayrgundter (changes requested):
Looks like you got an unexpected file included in this

OlegMoshkovich:
removed. PTAL

pablo-mayrgundter (approved):

PR #339, opened 2022-08-19
System: Get element properties correctly when the tree is not loaded into the DOM.

OlegMoshkovich:
This PR resolves https://github.com/bldrs-ai/Share/issues/318.
For the Property panel to display the properties of an element, the properties need to be obtained from the ifc.js.
On element selected is triggered when the location is changed, in the PR it was moved from the Nav Panel to CadView, so it is triggered regardless if the tree is loaded into the DOM.

### DIFF
--- a/__mocks__/web-ifc-viewer.js
+++ b/__mocks__/web-ifc-viewer.js
@@ -1,0 +1,25 @@
+const ifcjsMock = jest.createMockFromModule('web-ifc-viewer')
+export default ifcjsMock
+
+const constructorMock = ifcjsMock.IfcViewerAPI
+// Not sure why this is required, but otherwise these internal fields
+// are not present in the instantiated IfcViewerAPI.
+const impl = {
+  IFC: {
+    context: {
+      ifcCamera: {
+        cameraControls: {},
+      },
+    },
+    setWasmPath: jest.fn(),
+    loadIfcUrl: jest.fn(),
+  },
+  clipper: {
+    active: false,
+  },
+  context: {
+    resize: jest.fn(),
+  },
+}
+constructorMock.mockImplementation(() => impl)
+export {constructorMock as IfcViewerAPI}
--- a/src/Containers/CadView.test.jsx
+++ b/src/Containers/CadView.test.jsx
@@ -3,7 +3,7 @@ import {render, waitFor, renderHook, act} from '@testing-library/react'
 import CadView from './CadView'
 import ShareMock from '../ShareMock'
 import useStore from '../store/useStore'
-import {IfcViewerAPI} from 'web-ifc-viewer'
+// import {IfcViewerAPI} from 'web-ifc-viewer'
 
 
 // TODO(pablo): This mock suppresses "WARNING: Multiple instances of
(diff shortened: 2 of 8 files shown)


pablo-mayrgundter:
That's in the right direction, but unfortunately breaks permalinks:

https://deploy-preview-339--bldrs-share.netlify.app/share/v/p/index.ifc/89/112/139/154/396

Isn't selected.

A test that catches this is very welcome!

OlegMoshkovich:
@pablo-mayrgundter PTAL

pablo-mayrgundter:
I think the test here is URL -> assert that selectedElement(s) matches.

pablo-mayrgundter:
I think the test here is URL -> assert that selectedElement(s) matches.

OlegMoshkovich:
ok ... that makes sense.

OlegMoshkovich:
@pablo-mayrgundter it is not clear how to test this - perhaps it is blocked by pr#348.
Please help with tests to make it production ready.
The pr has been opened for 5 days.

pablo-mayrgundter:
Sure how can I help?  #348 is merged.  You should be able to make use of it

pablo-mayrgundter:
I took a crack.  Still not working, so handing off for you today.

CadView.test.jsx
```
  it('', async () => {
    const modelId = 1234
    const modelPath = {
      filepath: `index.ifc/${modelId}`,
      gitPath: 'foo',
    }
    const {result} = renderHook(() => useStore((state) => state))
    act(() => result.current.setModelPath(modelPath))
    render(
        <ShareMock>
          <CadView
            installPrefix={''}
            appPrefix={''}
            pathPrefix={''}
          />
        </ShareMock>)
    await waitFor(() => screen.getByTitle(/Bldrs: 1.0.0/i))
    // The mocked class actually returns a singleton instance.
    const viewer = new IfcViewerAPI()
    expect(viewer.getProperties.mock.calls.length).toBe(1) # fails, actually 0
    const firstCall = viewer.getProperties.mock.calls[0]
    expect(firstCall[0]).toBe(0) // modelId
    expect(firstCall[1]).toBe(modelId) // elementId
  })
```
small change to the ifcjs mock (the line starting with a plus):
```
pablo@top:~/c/pm/Share> git diff __mocks__/web-ifc-viewer.js
...
@@ -20,6 +19,7 @@ const impl = {
   context: {
     resize: jest.fn(),
   },
+  getProperties: jest.fn(),
 }
 constructorMock.mockImplementation(() => impl)
 export {constructorMock as IfcViewerAPI}
```

OlegMoshkovich:
working on this - are you getting the viewer object from the mock?

__mocks__/web-ifc-viewer.js
```diff
@@ -0,0 +1,25 @@
+const ifcjsMock = jest.createMockFromModule('web-ifc-viewer')
+export default ifcjsMock
+
```
OlegMoshkovich, line 3: @pablo-mayrgundter have it in this branch

src/Containers/CadView.test.jsx
```diff
@@ -0,0 +1,43 @@
...
+          <CadView
+            installPrefix={''}
+            appPrefix={''}
+            pathPrefix={''}
+            modelPath={modelPath}
+          />
+        </ShareMock>)
+    debug()
+    await act(() => result.current.setModelPath(modelPath))
+    const viewer = new IfcjsMock()
```
OlegMoshkovich, line 37: @pablo-mayrgundter how are we making this viewer?

pablo-mayrgundter:
I think you can delete this PR.. I ended up fixing #318 along the way in #358.

OlegMoshkovich:
nice!

Issue #337, opened 2022-08-18
Tests: zustand test warnings


pablo-mayrgundter:
Many zustand tests currently generate a warning as below.  Not sure how to solve, but may require refactoring zustand tests to just mock zustand instead of using it.

```
Warning: An update to TestComponent inside a test was not wrapped in act(...).

      When testing, code that causes React state updates should be wrapped into act(...):

      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */

      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at TestComponent (/Users/pablo/c/pm/Share/node_modules/@testing-library/react/dist/pure.js:281:5)
```

pablo-mayrgundter:
@OlegMoshkovich could you take this?  I tried debugging this but figure you'll have a better time of it.

OlegMoshkovich:
yes i can this one

pablo-mayrgundter:
I think it's possible the use of renderHook here is not as intended.. check this:
https://stackoverflow.com/questions/64287448/react-hook-testing-with-renderhook

pablo-mayrgundter:
I'm perplexed.. I banged my head on this for awhile.

I made a new repo to just test zustand with react hooks.  Everything works fine.

https://github.com/pablo-mayrgundter/react-zustand-jest

pablo-mayrgundter:
I've whittled it down to just this causing the warning in SideDrawer.test.jsx, first test, and surprisingly, it's the expect line that triggers the warning:

```
test('side drawer notes', async () => {
  render(<ShareMock><SideDrawerWrapper/></ShareMock>)
  expect(await screen.findByText('Notes')).toBeInTheDocument()
})
```

I read up on this:

https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

Current theory is that we have async code resolving after the expect, and testing-lib is telling us with the warning.  I suspect it's loading the issues async somehow.. so it may be like needing a findBy after render that looks for something in the issues.

Better yet would be to mock out the async results with something synchronous.  Would speed up tests too.

pablo-mayrgundter:
https://github.com/pmndrs/zustand/issues/363#issuecomment-845536384

"if you have an effect that updates the store, then the problem will happen again".  Sounds like us!

PR #353, opened 2022-08-25
Tests: Fix test warning/error caused by updating props in the zustand store

OlegMoshkovich:
This PR address #337.
The warnings were caused by updating the store before render, which is incorrect order, and not awaiting for the update.
- Button Test is still throwing the error,  since we are testing for this exact error - the behavior is correct.
- Side drawer - opened via URL'  still causing a warning - it is commented out for now. 
- Permalink Mock needs more attention


src/Components/IssuesControl.test.jsx
```diff
@@ -1,57 +1,57 @@
...
 })
 
 
-test('Issues ', () => {
+test('Setting comments in zustand ', async () => {
   const {result} = renderHook(() => useStore((state) => state))
   const testIssueId = 10
-  act(() => {
+  const {getByText} = render(<ShareMock><Issues/></ShareMock>)
+  await act(() => {
```
OlegMoshkovich, line 43: @pablo-mayrgundter this worked to resolve the warning

OlegMoshkovich (commented):
@pablo-mayrgundter left a comment

src/Components/IssuesControl.test.jsx
pablo-mayrgundter: Heya, just getting to this.  Thanks but also needs a merge now.  Ogali has a PR for just the await fix that will probably be in soon too.

OlegMoshkovich: I think Ogali merged his fix.
there are conflicts now. 
shall I rebase?

pablo-mayrgundter: Yes sure

pablo-mayrgundter (commented):
Ok, basically looks good.  Maybe we can chat on Discord or talk?  I'd like to understand the fix better.

High-level question too, looks like you reverted my switch to describe/it?

pablo-mayrgundter:
Per discussion, please use describe/it.  describe/it provides 3 scopes: global, suite and test.  global is top-level statements.. suite allows setup and tear down for this set of tests.  I've been moving us towards that.

OlegMoshkovich:
@pablo-mayrgundter PTAL

pablo-mayrgundter (approved):

PR #361, opened 2022-08-30
Correct improper usage of async/await

oogali:
Within the test suites, `await` is needed to ensure that the environment is properly set up before performing our assertions.

Within the code base:
* Functions were declared as `async`, but did not contain any asynchronous execution
* Asynchronous functions were being executed without resolving the promise (or otherwise awaiting for the result)
* The result of an asynchronous function was being directly returned with `return await` which would obscure a stack trace.


pablo-mayrgundter (approved):
LGTM!  Didn't know act was async.  Thanks :)

PR #363, opened 2022-08-30
Fix bad/missing imports

oogali:
PR to clean up issue regarding ES6 imports:
* Some imports weren't directly referenced. This means if the parent dependencies were removed from the project for whatever reason (e.g. refactor), other portions of the code would appear to no longer work.
* Correct reference to `react-router-dom`

This PR also removes some unused code regarding Google Analytics, which could not possibly work because:
1. the import it required was not installed in the project, and
2. it contained placeholder GA IDs.


OlegMoshkovich:
LGTM

OlegMoshkovich (approved):

pablo-mayrgundter (approved):

PR #367, opened 2022-08-31
De-duplicate bundling efforts

oogali:
Currently, there are some actions that are undertaken for building the project that duplicate esbuild plugin functionality.

This PR adds and configures said plugins, then removes the corresponding actions that were previously performed externally.

### DIFF
--- a/config/common.js
+++ b/config/common.js
@@ -1,12 +1,20 @@
 import svgrPlugin from 'esbuild-plugin-svgr'
 import copyStaticFiles from 'esbuild-copy-static-files'
 import {cleanPlugin} from 'esbuild-clean-plugin'
+import {fileURLToPath} from 'url'
+import * as path from 'node:path'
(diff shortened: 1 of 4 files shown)


pablo-mayrgundter (commented):
I need to try this.. will do this weeken. I think I had the shared docs build for a reason.

pablo-mayrgundter:
TODO: if this goes in pls update https://github.com/bldrs-ai/Share/wiki/Dev:-Guide

oogali:
> I need to try this.. will do this weeken. I think I had the shared docs build for a reason.

`yarn build-docs` would remove the target directory and copy the static assets from `public/` into it.

This functionality has not been removed. Instead, it is now performed by the esbuild plugins: `cleanPlugin()` and `copyStaticFiles()`.

With the plugins in place, the yarn script is redundant.

PR #393, opened 2022-09-15
Move testing library into development dependencies

oo-bldrs:
We don't want development dependencies in our production builds.


pablo-mayrgundter:
Whoops, thanks!

PR #397, opened 2022-09-17
Restore live-reloading.  Fix 404 bounce handler!

pablo-mayrgundter:
This fixes the live-reload bug.

I bisected it down to cleanPlugin.  No idea why that was causing trouble.  Maybe related, there's no contents in the docs directory.. esbuild keeps the build in memory.

ALSO..... Fixed the 404 bounce to work locally.  Not sure this ever worked before :(. I hadn't set the proxy up right.

We can finally reload any URL no problem!


OlegMoshkovich (approved):
LGTM

oo-bldrs (approved):

Issue #406, opened 2022-09-26
Bug: Text in Notes window is cut at both sides


MarkusSteinbrecher:
Some users have an issue with notes display:

![image](https://user-images.githubusercontent.com/74647806/192228635-a126a57c-e923-4fe4-9ef5-6e1ea2da053a.png)

Chrome version: Version 105.0.5195.125 (Offizieller Build) (x86_64)

pablo-mayrgundter:
Could you post a link?  Can't reproduce.  Desktop?

MarkusSteinbrecher:
https://bldrs.ai/share/v/gh/Swiss-Property-AG/Turmfalke-Public/main/TURMFALKE.ifc#i:1340200278::c:35.31,67.23,58.84,-2.77,7.81,15.5

Desktop

pablo-mayrgundter:
Markus can repro in Chrome/Safari and FF.

I can't repro in those.

We don't have an IE tester.

OlegMoshkovich:
Look ok for me.

MarkusSteinbrecher:
fixed with #407

OlegMoshkovich:
<img width="1402" alt="image" src="https://user-images.githubusercontent.com/3433606/192374999-c1050607-eb1c-4476-9f0d-3c39daefe5c6.png">

Steps to reproduce - set the chrome default font to small in settings

MarkusSteinbrecher:
Looking good

PR #407, opened 2022-09-26
Testing overflow fix for Markus browser.

pablo-mayrgundter:
Fix for Markus's browser bug #406.

Just a couple of padding changes seems to do it for both of us.

### DIFF
--- a/src/Components/IssueCard.jsx
+++ b/src/Components/IssueCard.jsx
@@ -277,7 +277,6 @@ const CardActions = ({
 
 const useStyles = makeStyles((theme) => ({
   container: {
-    'width': '29em',
     'marginBottom': '20px',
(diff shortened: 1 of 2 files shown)


MarkusSteinbrecher:
Looks good to me!

src/Components/IssueCard.jsx
```diff
@@ -277,7 +277,6 @@ const CardActions = ({
 
 const useStyles = makeStyles((theme) => ({
   container: {
-    'width': '29em',
```
OlegMoshkovich, line 280: @pablo-mayrgundter I am not sure how to set the issue card size based on the outer container.
The issue cards are located in the notes panel, if the width of the notes panel is set explicitly it just scrolls in the y direction, because the body of the issue takes more space than the width of the notes panel.

Issue #401, opened 2022-09-20
Bug: Scroll bar always visible ... just on my laptop :(


MarkusSteinbrecher:
![image](https://user-images.githubusercontent.com/74647806/191254197-742c9c22-5eb8-46e8-8fd9-875646d642fa.png)

Chrome: Version 105.0.5195.125 (Offizieller Build) (x86_64)
Mac: 12.6 (21G115)

pablo-mayrgundter:
@MarkusSteinbrecher can you check again.  We did a big refactor.. hopefully this resolved

MarkusSteinbrecher:
Yes, it is gone! Great!

pablo-mayrgundter:
We did fix 2 of the vertical scrollbars, but the third remains elusive.

MarkusSteinbrecher:
The third one is a setting on mac, depending on whether one uses the trackpad or a mouse. I am closing this issue.

PR #408, opened 2022-09-26
ItemProperties: autoscroll fix more scrollbars in #401.

pablo-mayrgundter:
For #401.

### DIFF
--- a/src/Components/ItemProperties.jsx
+++ b/src/Components/ItemProperties.jsx
@@ -113,7 +113,6 @@ const useStyles = makeStyles((theme) => ({
     height: '100%',
     width: '100%',
     padding: '0px 0px 50px 0px',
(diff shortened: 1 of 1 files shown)


MarkusSteinbrecher:
Looks good!

MarkusSteinbrecher:
###Just as a note to this work:

Regarding scroll bars, there is a different behavior on macs when using the touch pad versus a mouse. 

If I have a am using a mouse the scroll bar is always visible:

image

if I am not using a mouse but the touch pad, that scroll bar on the right disappears.

https://stackoverflow.com/questions/44668257/different-looks-on-scrollbars-when-using-usb-mouse-vs-wireless-trackpad-in-os-x

"Conclusion: I think the solution is beyond our/developers' control. Just leave it to the user."

PR #421, opened 2022-10-03
add loader

OlegMoshkovich:
Add loader component and integrated it into the notes panel.

PR #427, opened 2022-10-07
Use Mock Service Worker

oo-bldrs:
[Mock Service Worker](https://mswjs.io) is a tool for mocking APIs by operating as a service worker and intercepting network requests.

This PR replaces the existing Octokit mock (and its associated conditionals) with the native Octokit library and interception of specific URLs to achieve the same mocking functionality as before.

MSW will allow us to mock other network calls in the future as we expand the functionality of Share.

No changes to unit tests were necessary and all continue to pass as before.

### DIFF
--- a/src/index.jsx
+++ b/src/index.jsx
@@ -4,6 +4,11 @@ import {BrowserRouter} from 'react-router-dom'
 import BaseRoutes from './BaseRoutes'
 
 
+if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test') {
+  const {worker} = require('./__mocks__/browser')
+  worker.start()
+}
+
 const root = createRoot(document.getElementById('root'))
 root.render(
     <BrowserRouter>
--- a/package.json
+++ b/package.json
@@ -94,8 +94,12 @@
     "jest": "^28.1.3",
     "jest-environment-jsdom": "^28.1.3",
     "jsdom": "^20.0.0",
+    "msw": "^0.47.4",
     "shx": "^0.3.4",
     "storybook-addon-material-ui": "^0.9.0-alpha.24",
     "storybook-addon-turbo-build": "^1.1.0"
+  },
+  "msw": {
+    "workerDirectory": "public"
   }
 }
--- a/public/mockServiceWorker.js
+++ b/public/mockServiceWorker.js
@@ -1,0 +1,303 @@
+/* eslint-disable */
+/* tslint:disable */
+
+/**
+ * Mock Service Worker (0.47.4).
+ * @see https://github.com/mswjs/msw
+ * - Please do NOT modify this file.
+ * - Please do NOT serve this file on production.
+ */
+
+const INTEGRITY_CHECKSUM = 'b3066ef78c2f9090b4ce87e874965995'
+const activeClientIds = new Set()
+
+self.addEventListener('install', function () {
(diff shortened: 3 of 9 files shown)


oo-bldrs:
This works when I test it locally on my machines (both primary and secondary with a clean directory) but not in GitHub actions. I'll need to take a look at this later.

pablo-mayrgundter:
Did you see #421?  We introduced https://github.com/bldrs-ai/Share/blob/main/__mocks__/%40octokit/rest.js

That achieves a similar goal and gives us more explicit control of preconditions for tests.

One reason for having the MockOctokit was to enable localhost serving during dev.  I often hit GH rate limit for comments without it.  Now realizing the name is confusing.  LocalOctokit?

oo-bldrs:
@pablo-mayrgundter I did not see #421 prior to this PR, thank you for pointing that out.

I'm not sure I understand your statement about controlling test preconditions, do you mean being able to set expected request paths per individual test? If not, can you elaborate?

https://www.danieljcafonso.com/configuring_msw

MSW intercepts requests made in the browser, meaning if you define a route handler (and corresponding response), it will be handled locally and not touch the Internet despite the target URL having an external URL -- GitHub rate limits won't come into play.

https://github.com/bldrs-ai/Share/pull/427/files#diff-a4994793da5ecfe4df45523fd07d4c2d368d46c4d3c2a3e4d902d11175024ab7R9

MSW's request interception allows us to have a single code path regardless of whether we're running in a local development environment, a CI/CD pipeline, or in production.

As part of the testing for the authentication features, I need to be able to assert against headers (or their lack thereof) and send back line level responses (e.g. 401 Unauthorized) that are processed natively by Octokit.

Function-level mocking of Octokit doesn't get me there and it would get fairly messy (and fragile) to inject a HTTP client instance into Octokit to achieve the same level of functionality, if that's even possible.

src/index.jsx
```diff
@@ -4,6 +4,17 @@ import {BrowserRouter} from 'react-router-dom'
 import BaseRoutes from './BaseRoutes'
 
 
+if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test') {
```
pablo-mayrgundter, line 7: Per convo, can this be injected only in dev from something clean in esbuild?  Would be nice to not have this present in prod.

oo-bldrs: I've confirmed that this is already optimized out by esbuild's tree shaking algorithm.

At build time, the value of `process.env.NODE_ENV` in any bundled JavaScript is evaluated to a value of `development` or `production` depending on whether minification has been requested.

https://github.com/evanw/esbuild/blob/master/pkg/api/api_impl.go#L552..L572

In #367, I introduced conditional minification depending on the value of the `NODE_ENV` environment variable because I wanted to be better troubleshoot issues in non-production environments.

https://github.com/bldrs-ai/Share/blob/main/config/common.js#L19

Then because the values are known are build time, the expression is evaluated down to `true` or `false`. If false, esbuild throws it out.

Here's the evidence in the form of the code result in a non-production environment (from the build artifact: `docs/index.js`):

image

oo-bldrs:
@pablo-mayrgundter PTAL, I've merged in the code from #421 and added logic to the build process to delete the MSW static asset from non-development builds.

pablo-mayrgundter (approved):

PR #430, opened 2022-10-08
Minor UI adjustments

OlegMoshkovich:
- highlight open button on the first load
- close the cutplane menu when the direction of the cut plane is selected

### DIFF
--- a/src/Components/AboutControl.jsx
+++ b/src/Components/AboutControl.jsx
@@ -80,7 +80,7 @@ function AboutContent({installPrefix}) {
   const privacyLevelFunctional = 0
   const privacyLevelUsage = 10
   const privacyLevelSocial = 20
-  const bulletStyle = {textAlign: 'center'}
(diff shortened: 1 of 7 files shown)


pablo-mayrgundter (approved):
Nice, that cut plane menu was causing some consternation!  :)

Maybe colorize the open highlight?  Green or blue?

OlegMoshkovich:
@pablo-mayrgundter 
I don't understand what you mean 'colorize the open highlight'.
But in general dark green specified in the theme is the color of choice.

OlegMoshkovich:
Ah yeah - colorizing the open highlight would be nice, but a bit more involved. Can do in another PR. Would be nice to get this one in.

pablo-mayrgundter:
Alright, LGTM.. would like to see color.. as noted in discussion, this doesn't really come off as a button.  I think since it's near the input, it just looks like a pointer or label for that.  But however you want to do it!

PR #446, opened 2022-10-21
Branches UI

OlegMoshkovich:
- Add logic to fetch branches from Github.
- Add Branch UI drop down containing the list of branches.

-- Tests are missing

### DIFF
--- a/src/Components/BranchesControl.jsx
+++ b/src/Components/BranchesControl.jsx
@@ -1,0 +1,144 @@
+import React, {useContext, useEffect, useState} from 'react'
+import {useNavigate} from 'react-router-dom'
+import TextField from '@mui/material/TextField'
+import MenuItem from '@mui/material/MenuItem'
+import Paper from '@mui/material/Paper'
+import Typography from '@mui/material/Typography'
+import {makeStyles} from '@mui/styles'
+import {ColorModeContext} from '../Context/ColorMode'
+import debug from '../utils/debug'
+import {getBranches} from '../utils/GitHub'
+import useStore from '../store/useStore'
+import {navigateBaseOnModelPath} from '../utils/location'
+
+
+/**
+ * @param {boolean} isDialogDisplayed
+ * @param {Function} setIsDialogDisplayed
+ * @return {object} React component
+ */
+export default function Branches() {
+  const classes = useStyles()
+  const navigate = useNavigate()
+  const colorMode = useContext(ColorModeContext)
+  const repository = useStore((state) => state.repository)
+  const [branches, setBranches] = useState([])
+  const [versionPaths, setVersionPaths] = useState([])
+  const [selected, setSelected] = useState(0)
+  const modelPath = useStore((state) => state.modelPath)
+
+  useEffect(() => {
+    if (!repository) {
+      debug().warn('IssuesControl#Issues: 1, no repo defined')
+      return
+    }
+    const fetchBranches = async () => {
+      try {
+        const branchesData = await getBranches(repository)
+        const versionPathsTemp = []
+        if (branchesData.data.length > 0) {
+          setBranches(branchesData.data)
+        }
+        branchesData.data.map((branch, i) => {
+          if (branch.name === modelPath.branch) {
+            // select the current branch
+            setSelected(i)
+          }
+          const versionPath = navigateBaseOnModelPath(modelPath.org, modelPath.repo, branch.name, modelPath.filepath)
+          versionPathsTemp.push(versionPath)
+        })
+        setVersionPaths(versionPathsTemp)
+      } catch (e) {
+        debug().warn('failed to fetch branches', e)
+      }
+    }
+
+    if (branches.length === 0 && modelPath.repo !== undefined) {
+      fetchBranches()
+    }
+  }, [repository])
+
+
+  const handleSelect = (event) => {
+    const versionNumber = event.target.value
+    setSelected(versionNumber)
+    navigate({
+      pathname: versionPaths[versionNumber],
+    })
+  }
+
+  return (
+    <>
+      {branches.length > 1 && modelPath.repo !== undefined &&
+        <Paper elevation={0}
+          sx={{
+            backgroundColor: colorMode.isDay() ? '#E8E8E8' : '#4C4C4C',
+            marginTop: '34px',
+            opacity: .8,
+          }}
+        >
+          <TextField
+            className={classes.dropDown}
+            onChange={(e) => handleSelect(e)}
+            variant='outlined'
+            label='Git Branches / Project Versions'
+            value={selected}
+            select
+            role="button"
+          >
+            {branches.map((branch, i) => {
+              return (
+                <MenuItem
+                  key={i}
+                  value={i}
+                >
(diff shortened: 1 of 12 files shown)


OlegMoshkovich:
@oo-bldrs when you have a chance, can you please take a look at the logic of the BranchControl component, and suggest the way to test it?

oo-bldrs:
"Versions" is a bit of an ambiguous term.

In a Git repository, you can refer to a "version" through branches or tags. In the context of Git, a version would refer to a tag that generally starts with `v` (e.g. `v0.0.1`, `v1.0`, `v2.0.1`), and I imagine folks would expect that label to correspond with a list of versions.

Maybe it's best to change the label to Git branch or another term that is less overloaded.

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,132 @@
...
+          setBranchNames(branchesData.data)
+        }
+        branchesData.data.map((branch) => {
+          modelPath.branch = branch.name
+          const versionPath = `/share/v/gh/${modelPath.org}/${modelPath.repo}/${modelPath.branch}${modelPath.filepath}`
+          versionPathsLocal.push(versionPath)
+        })
+        setVersionPaths(versionPathsLocal)
+      } catch (e) {
+        debug().warn('failed to fetch issues', e)
```
oo-bldrs, line 48: Is this warning supposed to reference `issues`?

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,132 @@
...
+    const fetchBranches = async () => {
+      try {
+        const branchesData = await getBranches(repository)
+        const versionPathsLocal = []
+        if (branchesData.data.length > 0) {
+          setBranchNames(branchesData.data)
+        }
+        branchesData.data.map((branch) => {
+          modelPath.branch = branch.name
+          const versionPath = `/share/v/gh/${modelPath.org}/${modelPath.repo}/${modelPath.branch}${modelPath.filepath}`
```
oo-bldrs, line 43: Between this invocation and the logic in `<ShareRoutes>`, it makes sense to create a utility function for building a URL to a GitHub-hosted model -- this logic should no be duplicated in multiple places.

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,132 @@
...
+    })
+  }
+
+  return (
+    <Paper elevation={0}
+      sx={{
+        backgroundColor: colorMode.isDay() ? '#E8E8E8' : '#4C4C4C',
+        marginTop: '14px',
+        opacity: .8,
+        display: ( modelPath.repo !== undefined && branchNames.length ) < minVersions ? 'none' : 'box',
```
oo-bldrs, line 70: Is there a compelling reason to use CSS to hide this from being displayed as opposed to _not_ rendering these components at all if the necessary conditions aren't met?

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,132 @@
...
+      }}
+    >
+      <TextField
+        className={classes.dropDown}
+        onChange={(e) => handleSelect(e)}
+        variant='outlined'
+        label='Versions'
+        select
+      >
+        {branchNames.length > 0 && branchNames.map((branchName, i) => {
```
oo-bldrs, line 80: Checking if the list of known branches is greater than zero is redundant when using map to iterate through each member of the array.

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,132 @@
...
+          return (
+            <MenuItem
+              key={i}
+              value={i}
```
oo-bldrs, line 84: What is the `value` property used for? It doesn't appear to be part of the defined props for `<MenuItem>`.

https://mui.com/material-ui/api/menu-item/

src/Components/BranchesControl.jsx
OlegMoshkovich: addressed.

src/Components/BranchesControl.jsx
OlegMoshkovich: addressed.

src/Components/BranchesControl.jsx
OlegMoshkovich: addressed from the CadView. 
when the model is not loaded form the repo the BranchControl is not rendered.

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,132 @@
...
+  const [versionPaths, setVersionPaths] = useState([])
+  const modelPath = useStore((state) => state.modelPath)
+  const minVersions = 2
+
+  useEffect(() => {
+    if (!repository) {
+      debug().warn('IssuesControl#Issues: 1, no repo defined')
+      return
+    }
+    const fetchBranches = async () => {
```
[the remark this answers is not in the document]
OlegMoshkovich: the function is called lated in the useEffect.
it is only called when the model is loaded from github repo.

src/Components/BranchesControl.jsx
OlegMoshkovich: @oo-bldrs In the share routes, there is extractOrgPrefixedPath function, but I am not seeing where are are using ModelPath object to navigate.
Can you please point me to the function you have in mind that makes sense to include in the utility?

src/Components/BranchesControl.jsx
```diff
@@ -0,0 +1,144 @@
...
+          versionPathsTemp.push(versionPath)
+        })
+        setVersionPaths(versionPathsTemp)
+      } catch (e) {
+        debug().warn('failed to fetch branches', e)
+      }
+    }
+
+    if (branches.length === 0 && modelPath.repo !== undefined) {
+      fetchBranches()
```
pablo-mayrgundter, line 57: Does await on this break the test?  If not I think the way we had it with the IIFE is more correct

pablo-mayrgundter (commented):
Heya, I'm seeing a lot of merge deltas.. does this need to be rebased or smth?

Also, do you have a link that shows the branches?

pablo-mayrgundter:
Found the Seestrasse one:

https://deploy-preview-446--bldrs-share.netlify.app/share/v/gh/Swiss-Property-AG/Seestrasse-Public/main/SEESTRASSE.ifc

Cool!

src/Components/BranchesControl.jsx
OlegMoshkovich: I don't think we need IIFE - i have never seen this pattern, and the logic works.

src/__mocks__/web-ifc-viewer.js
```diff
@@ -0,0 +1,26 @@
...
+  clipper: {
+    active: false,
+  },
+  context: {
+    resize: jest.fn(),
+  },
+  getProperties: jest.fn(),
+}
+constructorMock.mockImplementation(() => impl)
+export {constructorMock as IfcViewerAPI}
```
OlegMoshkovich, line 26: @pablo-mayrgundter not sure why this file is the part of the PR.

pablo-mayrgundter (approved):

PR #455, opened 2022-10-26
VYZN: Matrix Messaging Integration

clecherbauer:
This pull request implements the lib "matrix-widget-api" which allows us to send control-messages to an embedded (iframe) bldrs instance.

Messages bldrs can receive:
- ai.bldrs-share.LoadModel (bldrs loads a ifc-model from github)
- ai.bldrs-share.SelectElements (bldrs select elements from the current model in the viewer)

Messages bldrs will emit:
- ai.bldrs-share.ElementsSelected (triggers if elements are added to the selection)
- ai.bldrs-share.ElementsDeSelected (triggers if elements are removed from the selection)


oo-bldrs:
The title of this PR should be changed to reflect the contents, which if I'm guessing correctly is: Matrix Messaging Integration?

oo-bldrs (changes requested):
I've added some comments regarding what's present so far.

src/WidgetApi/ApiConnection.js
```diff
@@ -0,0 +1,26 @@
+class AbstractApiConnection {
+
+    on(event, callable) {
```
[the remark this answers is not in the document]
clecherbauer: good point

src/index.jsx
```diff
@@ -2,9 +2,12 @@ import React from 'react'
 import {createRoot} from 'react-dom/client'
 import {BrowserRouter} from 'react-router-dom'
 import BaseRoutes from './BaseRoutes'
-
+import WidgetApi from "./WidgetApi/WidgetApi";
 
 const root = createRoot(document.getElementById('root'))
+
+new WidgetApi()
```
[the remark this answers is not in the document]
clecherbauer: The Problem here is that we need to initialize this class at a place which doesn't get reinitialized. I debugged this and the outcome was that any component gets reloaded at routing time. Since both ends of matrix-widget-api relies on a specific order of events there is no stable communication when initialized in routed components.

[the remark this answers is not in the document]
clecherbauer: but i'm not specialized in building react-apps maybe you guys know a better way of doing this :)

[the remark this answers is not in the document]
pablo-mayrgundter: I don't know the answer, but I think this is worth solving.  Happy to help debug together.  Have you tried in src/Share.jsx or its parent components?  It only rerenders on new models, but I think that's probably desirable here.

src/WidgetApi/WidgetApi.js
```diff
@@ -0,0 +1,23 @@
...
+        if (this.detectIframe()) {
+            let apiConnection = new ApiConnectionIframe()
+            new ApiEventsRegistry(apiConnection)
+            apiConnection.start()
+        }
+    }
+
+    detectIframe () {
+        try {
+            return window.self !== window.top;
```
pablo-mayrgundter, line 16: Can this actually throw?

src/WidgetApi/WidgetApi.js
```diff
@@ -0,0 +1,23 @@
+import ApiConnectionIframe from "./ApiConnectionIframe";
+import ApiEventsRegistry from "./ApiEventsRegistry";
+
+class WidgetApi {
+
+    constructor() {
+        if (this.detectIframe()) {
+            let apiConnection = new ApiConnectionIframe()
```
pablo-mayrgundter, line 8: const

src/store/useStore.js
```diff
@@ -6,6 +6,7 @@ import createRepositorySlice from './RepositorySlice'
 
 
 const useStore = create((set, get) => ({
+  widgetApiUri: null,
```
pablo-mayrgundter, line 9: I suspect this will grow quickly.  Please make a slice for it called ApiSlice with field URI

pablo-mayrgundter (changes requested):
Thanks, nice start!

Let's leave this in draft until we get tests and styles.  Left you some initial commements.  I'll help with the global init when we can chat.

pablo-mayrgundter:
Here's our style guide:

  https://github.com/bldrs-ai/Share/wiki/Dev:-Style

Please use our eslint config or run ```yarn lint``` manually

src/WidgetApi/WidgetApi.js
clecherbauer: This is the source of this function: [https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t](https://stackoverflow.com/questions/326069/how-to-identify-if-a-webpage-is-being-loaded-inside-an-iframe-or-directly-into-t)

src/index.jsx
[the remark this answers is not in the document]
clecherbauer: i tested it again today and it only works stable at this place

pablo-mayrgundter:
Per discussion, I'd try useMemo or useCallback for the init logic.  Assuming that works, the right place is probably in src/Share.jsx.

pablo-mayrgundter:
Hey Christian, how's this going?

clecherbauer:
@pablo-mayrgundter i still need to write tests. Could you do another review now before is start with that?

src/WidgetApi/WidgetApi.js
```diff
@@ -0,0 +1,33 @@
+import ApiConnectionIframe from './ApiConnectionIframe'
+import ApiEventsRegistry from './ApiEventsRegistry'
+
+
+/**
+ * WidgetApi main class
+ */
+class WidgetApi {
```
pablo-mayrgundter, line 8: export default class WidgetApi

src/WidgetApi/WidgetApi.js
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * constructor
```
pablo-mayrgundter, line 10: pls add type annotations

src/WidgetApi/ApiEventsRegistry.js
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * check if state has changed.
+   *
+   * @param {object} state
+   * @param {string[]} lastSelectedElementIds
+   * @return {boolean}
+   */
+  selectedElementIdsHasChanged(state, lastSelectedElementIds) {
+    if (Array.isArray(state.selectedElements)) {
+      return JSON.stringify(lastSelectedElementIds) !== JSON.stringify(this.getSelectedElementIds(state))
```
pablo-mayrgundter, line 92: This seems brittle.. maybe array->set for each and compare sets?

pablo-mayrgundter (changes requested):
Basically looking good.

Would you please update the description of the PR to summarize the changes and a link to the staging instance you're testing with.

pablo-mayrgundter (approved):
Hey Christian, tests look good!

It's been a long time since you started, so there's a chance this may conflict with some in-flight changes.  Looks like you need to merge.  If that goes fine, I'll merge.

Thanks!  Happy Holidays!
Pablo

clecherbauer:
hey @pablo-mayrgundter, i did the merge

pablo-mayrgundter:
Hmm, I'm not seeing the netlify dialog change.  can you try updating smth like that yarn.lock and see if that triggers it?

clecherbauer:
OK, i did update yarn.lock. But does not seem like it triggered the deployment. We really need to have the current version of this code deployed since the viewer in our production is loaded from this preview-instance.

oo-bldrs:
I manually flushed the cache and retried the build. Does the deploy preview reflect your changes?

https://deploy-preview-455--bldrs-share.netlify.app/

clecherbauer:
yes, thank you

Adrian62D:
Hi Pablo, going directly for #525 would be easier. However, if you wish you can first merge this #455. But we should not invest more time #455.

pablo-mayrgundter:
Per discussion with Adrian yesterday, Christian, please do finish this out.  I think we're good to go but pending a successful merge and final review (hopefully a no-op).

clecherbauer:
@pablo-mayrgundter Hi Pablo, @aozien has rebased and fixed my changes. I think we're ready for your final review now :)

pablo-mayrgundter:
Looks like we got broken somehow :(

I tried loading a model and got this error.  I see some maybe related changes in the last update.

image

aozien:
@pablo-mayrgundter  It turns out the problem was caused by the (yarn.lock) file re-generated in commit (a8ed57d), since that did some minor-version updates to some packages, which caused some functionality to break.
The yarn lock file is now changed to match exactly the one on share/main for the already existing packages.

pablo-mayrgundter:
Works for me now, thanks.

Found another issue.. permalinks for camera and cutplane are broken :(

https://deploy-preview-455--bldrs-share.netlify.app/share/v/p/index.ifc/89/112/139/154/621#c:43.83,54.18,47.56,-47,15,-5.73::p:x=0

When I follow this link, everything after the index.ifc is removed after load.  Same on a loaded model.

Please fix that and add an integration test for it.

Thank you!

aozien:
@pablo-mayrgundter  this should be fixed now as well :)

pablo-mayrgundter:
Finally!  Congrats guys!

PR #456, opened 2022-10-30
Calculate Section Height based on Level Height

mkeshavarzi:
Section height for each level is calculated based on the level height (l2-l1), instead of an arbitrary section plane. This PR aimed to address issues of IFC with large unit scales.

### DIFF
--- a/src/Components/ExtractLevelsMenu.jsx
+++ b/src/Components/ExtractLevelsMenu.jsx
@@ -32,8 +32,8 @@ export default function ExtractLevelsMenu({listOfOptions, icon, title}) {
   const open = Boolean(anchorEl)
 
   const LEVEL_PREFIX = 'p'
-  const floorOffset = 0.2
-  const ceilingOffset = 0.4
+  const floorOffsetCoefficient = 0.1
+  const ceilingOffsetCoefficient = 0.1
 
   useEffect(() => {
     const planeHash = getHashParams(location, 'p')
@@ -44,6 +44,10 @@ export default function ExtractLevelsMenu({listOfOptions, icon, title}) {
         const levelHash = planeHash.split(':')[1]
         if (isNumeric(levelHash)) {
           const level = parseInt(levelHash)
+          const levelHeight = allLevels[level + 1] - allLevels[level]
+          console.log(levelHeight)
+          const floorOffset = levelHeight * floorOffsetCoefficient
+          const ceilingOffset = levelHeight * ceilingOffsetCoefficient
           createFloorplanPlane(allLevels[level] + floorOffset, allLevels[level + 1] - ceilingOffset, level)
         }
       }
@@ -76,6 +80,10 @@ export default function ExtractLevelsMenu({listOfOptions, icon, title}) {
   }
 
   const isolateFloor = (level) => {
+    const levelHeight = allLevelsState[level + 1] - allLevelsState[level]
+    console.log(levelHeight)
+    const floorOffset = levelHeight * floorOffsetCoefficient
+    const ceilingOffset = levelHeight * ceilingOffsetCoefficient
     createFloorplanPlane(allLevelsState[level] + floorOffset, allLevelsState[level + 1] - ceilingOffset, level)
   }
 
@@ -143,7 +151,7 @@ export default function ExtractLevelsMenu({listOfOptions, icon, title}) {
           <MenuItem
             key={i}
             onClick={() => isolateFloor(i)}
-            selected={levelInstance === (allLevelsState[i] + floorOffset)}
+            selected={levelInstance === (allLevelsState[i])}
           >
           L{i}
           </MenuItem>))


pablo-mayrgundter:
So as discussed, I think the alg is:
```
StoreyHeight = DEFAULT_STOREY_HEIGHT // works for non-Schep examples so far
if (storey has an IfcElementQuantity.Height property in its Property Sets) {
    // fix for Schep
    StoreyHeight = IfcElementQuantity.Height
}
// calculate storey elevation for section as usual
```
image

mkeshavarzi:
> So as discussed, I think the alg is:
> 
> ```
> StoreyHeight = DEFAULT_STOREY_HEIGHT // works for non-Schep examples so far
> if (storey has an IfcElementQuantity.Height property in its Property Sets) {
>     // fix for Schep
>     StoreyHeight = IfcElementQuantity.Height
> }
> // calculate storey elevation for section as usual
> ```
> 

I think I came up with a more comprehensive fix in which we extract the unitType, and incorporate that in the level calculation.

src/utils/extractHeight.js
```diff
@@ -9,12 +9,32 @@ export async function extractHeight(ifcModel) {
     const ifcBuildingStoreyID = 3124254112
```
pablo-mayrgundter, line 9: It's great you found this!  Much better than guessing/computing from storey height.

However, taking a closer look at this code, I think it should be restructured, see below.  More importantly, not sure this is the algorithm we want.  Let's discuss on chat

```
const IFC_TYPE_ID_BUILDING_STOREY = 3124254112
const IFC_TYPE_ID_SI_UNIT = 448429030
const SI_PREFIX_VALUE_MILLI = 0.001
const SI_PREFIX_VALUE_CENTI = 0.01


export async function extractHeight(ifcModel) {
  const elevValues = []
  const storeys = await ifcModel.getAllItemsOfType(IFC_TYPE_ID_BUILDING_STOREY, true)
  const siUnits = await ifcModel.getAllItemsOfType(IFC_TYPE_ID_SI_UNIT, true)

  let unitScale = 1

  for (let i = 0; i < siUnits.length; i++) {
    if (siUnits[i].UnitType.value === 'LENGTHUNIT') {
      const prefix = siUnits[i].Prefix.value
      switch (prefix) {
        case 'MILLI': unitScale = SI_PREFIX_VALUE_MILLI; break
        case 'CENTI': unitScale = SI_PREFIX_VALUE_CENTI; break
        default: {
          console.warn('Unhandled length unit: ', prefix)
          unitScale = 1
        }
      }
    }
  }

  for (let i = 0; i < storeys.length; i++) {
    const elevation = storeys[i].Elevation.value
    if (!isFinite(elevation)) {
      console.warn('Found invalid elevation value: ', elevation)
      continue
    }
    elevValues[i] = elevation * unitScale
  }

  return elevValues
}
```

pablo-mayrgundter (changes requested):

src/utils/extractHeight.js
pablo-mayrgundter: Ah yeah, you should restore the magic numbers for now, for IFCBUILDINGSTOREY and IFCSIUNIT.  It really should use an import to get those as symbols, but can handle that later

src/utils/extractHeight.js
```diff
@@ -6,23 +11,42 @@
...
+        continue
+      }
+      elevValues[i] = elevation * unitScale
     }
-    return await printStorey()
+
+    return elevValues
   } catch {
-    console.log('No Levels detected')
+    console.log('No levels found')
```
mkeshavarzi, line 50: @pablo-mayrgundter please note`console.warn` caused test failures, hence I changed to `console.log`. I believe the reason of the test failures is that the ifcmodel is often loaded multiple times, before the calculation of the extract levels take place.

src/utils/extractHeight.js
mkeshavarzi: Done

pablo-mayrgundter (changes requested):
Ok, looking good!

Now please include a unit test using a testdata file, e.g. src/utils/testdata/bld-with-units.ifc and verify its heights get correctly interpreted

mkeshavarzi:
> Ok, looking good!
> 
> Now please include a unit test using a testdata file, e.g. src/utils/testdata/bld-with-units.ifc and verify its heights get correctly interpreted

In ifctool we did `fs.readFileSync` to parse the whole test ifc file. Here, if I'm not mistaken, we are using the `ifcModel` loaded by web-ifc, in runtime with different functions calls. Hence, for the tests, which one do you advise?

mkeshavarzi:
> > Ok, looking good!
> > Now please include a unit test using a testdata file, e.g. src/utils/testdata/bld-with-units.ifc and verify its heights get correctly interpreted
> 
> In ifctool we did `fs.readFileSync` to parse the whole test ifc file. Here, if I'm not mistaken, we are using the `ifcModel` loaded by web-ifc, in runtime with different functions calls. Hence, for the tests, which one do you advise?

@pablo-mayrgundter - following up on this matter. Thanks.

pablo-mayrgundter:
I'll take it..

PR #485, opened 2022-11-09
delete viewer mock from src and adjust debug logs

OlegMoshkovich:


pablo-mayrgundter:
I think this is fine, but also reworking the debug stuff in a PR I just sent.

pablo-mayrgundter (approved):

PR #487, opened 2022-11-10
Plane location in URL

OlegMoshkovich:
Add location of a single cut plane to the URL.
Add Toggle General Component.
Refactor Share Control.


src/Components/CutPlaneMenu.test.jsx
```diff
@@ -55,5 +56,38 @@ describe('CutPlane', () => {
...
+    await act(() => {
+      result.current.setViewerStore(viewer)
+    })
+    const callDeletePlanes = viewer.clipper.deleteAllPlanes.mock.calls
+    const callCreatePlanes = viewer.clipper.deleteAllPlanes.mock.calls
+    expect(callDeletePlanes.length).toBe(1)
+    expect(callCreatePlanes.length).toBe(1)
+  })
+
+  it('Plane offset', async () => {
```
pablo-mayrgundter, line 78: Did you mean to test an offset in this?

src/Components/CutPlaneMenu.test.jsx
```diff
@@ -55,5 +56,38 @@ describe('CutPlane', () => {
     expect(callDeletePlanes.length).toBe(1)
     expect(callCreatePlanes.length).toBe(1)
   })
+
+  it('X Section in URL', async () => {
+    render(
+        <ShareMock
+          initialEntries={['/v/p/index.ifc#p:x']}
```
pablo-mayrgundter, line 63: related to below.. might as well test the offset here?

src/utils/cutPlane.js
```diff
@@ -89,3 +90,32 @@ export function removePlanes(viewer) {
...
+    let planeHash
+    const planeOffsetFromModelBoundary = viewer.clipper.planes[0].plane.constant
+    const modelCenter = getModelCenter(ifcModel)
+    for (const [key, value] of Object.entries(viewer.clipper.planes[0].plane.normal)) {
+      if (value !== 0 ) {
+        planeNormal = key
+        planeAxisCenter = modelCenter[planeNormal]
+        planeOffsetFromCenter = planeOffsetFromModelBoundary - planeAxisCenter
+        const planeOffsetFromCenterTrim = planeOffsetFromCenter
+        planeHash = `${planeNormal},${planeOffsetFromCenterTrim}`
```
pablo-mayrgundter, line 115: So I think we want to reserve comma for lists of params in general (like we have in camera), and here when we have multiple planes.

I put up a syntax spec on the wiki:

https://github.com/bldrs-ai/Share/wiki/Design:-URL-Structure#param-syntax

That work?  So change the "," here to "=".

OlegMoshkovich: yep

src/Components/CutPlaneMenu.jsx
```diff
@@ -114,10 +121,147 @@ export default function CutPlaneMenu() {
...
+
+/* eslint-disable no-magic-numbers */
+/**
+ * getSelectionAxisFromBoundingBox is the helper method for the cutplane logic
+ *
+ * @param {object} boundingBox bouding box
+ * @return {object}
+ */
+export function getSelectionAxisFromBoundingBox(boundingBox) {
+  return {
```
pablo-mayrgundter, line 141: assertDefined for bb.min.x, etc.

src/Components/CutPlaneMenu.jsx
```diff
@@ -114,10 +121,147 @@ export default function CutPlaneMenu() {
...
+
+
+/**
+ * getModelCenter return the center of the model based on bounding box
+ *
+ * @param {object} ifcModel bouding box
+ * @return {object} centerCoordinates
+ */
+export function getModelCenter(ifcModel) {
+  return new Vector3(
```
pablo-mayrgundter, line 165: as noted in discussion, i think you can pass in bb and then use the one-liner from three.js to do the calculation

src/Components/CutPlaneMenu.jsx
```diff
@@ -114,10 +121,147 @@ export default function CutPlaneMenu() {
...
+}
+
+
+/**
+ * getElementBoundingBox creates a bounding box around the model
+ *
+ * @param {object} selection seclected meshes
+ * @return {object} boudingBox geometry
+ */
+export function getElementBoundingBox(selection) {
```
pablo-mayrgundter, line 185: per discussion, looks like this isn't being used anymore, so can be deleted

src/Components/CutPlaneMenu.jsx
```diff
@@ -114,10 +121,147 @@ export default function CutPlaneMenu() {
...
+        planeAxisCenter = modelCenter[planeNormal]
+        planeOffsetFromCenter = planeOffsetFromModelBoundary - planeAxisCenter
+        const planeOffsetFromCenterTrim = planeOffsetFromCenter
+        planeHash = `${planeNormal},${planeOffsetFromCenterTrim}`
+      }
+    }
+    const planeOffsetObj = {planeAxis: planeHash}
+    return planeOffsetObj
+  }
+  return null
```
pablo-mayrgundter, line 249: this null return looks a little suspicious.. is it still useful?

src/Components/CutPlaneMenu.jsx
```diff
@@ -114,10 +121,147 @@ export default function CutPlaneMenu() {
...
+}
+
+
+/**
+ * helper method to get the location of all of the planes currently in the scene
+ *
+ * @param {object} viewer
+ * @param {object} model
+ * @return {object} add plane to url
```
pablo-mayrgundter, line 258: Should there be a return value on this method?  I don't see a return value for addHashParams...

pablo-mayrgundter (changes requested):
Couple more comments per discussion

src/Components/CutPlaneMenu.test.jsx
OlegMoshkovich: yes the offset is tested.

src/Components/CutPlaneMenu.test.jsx
OlegMoshkovich: yes of set is tested.

src/utils/cutPlane.js
pablo-mayrgundter: Looks like this still needs to be resolved.

pablo-mayrgundter (approved):

PR #492, opened 2022-11-17
Introduce spreadsheet export

pablo-mayrgundter:
Very very drafty.. even with the element types split out to separate tabs I still find this pretty poor.  I can do some work to cleanup JSON->string values, but I think the sparseness here is a real problem. Would love to see a comparable for Momentum from another tool for inspiration!


MarkusSteinbrecher:
here is how I"heart"ifc did it: https://docs.google.com/spreadsheets/d/1h3BL5Qj_8d81AKsC_2Jd0K-KRFyTbjYa/edit#gid=744899061

Issue #497, opened 2022-12-02
Support cutting-plane orientation based on surface normal


Adrian62D:
Steps to reproduce
1. Navigate to [Eisvogel](https://bldrs.ai/share/v/gh/Swiss-Property-AG/Eisvogel-Public/main/EISVOGEL.ifc#c:107.36,8.46,156.67,3.52,2.03,16.71)
2. With your cursor hover over the roof:
![image](https://user-images.githubusercontent.com/104829208/205259733-dec7de75-5528-4eb2-ac78-8a46119de521.png)
3. Press Q

**Expected**
The cutting plane is created taking into account the surface normal of the roof
![image](https://user-images.githubusercontent.com/104829208/205260471-e44b27ee-b014-4ac8-9fdb-33f91de993d3.png)

**Actual**
The cuttin plane does not take into account the surface normal of the roof
![image](https://user-images.githubusercontent.com/104829208/205259888-1dde7143-2c02-41ae-8787-251667ebbd05.png)

Relates to #332

Adrian62D:
Example from BimCollab:

![image](https://user-images.githubusercontent.com/104829208/205263989-e8f1eb02-aae8-4ff3-9220-435cf94cb026.png)

MarkusSteinbrecher:
There is only xyz cutting planes at the moment, so it works as intended. This would be a feature request I guess.

Adrian62D:
It seems, orientations based on surface normals are already supported for *walls*. See here, it works:

![image](https://user-images.githubusercontent.com/104829208/205270338-b915cae2-e329-4469-a262-6beb863c4041.png)

I renamed the title as feature request.

PR #498, opened 2022-12-02
fix: Support cutting-plane orientation based on surface normal

Ibrahim5aad:
This PR solves #497 .

The issue is that the clipper component of the IFC viewer has a flag `orthogonalY` that controls whether to lock the cut plane normal vector to the y-axis if it exceeds a certain tolerance `toleranceOrthogonalY`. This flag default value is true, which was causing the issue.


pablo-mayrgundter (approved):
Nice :)  Thanks so much!

PR #504, opened 2022-12-08
Update types for location

oo-bldrs:
PR to update the location handling functions to use the Location type.

This uses the type and shape from the history react-router package, since the native DOM type is immutable (which prevents instantiation for testing).

It's not a big deal since the input into these functions is the location object emitted from react-router.

### DIFF
--- a/src/utils/location.js
+++ b/src/utils/location.js
@@ -1,4 +1,6 @@
+/** @type {Object<string, Function>} */
 const hashListeners = {}
+
 window.onhashchange = () => {
   for (const name in hashListeners) {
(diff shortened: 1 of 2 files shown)


pablo-mayrgundter:
Looks good, but wanted to check what these are:

image

oo-bldrs:
@pablo-mayrgundter I think I mixed in some TypeScript syntax in my pull request. I'll clean it up and move it out of draft status.

pablo-mayrgundter:
ping

OlegMoshkovich:
ping

Issue #503, opened 2022-12-08
ControlButton - onclick setIsDialogDisplayed error


howkymike:
### Description 
The ControlButton's  `setIsDialogDisplayed` parameter is a React state setter function, but it is used as HTML onClick function which provides a following error:
```
Warning: Failed prop type: Invalid prop `selected` of type `object` supplied to `ForwardRef(ToggleButton)`, expected `boolean`.
    at ToggleButton (https://bldrs.ai/index.js:227719:17)
    at ControlButton (https://bldrs.ai/index.js:228120:3)
    at AboutControl (https://bldrs.ai/index.js:228183:81)
    at div
    at https://bldrs.ai/index.js:26850:31
    at ButtonGroup (https://bldrs.ai/index.js:206335:17)
    at div
    at OperationsGroup (https://bldrs.ai/index.js:231771:3)
```

### Step to reproduce
1. Go to [bldrs.ai](https://bldrs.ai/)
2. Click "About" icon

## Potencial bug fix
In the Buttons.jsx change ToggleButton's onClick param to
```
onClick={() => setIsDialogDisplayed(true)}
```

OlegMoshkovich:
Cool - thank you so much for the issue. 
I think I know where it is coming from.. we are passing in a state variable into the dialog that gets misinterpreted.

PR #507, opened 2022-12-14
Fix ControlButton - onclick setIsDialogDisplayed error

rondoor124:
The issue is because onClick event was fired instead of boolean.
This will fix this error: https://github.com/bldrs-ai/Share/issues/503
image
image


pablo-mayrgundter (approved):
Thank you!

PR #509, opened 2022-12-16
Organization

rondoor124:
Hi Pablo, I'm making new PR since I moved repo to dev account.

The completed tickets are:
- Organize Components folder.
- Upgrade style from @mui/styles to sx.

I'll complete the following tickets asap:
- Run new tsconfig on @jsdoc and improve the doc types.
- Add some testing where it's lite.

Please review current result, and give me feedback.
Thank you.


pablo-mayrgundter (changes requested):
Ok, looking better!

1) Conversion to SX is looking good.  I'll need to do a careful pass through the netlify preview.. more on that later.

2)  Still probably need to address some overall style issues first and I'm not sure they're worth it right now.  But let's see...  also, maybe double-check your eslint config, these should be showing up as errors for you:
  - prefer ' over "
  - double space between functions (i see you flattened some to single)
  - prefer UNIX line breaks over Windows
  - omit trailing ;
  - no space before self-closing tag: <Tag /> -> <Tag/>
  
3) Nice to see the new component folders!  For the names, let's use the same name as the component, so
  - notes -> Notes
  - item.properties -> ItemProperties
  - side.drawer -> SideDrawer

pablo-mayrgundter:
Also seeing one style breakage.. the user icons lost their sizing:

image

PR #510, opened 2022-12-18
Introduce tsc for some whitelisted jsdoc checks in src/utils.

pablo-mayrgundter:
Also add a dependency diagram generator and check the first graph in.

To use this, just run use `yarn lint` as usual.  It will now include a jsdoc type check for the files listed in the tsconfig.


tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "./src/utils/location.test.js",
+    "./src/utils/math.js",
+    "./src/utils/math.test.js",
+    "./src/utils/objects.js",
+    "./src/utils/strings.js",
+    "./src/utils/strings.test.js",
+  ],
+  "compilerOptions": {
+    "incremental": true, /* Save .tsbuildinfo files to allow for incremental compilation */
+    "outDir": "./built",
```
oo-bldrs, line 16: Since it's just soft checks, we should disable emitting files from compilation.

This should be replaced with the `noEmit` option.

tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "./src/utils/location.js",
+    "./src/utils/location.test.js",
+    "./src/utils/math.js",
+    "./src/utils/math.test.js",
+    "./src/utils/objects.js",
+    "./src/utils/strings.js",
+    "./src/utils/strings.test.js",
+  ],
+  "compilerOptions": {
+    "incremental": true, /* Save .tsbuildinfo files to allow for incremental compilation */
```
oo-bldrs, line 15: Given the provisional state we're in, the low percentage of the project being monitored, and how the foundational elements can and will change, do we really want incremental compilation?

tsconfig.json
pablo-mayrgundter: Ah cool.. and i think not using incremental mode as well?

pablo-mayrgundter: Looks like it may be slowing down tsc a bit.  But works for now if that's good with you

tsconfig.json
pablo-mayrgundter: Yep, removed

pablo-mayrgundter:
PTAL

src/utils/location.test.js
```diff
@@ -1,3 +1,6 @@
+// TODO(pablo): The type for loc is really verbose below.  Gotta be a
```
oo-bldrs, line 1: `Location` by default, is a type that is sourced from the DOM. That type is immutable and does not allow you to create it as it has no constructor.

If we were in pure TypeScript mode, we could use a type that implements the DOM's interface. Instead, in my draft PR, I create an object that mimics the members' types and a function to return new instances of it.

Then within the tests I manipulate the Location-like object for my test subjects.

https://github.com/bldrs-ai/Share/pull/504/files#diff-11e2a89e0bb7478d9e0add3e59943227c24b3d70c93ec5339007422ff993d335

src/utils/strings.test.js
```diff
@@ -34,4 +34,20 @@ describe('strings', () => {
       'http://localhost:8080/share/v/p/index.ifc#c:-26.91,28.84,112.47,0,16.21,-3.48',
     ])
   })
+
+
+  it('isNumeric recognizes common cases', () => {
```
oo-bldrs, line 39: While there's nothing listed in the style guide for this, I'm not a fan of functions have have a lot of linter exclusions.

To put a number on it, this has 1 exclusion for every 2 lines of code. I would argue that this high ratio warrants that this file should not be included in compilation.

.gitignore
```diff
@@ -4,6 +4,9 @@
 # no longer check-in docs for builds.  This is handled by Netlify
 /docs
 
+# tsc checks build here
```
oo-bldrs, line 7: With the introduction of `noEmit: true`, this pattern addition is no longer necessary.

oo-bldrs:
Can we not exclude the multi-megabyte output of the dependency graph results so they're not committed to the repository?

src/utils/location.test.js
pablo-mayrgundter: Done.. used chatgpt saved me on the typedef!

src/utils/strings.test.js
pablo-mayrgundter: Disagree.  This was defensive for being called with numbers when strings are expected.

But!  Turns out we only use it once, so will turn it into an exception and not test the numeric arg.

> 

yep, didn't realize it was that big.  SVG is more useful too and 10x smaller.  Removed from checkin

pablo-mayrgundter:
PTAL

pablo-mayrgundter (commented):
PTAL

oo-bldrs (approved):
LGTM

Issue #314, opened 2022-08-04
Notes not updating if link to ifc on Github is pasted into search


MarkusSteinbrecher:
Go to Bldrs homepage, open notes and then paste a link to an ifc into the search ... new ifc model is loaded but the notes do not update:

![image](https://user-images.githubusercontent.com/74647806/182809273-c12d0379-570b-4c4e-a61b-1ea20cbc8b09.png)

OlegMoshkovich:
yes. I think it is happening because we are not triggering issue pull on the model load.

OlegMoshkovich:
Problem: it is happening because the user loads the new project while on the "SINGLE NOTE" view in the current project, 
When the new project is loaded the specific note cannot be located.

Proposed solution: Change the "SINGLE NOTE" view to the "LIST of NOTES" view automatically when the new project is loaded.

PR #519, opened 2022-12-21
add guide link to the about panel

OlegMoshkovich:
Add guide link to the about panel which points to the guide in the notes.


pablo-mayrgundter (commented):
Hey, I like this idea, but can't read it myself.. too small.  I'm also noticing the other text in the about has become 0.9.. I can read it but not great.

Trying to think how to make the text in the graphic responsive.  Can you make a lookalike SVG in Figma?  We could also style it according to theme then

OlegMoshkovich:
If you click on the graphic, it is suppose to take you to the note where all of the features are explained in more detail.

OlegMoshkovich:
This is also a first step towards a better solution.

OlegMoshkovich:
@pablo-mayrgundter individual note sharing was broken, fixed now - which will enabling opening a guide now when the guide graphic is clicked.

pablo-mayrgundter:
Heya, see my comment on that PR.. it fixed the individual note sharing, but then re-broke, #314

PR #525, opened 2022-12-21
VYZN: Viewer: UI configuration via Matrix config message

### LINKED ISSUE #391: VYZN: Viewer: UI configuration via Matrix config message
Regarding the viewer itself: 
allow to hide or show different UI elements, such as the tree view, the share buttons, the search bar…

Available programmatically via app.

Optional: Let the user customize this if the app allows it.

Fits to product strategy: probably
Estimation: 5

### LINKED ISSUE #515: VYZN: Matrix Widget API: Improve debouncing pattern
The current debouncing pattern should be improved as follows:

- Have a single SelectionChanged message instead of selection/deselection
- The SelectionChanged message should be emitted regardless of the trigger
  - Currently the SelectionChanged message is not emitted if the selection was done via message instead of an actual user

### LINKED ISSUE #516: VYZN: Matrix Widget API: Improve how model-loading is covered in automated tests
- Add a new ModelLoaded message that is emitted when the model has been loaded
- From the automated tests, supply a local resource of an .ifc model instead of a network fetch to be independent of the network

### LINKED ISSUE #517: VYZN: Matrix Widget API: make selection of elements in tests more reliable
- Currently the selection of elements in automated tests are based on positions ("center")
- It would be more reliable to perform the selection via the tree-view on the left or other methods
- In order to do this, the opening of the tree-view needs to be guaranteed (currently it does not open up 1 of 5 times)

Ibrahim5aad:
This PR refactors the underlying classes for the widget API communication mechanism that allows the Share app to be treated as an iframe-embeddable app that supports receiving and sending messages back and forth with the embedding apps. 

The main theme for this PR is supporting #391 **UI configuration** via this previously mentioned widget API mechanism.
 
The PR also stabilizes the e2e tests for this API integration and adds the required test for the new functionality **UI configuration**.

09/01 Update:  Closes #391, closes #515, closes #516, closes #517.


Ibrahim5aad:
It's weird that the load-sample-model e2e test pass for me locally, but "sometimes" fail when triggered by the github action.

pablo-mayrgundter:
> It's weird that the load-sample-model e2e test pass for me locally, but "sometimes" fail when triggered by the github action.

Hi Ibrahim!  Looks good overall.  We have a fix underway for the e2e test.. will try to get it in tomorrow.  Hopefully that fixes the issue for you too.

One high-level comment on the UI config.  I'm not sure it makes sense to have individual button enable toggles.  Do you guys have use-cases for that?  I'll go back to the bug for more discussion...

Ibrahim5aad:
@pablo-mayrgundter Hi Pablo! I modified the implementation to be more coarse in hiding/showing UI components. (per button group). Is this making more sense now?

Ibrahim5aad:
This now should close #391, #515, #516, #517.

src/Components/OperationsGroup.jsx
```diff
@@ -37,6 +37,12 @@ export default function OperationsGroup({
   const selectedElement = useStore((state) => state.selectedElement)
   const colorMode = useContext(ColorModeContext)
 
+  const isCollaborationGroupVisible = useStore((state) => state.isCollaborationGroupVisible)
+  const isModelInteractionGroupVisible = useStore((state) => state.isModelInteractionGroupVisible)
+  const isSettingsVisible = useStore((state) => state.isSettingsVisible)
+
+  const isFirstDividerVisible = useStore((state) => state.getFirstDividerVisiblility)
```
OlegMoshkovich, line 44: Dividers should be a part of the group, so we do not need to explicitly control them.

Ibrahim5aad: Users/Third-parties don't have an explicit control over them. Their visibility is derived from the visibility of the surrounding groups, because if they belonged to the group you will have visual situations where one group is visible (and the divider as well if it belonged to the group) but the divider should be invisible in that case.

pablo-mayrgundter:
Hi Ibrahim!  Thanks for waiting to get #455 in.  Could you merge and we'll go from there?  Thanks.

> @pablo-mayrgundter Hi Pablo! I modified the implementation to be more coarse in hiding/showing UI components. (per button group). Is this making more sense now?

Yes, great, thanks!

OlegMoshkovich:
@Ibrahim5aad can you please join our discord?
https://discord.gg/BxHA5TBe

Ibrahim5aad:
Hi @pablo-mayrgundter, conflicts are fixed, and this PR is ready to get through now.

src/WidgetApi/Utils.js
```diff
@@ -0,0 +1,77 @@
...
+
+  // @source https://github.com/30-seconds/30-seconds-blog/blob/master/blog_posts/javascript-array-comparison.md
+  /**
+   * check if two arrays are equal.
+   *
+   * @param {object[]} a
+   * @param {object[]} b
+   * @return {boolean}
+   */
+  arraysAreEqual(a, b) {
```
pablo-mayrgundter, line 63: Hm, this name is confusing given the use case.

This is a common enough utility for the whole project.  Would you please move it into src/utils/sets.js

Also I think it's worth separating out into a primary and helper
```
/**
 * @param {Array} a
 * @param {Array} b
 * @return {boolean} if sets are the same
 */
export function setsFromArraysAreEqual(a, b) {
  return setsAreEqual(new Set(a), new Set(b))
}

/**
 * @param {Set} a
 * @param {Set} b
 * @return {boolean} if sets are the same
 */
export function setsAreEqual(a, b) {
  // https://stackoverflow.com/questions/31128855/comparing-ecma6-sets-for-equality
  return a.size === b.size && [...a].every((x) => b.has(x))
}
```

Here's a preferable arraysAreEqual if that's what you want:
```
arraysAreEqual(a, b) {
  if (a.length !== b.length) {
    return false
  }
  for (let i = 0, n = a.length; i < n; i++) {
    if (a[i] !== b[i]) {
      return false
    }
  }
  return true
}
```

pablo-mayrgundter (changes requested):
Ok, looking pretty good.  One refactor in Utils.

Also, let's avoid including Momentum (5MB) as a test asset.  Can you use a constructed IFC to get what you need?  Here's a minimal IFC file to start with:

https://discord.com/channels/853953158560743424/945373774402945074/1048365704077910096

src/WidgetApi/Utils.js
Ibrahim5aad: Yeah you are right, the element ids lists aren't expected to have duplicates at the end. But you provided implementation requires the two arrays to be of the same order to consider equal, which is not what we seek here.

Ibrahim5aad: @pablo-mayrgundter, I introduced a minimized/refactored implementation that should cover the current use case of comparing lists of element ids.

Ibrahim5aad: "Momentum" model are removed from e2e tests as well.

PR #528, opened 2022-12-22
fix individual note sharing

OlegMoshkovich:
This PR fixes individual note sharing via permalink, this https://deploy-preview-528--bldrs-share.netlify.app/share/v/p/index.ifc#c:-111.37,14.94,90.63,-43.48,15.73,-4.34::i:1493510953 should open to an individual note.

### DIFF
--- a/src/Components/notes/Notes.jsx
+++ b/src/Components/notes/Notes.jsx
@@ -16,7 +16,6 @@ export const NOTE_PREFIX = 'i'
 export default function Notes() {
   const classes = useStyles()
   const selectedNoteId = useStore((state) => state.selectedNoteId)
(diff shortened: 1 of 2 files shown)


pablo-mayrgundter (changes requested):
Looks like this brings back the problem noted in the comment.  I tried reproducing the bug in #314 on your demo and it's back :(

pablo-mayrgundter:
> Looks like this brings back the problem noted in the comment. I tried reproducing the bug in #314 on your demo and it's back :(

Whoops, I was pasting in a bldrs.ai link, not a github link

OlegMoshkovich:
Hmm works for me when I open to the individual note and then load a different model.

pablo-mayrgundter (approved):

PR #530, opened 2022-12-22
Fix render bug of h2 nested inside of h1.

pablo-mayrgundter:
Just saw this in prod.. one-liner fix.  New size looks fine to me as well.

image


oo-bldrs (approved):
LGTM

PR #543, opened 2023-01-13
Create IFC test helper functions

oo-bldrs:
PR to introduce some basic helper functions to create parts of an IFC project tree structure.

Perhaps this should go into a dedicated helper directory/structure in the future as we accumulate more items like this.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -30,6 +30,7 @@ const esModules = [
   'unist-util-position',
   'unist-util-stringify-position',
   'unist-util-visit',
+  'uuid',
   'vfile',
   'vfile-message',
   'web-ifc',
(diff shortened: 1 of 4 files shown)


pablo-mayrgundter (approved):
Thanks, very useful!

One concern, nanoid looks cool but isn't quite the same UUID as IFC

https://technical.buildingsmart.org/resources/ifcimplementationguidance/ifc-guid/

We already have the uuid pkg in our node_modules, so maybe just use it?

oo-bldrs:
@pablo-mayrgundter Thanks for the heads up.

That explains why in IFC project files I saw something that didn't resemble a UUID, but I incorrectly assumed it was a different library.

In the future, I'll look into creating a small library that generates values that align with the IFC-GUID standard.

PR #545, opened 2023-01-15
VYZN: Feat: rendering of individual elements meshes (subset ops)

### LINKED ISSUE #518: VYZN: Rendering of individual nodes
Technical changes:
- Currently the rendering happens per-model 
- This needs to be changed so that individual nodes can be rendered

This is a precondition to #384, #475, #417

We'll provide a more detailed assessment on this issue.

Ibrahim5aad:
This PR closes #518 by the simplest way available without breaking too much of the IFC.js out of the box abstractions.

This PR uses the IFC.js subset operations to break down the spatial structure of the model into subsets and instead of rendering the IFC model combined mesh, it renders every subset instead. Although the combined model mesh is not rendered, it is kept in memory for other dependencies that consume it.


Ibrahim5aad:
I'm drafting this for now, the provided implementation gives a big hit to performance under big models.

Adrian62D:
@Ibrahim5aad Can we discard this one or is it still relevant?

Ibrahim5aad:
> @Ibrahim5aad Can we discard this one since it probably has been superseeded by #617?

Yes this can be discarded.

Issue #544, opened 2023-01-13
MISC: Use same logic for coordinate truncation in CutPlane as we do in CameraControl


pablo-mayrgundter:
See:
  https://github.com/bldrs-ai/Share/blob/main/src/Components/CutPlaneMenu.jsx

There we are using full precision floating point, but should use only a few digits like in CameraControl:

https://github.com/bldrs-ai/Share/blob/main/src/Components/CameraControl.jsx#L121

Also, it looks like both of these should be refactored to use some pure utility functions and reduce their redundancy.

pablo-mayrgundter:
e.g., camera is rounded to 2 decimals, cutplane not rounded:

https://bldrs.ai/share/v/p/index.ifc#c:-111.37,14.94,90.63,-43.48,15.73,-4.34::p:x=-14.125452060112806

And actually, the unit here is meter, and I've been telling people we use millimeter accuracy, but apparently that's not true! :-(

Both should be rounded to 3 digits: 12.345.

If not too hard, trailing zeros should be omitted, so 12.000 should simply be 12, 12.100 -> 12.1, etc.

pablo-mayrgundter:
Fixed in #547

PR #546, opened 2023-01-16
Use same `stof` logic in both `CutPlaneMenu` and `CameraControl`

rondoor124:
Hi @pablo-mayrgundter, I finished issue https://github.com/bldrs-ai/Share/issues/544.
Please check it and give me feedback.

### DIFF
--- a/src/utils/strings.js
+++ b/src/utils/strings.js
@@ -58,3 +58,21 @@ export function findUrls(str) {
     }
   })
 }
+
+/**
+ * Convert string to finite float.
+ *
+ * @param {string} str Input string.
+ * @return {number} float.
+ */
+export function stof(str) {
+  const floatDigits = 2
+  const val = parseFloat(parseFloat(str).toFixed(floatDigits))
+  if (isFinite(val)) {
+    const rounded = parseFloat(val.toFixed(0))
+    return rounded === val ? rounded : val
+  } else {
+    console.warn('Invalid coordinate: ', str)
+  }
+  return val
+}
--- a/src/utils/strings.js
+++ b/src/utils/strings.js
@@ -71,8 +71,6 @@ export function stof(str) {
   if (isFinite(val)) {
     const rounded = parseFloat(val.toFixed(0))
     return rounded === val ? rounded : val
-  } else {
-    console.warn('Invalid coordinate: ', str)
   }
   return val
 }
--- a/src/utils/strings.js
+++ b/src/utils/strings.js
@@ -59,18 +59,27 @@ export function findUrls(str) {
   })
 }
 
+
 /**
- * Convert string to finite float.
+ * Convert a string-encoded float to a truncated float, of fixed-length `len` or no decimal point expansion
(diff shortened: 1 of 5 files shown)


src/utils/strings.js
```diff
@@ -58,3 +58,21 @@ export function findUrls(str) {
...
+ * @param {string} str Input string.
+ * @return {number} float.
+ */
+export function stof(str) {
+  const floatDigits = 2
+  const val = parseFloat(parseFloat(str).toFixed(floatDigits))
+  if (isFinite(val)) {
+    const rounded = parseFloat(val.toFixed(0))
+    return rounded === val ? rounded : val
+  } else {
```
pablo-mayrgundter, line 74: Now that this is a util, it should throw instead of warn.

Please do that and update jsdoc as well.

src/utils/strings.js
```diff
@@ -58,3 +58,21 @@ export function findUrls(str) {
...
 }
+
+/**
+ * Convert string to finite float.
+ *
+ * @param {string} str Input string.
+ * @return {number} float.
+ */
+export function stof(str) {
+  const floatDigits = 2
```
pablo-mayrgundter, line 69: pls move this up to a param `len=3`; I added a comment asking for the default to be changed from 2 to 3:

https://github.com/bldrs-ai/Share/issues/544

src/utils/strings.js
```diff
@@ -58,3 +58,21 @@ export function findUrls(str) {
     }
   })
 }
+
+/**
+ * Convert string to finite float.
```
pablo-mayrgundter, line 63: I think this is a little more specific.  Here's an updated API.   Please also add unit tests for these and some boundary conditions with NaN, etc.

```
/**
 * Convert a string-encoded float to a truncated float, of fixed-length `len` or no decimal point expansion
 * if equal to zero.  e.g. with `len=3`:
 *  - '12.34567' -> 12.345
 *  - '12.340' -> 12.34 // same with '12.34'
 *  - '12.300' -> 12.3   // same with '12.3'
 *  - '12.000' -> 12 
 */

File: src/utils/strings.js [hunk 4]
```diff
@@ -58,3 +58,21 @@ export function findUrls(str) {
...
+ *
+ * @param {string} str Input string.
+ * @return {number} float.
+ */
+export function stof(str) {
```
pablo-mayrgundter, line 68: pls rename `floatStrTrim`

pablo-mayrgundter (changes requested):

src/utils/strings.js
rondoor124: i have done.

src/utils/strings.js
rondoor124: i have done.

src/utils/strings.js
rondoor124: i have done.

pablo-mayrgundter (approved):

PR #557, opened 2023-01-20
Use explicit imports for MUI icons

oo-bldrs:
PR to use an explicit import for individual MUI icons.

Without this change, Storybook builds take an incredibly long time on Netlify as it bundles every, single, icon. This results in 12K+ ES6 modules being bundled. This causes build time to explode, especially if we get unlucky and land on a slow build agent.

After this change, the total number of ES6 modules bundled chops down to ~2,800.


pablo-mayrgundter (approved):
Oh wow, good catch!  Guess it works like mui imports for individual classes.  Good to know.

Issue #148, opened 2022-03-15
Misc: Make properties window adjustable


MarkusSteinbrecher:
The properties window has a fixed size and some content is not showing. Can we make the size adjustable (by dragging it to the left)?

![image](https://user-images.githubusercontent.com/74647806/158375595-bcca7902-faa1-4319-a49b-53f9aa3545e0.png)

pablo-mayrgundter:
Like this would be very cool:

https://user-images.githubusercontent.com/2480879/214316542-26046fe6-b790-44d8-8f52-d51226f7928a.mov

PR #564, opened 2023-01-26
Apply adjustable size to the sidebar

rondoor124:
Hi @pablo-mayrgundter.
I made adjustable size feature to the sidebar, and it required many update of the code.
Please check the issue https://github.com/bldrs-ai/Share/issues/148 and give me feedback.
Thank you.


pablo-mayrgundter:
w00t!  Super cool :)

Per discussion, please remove the new borders and revert the background color changes, and see if you can find a button for the vertical separator to help the user discover the draggable behavior.

image

pablo-mayrgundter:
Also I think the margin for the buttons got shrunk:

image

here's the correct margin from production:

image

rondoor124:
`PTAL`

rondoor124:
`PTAL`

OlegMoshkovich:
super cool!!

pablo-mayrgundter:
Looking good!

Please use the theme for background color on the 3 dots.. it's currently white, which is fine, but might as well use the theme :)

For testing, I think there's 1 new behavior here, which is that close and open of the panel keeps the resize width.  See if you can add an expect unit test for that.

rondoor124:
When closing the panel, the system hide only the sidebar panel, not removing, and keep the sidebar width using state, so i think it's no useful to check the panel width when toggling. :)

rondoor124:
@OlegMoshkovich, thank you for your kind feedback. :)
yep, i think it'd be better to use the carrot to collapse and expand the below items fully instead of reducing the height to half, would like to know your suggestion. :)

OlegMoshkovich:
The point of the collapsed half drawer is to enable the user to access notes and the viewport simultaneously - it is useful to have both as long as the user is able to scroll the notes while in the drawer is in the collapsed state.

OlegMoshkovich:
Would be nice if you can figure out how to reduce the space at the top of the drawer.
image

This is the example sketch of the collapsed drawer.
![Uploading 215061494...]

pablo-mayrgundter:
Guys, let's keep this focused on the side drag.  I'm going to merge

PR #567, opened 2023-01-27
add undefine check for get properties

OlegMoshkovich:
This PR solves #554 properties undefined bug.

### DIFF
--- a/src/utils/itemProperties.jsx
+++ b/src/utils/itemProperties.jsx
@@ -164,7 +164,7 @@ export async function unpackHelper(model, eltArr, serial, ifcToRowCb) {
           throw new Error('Array contains non-reference type')
         }
         const refId = stoi(p.value)
-        if (model.getItemProperties) {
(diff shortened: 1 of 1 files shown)


src/utils/itemProperties.jsx
```diff
@@ -164,7 +164,7 @@ export async function unpackHelper(model, eltArr, serial, ifcToRowCb) {
           throw new Error('Array contains non-reference type')
         }
         const refId = stoi(p.value)
-        if (model.getItemProperties) {
+        if (model.getItemProperties && model.getItemProperties !== 'undfined') {
```
pablo-mayrgundter, line 167: Hey, this doesn't look right.  There's a typo in undefined, but moreover, if fixed the right side will always return true if it is evaluated.  Does this really fix the issue?

OlegMoshkovich: not really sure how to reproduce.
will turn it to a draft for now.

Issue #550, opened 2023-01-17
VYZN: Matrix Widget API: Highlighting of one or more elements programmatically


Adrian62D:
One or more elements of the model can be highlighted by an app programmatically via Matrix Widget API.

This means the element "shines through" even if it is occluded. This allows a user to see an element even if it is inside a building.

This provides programmatic access to the functionality of #382.

Adrian62D:
@aozien This can be closed, right?

Issue #382, opened 2022-09-15
VYZN: Viewer: Highlighting of one or more elements by the user.


Adrian62D:
One or more elements of the model can be **highlighted** by the user.

Programmatic access is separately defined in #550.

MarkusSteinbrecher:
@Adrian62D how is this different to #475?

Adrian62D:
#475 is about allowing a user to actually filter through the elements.

In BimCollab this functionality is called SmartViews:
https://www.bimcollab.com/en/products/bimcollab-zoom/smart-views

Adrian62D:
@aozien This can be closed, right?

PR #577, opened 2023-02-03
VYZN: Highlight selection inside a model when its occulded 

### LINKED ISSUE #454: VYZN: Viewer: Highlighting elements inside a model need to be visible
If I select an element that is inside a model, e.g. a certain room or inside wall, the element highlight is not visible because it is blocked by other elements. Highlighting elements should "shine through" somehow, similar to how Blender does it:

image

aozien:
**This PR closes #454** 
Update: it should also close **#382** and **#550** as well

It's achieved by using a PostProcessing effect to the scene after the initial render pass, I had to override [the render function ](https://github.com/IFCjs/web-ifc-viewer/blob/f8611ef91ba2081959e6531eb98f8c8cec72e14f/viewer/src/components/context/renderer/renderer.ts#L44) created by the IFC-Viewer to be able to create custom render pipeline to avoid duplicate rendering on the scene, as specified [here](https://threejs.org/docs/#manual/en/introduction/How-to-use-post-processing) 

> When using a composer, it's necessary to change the application's animation loop. Instead of calling the render method of [WebGLRenderer](https://threejs.org/docs/index.html#api/en/renderers/WebGLRenderer), we now use the respective counterpart of [EffectComposer](https://threejs.org/docs/index.html#examples/en/postprocessing/EffectComposer).

## how it works
initializing **IFCHighlighter** overrides the (viewer render function) to use the composer render instead, then the highlighter can set what's highlighted based on the current selected Meshes from **IFCSelection**

image


src/Infrastructure/IfcHighLighter.js
```diff
@@ -0,0 +1,62 @@
+
```
pablo-mayrgundter, line 1: No empty line

src/Infrastructure/IfcHighLighter.js
```diff
@@ -0,0 +1,62 @@
...
+    // override the viewer rendering pipeline
+    const renderer = context.getRenderer()
+    const scene = context.getScene()
+    const camera = context.getCamera()
+    const {composer, outlineEffect} = createComposer(renderer, scene, camera)
+    this._outlineEffect = outlineEffect
+    context.renderer.update = newUpdateFunction(context, composer)
+  }
+
+  /**
```
pablo-mayrgundter, line 30: 2 empty lines between functions (and throughout)

pablo-mayrgundter (approved):
Neat!

Needs 2 unit tests for selection/deselection expects to the jest mocks.  And a couple formatting nits.

Thanks!

aozien:
Hi @pablo-mayrgundter, added tests for highlighting, removed empty lines, PTAL

PR #579, opened 2023-02-06
Remove material-ui/core and upgrade MUI packages.

pablo-mayrgundter:
Fix for #325


pablo-mayrgundter:
Ah storybook uses it and looks like it's currently broken.  Will get back to this later

PR #591, opened 2023-02-09
Fixes #590 deselection bug, added null check

### LINKED ISSUE #590: Bug: Deselect button does not work
Steps to reproduce:
1. Select the element
2. Click deselect "x" button
3. Element stays selected

aozien:
Fixes #590 deselection bug, added null check


pablo-mayrgundter:
Thanks, just saw this too.  Easy fix!

PR #595, opened 2023-02-12
Overriding IfcElements render colors on loading file

aozien:
### The Idea
This is an alternative more of a static approach for colorizing single IfcElements at an early stage of file loading, without having to rely on the ifc-viewer subset functionality.

The viewer.loadIfcUrl now takes an extra parameter where you can specify options for overloading the loaded elements colors.

### The view-options is an object that contains
1. a map of key-value pair for elements and their colors.
2. a default color to be used for element as a fallback
3. if both failed, it will use the element's original color.

**This PR should be the first step of achieving the 'smart views' like functionality
The second step is to create the view-rules compiler**

Eventually those view settings will be the result of compiling the view-rules against the loaded model, 

for example having a view-rule  

> type:Wall set color:red; type:Slab set color:#0000FF;

will first be executed on the model file in the compiler that creates the viewOptions object to be passed to the viewer

image
image

### DIFF
--- a/__mocks__/web-ifc-viewer.js
+++ b/__mocks__/web-ifc-viewer.js
@@ -58,6 +58,11 @@ const impl = {
         material: null,
       },
     },
+    loader: {
+      ifcManager: {
+        parser: {},
+      },
+    },
   },
   clipper: {
     active: false,
@@ -88,6 +93,7 @@ const impl = {
   }),
   setSelection: jest.fn(),
   pickIfcItemsByID: jest.fn(),
+  loadIfcUrl: jest.fn(jest.fn(() => loadedModel)),
 }
 const constructorMock = ifcjsMock.IfcViewerAPI
 constructorMock.mockImplementation(() => impl)
--- a/src/Containers/CadView.jsx
+++ b/src/Containers/CadView.jsx
@@ -251,7 +251,7 @@ export default function CadView({
     setIsLoading(true)
 
     const ifcURL = (uploadedFile || filepath.indexOf('/') === 0) ? filepath : await getFinalURL(filepath, accessToken)
-    const loadedModel = await viewer.IFC.loadIfcUrl(
+    const loadedModel = await viewer.loadIfcUrl(
         ifcURL,
         !urlHasCameraParams(), // fitToFrame
         (progressEvent) => {
--- a/src/Infrastructure/IfcColor.js
+++ b/src/Infrastructure/IfcColor.js
@@ -1,0 +1,14 @@
+/**
+ * create an ifc color object
+ *
+ * @param {number} r the red component of the color as a fraction between 0 and 1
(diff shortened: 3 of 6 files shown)


pablo-mayrgundter:
Hi @aozien , sorry, missed this again.  Are you able to ask for a reveiw?  If not I can move you to outside collaborator with triage

pablo-mayrgundter:
How are you previewing this with the colors?  Would be nice to have a link to preview

aozien:
Hi @pablo-mayrgundter, no problem there's also another PR open for the multiselect in case you missed it.

> Are you able to ask for a reveiw?

I'm not sure where this option should be, but I'm not seeing it any where.


> How are you previewing this with the colors? Would be nice to have a link to preview

The [viewer.loadIfcUrl](https://github.com/aozien/Share/blob/ae683d798ad6eb2733d8f845b2f4790c508be0a5/src/Containers/CadView.jsx#L254) function now accepts an additional parameter of type [IfcCustomViewSettings](https://github.com/aozien/Share/blob/overriding-model-colors/src/Infrastructure/IfcCustomViewSettings.js) 

Where you can define the following when loading the file:
1- a default color to override all elements color with
2- a specific color per express Id


something like this for example:

image

can be achieved with the following settings:

```   
import IfcCustomViewSettings from '../Infrastructure/IfcCustomViewSettings'
import IfcColor from '../Infrastructure/IfcColor'
.
.

 
 const customViewSettings = new IfcCustomViewSettings(new IfcColor(0.7, 0.3, 0.1, 1), {
      396: new IfcColor(newFunction(), 0.3, 0.4, 0.8),
 })
const loadedModel = await viewer.loadIfcUrl(
ifcURL,
!urlHasCameraParams(), // fitToFrame
(progressEvent) => {
  if (Number.isFinite(progressEvent.loaded)) {
    const loadedBytes = progressEvent.loaded
    // eslint-disable-next-line no-magic-numbers
    const loadedMegs = (loadedBytes / (1024 * 1024)).toFixed(2)
    setLoadingMessage(`${loadingMessageBase}: ${loadedMegs} MB`)
    debug().log(`CadView#loadIfc$onProgress, ${loadedBytes} bytes`)
  }
},
(error) => {
  console.warn('CadView#loadIfc$onError', error)
  // TODO(pablo): error modal.
  setIsLoading(false)
  setAlertMessage(`Could not load file: ${filepath}`)
}, customViewSettings)
    
   ```

pablo-mayrgundter:
I'm sorry i was taking a while on this this week.  I'd like to get you guys in a better previewing situation.. i didn't quite follow how to evaluate this.

However, I do understand the goal here and the code as you have it doesn't touch any main workflow, so fine for now.

I'm going to merge this and see what else I we can get in this week.

We should talk soon.  I'll send an invite for a meeting after I get a chance to talk with Adrian.

Cheers!

Issue #541, opened 2023-01-12
VYZN: Matrix Widget API: Selection of one or more elements programmatically


Adrian62D:
One or more elements of the model can be selected programmatically via Matrix Widget API.

This programmatic access should allow for the following use cases:

- Select
- Deselect
- Select All
- Clear All

This corresponds to the actions in the UI as documented in #380.

PR #597, opened 2023-02-13
Viewer: Selection of one or more elements via tree view

### LINKED ISSUE #568: VYZN: Viewer: Selection of one or more elements via tree view
This is an extension of #380.

image

Behavior:
- Shift+ click on a tree view node to add add it to the selection
- Shift+ click on a already selected tree view node to remove it from the selection

aozien:
**This PR closes #568**

This adds the capability to select multiple elements from the side tree view

### DIFF
--- a/src/Components/NavPanel.jsx
+++ b/src/Components/NavPanel.jsx
@@ -24,6 +24,7 @@ export default function NavPanel({
   defaultExpandedElements,
   expandedElements,
   setExpandedElements,
+  selectWithShiftClickEvents,
   pathPrefix,
 }) {
   assertDefined(...arguments)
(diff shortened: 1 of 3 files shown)


pablo-mayrgundter (changes requested):
Nice!

Please add a unit test in NavTree.test.jsx.

This is a nice example of should be easy to test, but also easy to break without realizing it, so good to start testing.

aozien:
Hi @pablo-mayrgundter , I added a test for the NavTree component, the component doesn't manage it's own selection state anymore, instead, the current selected elements are all managed in the CadView component, so that's why the test is limited to the click/ shift click event, let me know what you think :)

Adrian62D:
@aozien Will this also close https://github.com/bldrs-ai/Share/issues/541 ?

aozien:
@Adrian62D  this PR doesn't affect any widget API functionality, however, you can already use the widget API to:
1- select multiple elements
2- remove a selected element and set a new selection (deselection)
3- clear all (set selection to empty array)
the only functionality that's missing from the app and the widget is Select All

pablo-mayrgundter (approved):

Issue #475, opened 2022-11-07
VYZN: As a user I can filter one or more elements via search


Adrian62D:
As a user I can filter one or more elements using search including the ability to toggle visibility or specify highlight.

**Concept**

![image](https://user-images.githubusercontent.com/104829208/210976092-e07669de-71ca-4a33-9cf4-fa7005415aa7.png)

- The Bldrs "search box" is changed from a declarative approach to an **imperative** approach.
  - One or more  commands are being executed in order
  - In each command a set of elements can be selected and acted upon
- Comparison
  - See [SmartViews](https://www.bimcollab.com/en/products/bimcollab-zoom/smart-views) as a reference for an imperative approach.
  - See [GitHub Filtering](https://docs.github.com/en/issues/tracking-your-work-with-issues/filtering-and-searching-issues-and-pull-requests) as a reference for a declarative approach.
  - The imperative approach allows for more complex scenarios and better extensibility.
  - The imperative approach serves as a foundation for UI's to be built on top of it (Similar to SmartViews). 
  - The imperative approach is less intuitive, less easy-to-use than a declarative approach. 

**Examples**

Color all walls red and all slabs blue
`type:Wall set color:red; type:Slab set color:#0000FF;`

Hide all walls where property PSet_x.y equals "hello world" and highlight walls where PSet_z.v greater 5
`type:Wall PSet_x.y:"hello world" set visibility:hidden; type:Wall PSet_z.v:>5 set visibility:highlighted;`

Colorize and label a wall with a given global id
`globalid:123456 set label:"this is the wall having the highest energy loss" color:#FF0000 visibility:highlighted`

Select elements by express ids
`expressid:88 expressid:89 expressid:90 selection add`

Color-scale walls linearly according to energy loss (red-ish) 
`type:Wall PSet_Energy.energyloss:defined set color:#{ (e.PSet_Energy.energyloss * 255).ToString(16) }0000 `

Export a set of elements with a handful of properties to CSV with column headings ID, X and Z
`type:Slab export globalid,PSet_x.y,PSet_v.z "ID","X",Z"`

**Selection**

The following attributes can be used for selection:
- type
- expressid
- globalid
- *.* (PSet attributes)

The following operators are supported for selections:
[...]

PR #605, opened 2023-02-15
Simple heatmap implementation example

### LINKED ISSUE #603: Simple heatmap implementation
When the parameter `&view=CH.SIA380-1.heatmap` is supplied in the URL, the following change is applied to the rendering of the model.

- Across all elements of the model find the MIN and the MAX values of the attributes PSet_vyzn.Verlust or PSet_Vyzn.SIA380-1.TransmissionHeatLoss 
- Color **each** element according to these rules
  - 0 or undefined -> #F0F0F0
  - Positive values ->  Scale colors linearly from #F0F0F0 (min) to #EB3324 (max)
  - Negative values -> Scale colors linearly from #F0F0F0 (min) to #22B14C (max)

See attached model for an example.
[220714_seestrasse Heatmap (2).zip](https://github.com/bldrs-ai/Share/files/10744693/220714_seestrasse.Heatmap.2.zip)

aozien:
**This PR closes #603** 

**TL;DR** add `&view=CH.SIA380-1.heatmap`  to the url, it will disappear but the setting will be applied nontheless

~~It shouldn't be merged with main, it's just a quick demo of~~ 
visualizing a property value as a heat map color scale from #110000 to #FF0000

Adding query parameters will be cleared upon model reload, so as a work around, the main container cadview 'remembers' whether or not the parameter `&view=CH.SIA380-1.heatmap`  existed in the url, and will use the view rules in all subsequent model loads.


pablo-mayrgundter:
I'm unclear on the status.. could you update the PR description with status and provide a demo link?

aozien:
Sorry, I missed this reply, This PR is ready to be merged.

This special coloring mode can be accessed by adding the view parameter to the url, for example

https://deploy-preview-605--bldrs-share.netlify.app/share/v/p/index.ifc?view=CH.SIA380-1.heatmap

src/Infrastructure/ColorHelperFunctions.js
```diff
@@ -0,0 +1,52 @@
+import IfcColor from './IfcColor'
+
+/**
+ * Convert Hex Color string to IfcColor Object
+ *
+ * @param {string} hexColor the color in string format '#C7C7C7'
+ * @return {IfcColor} the IfcColor object
+ */
+export function parseColor(hexColor) {
```
pablo-mayrgundter, line 9: Please add unit tests

src/Infrastructure/ColorHelperFunctions.js
```diff
@@ -0,0 +1,52 @@
...
+ * get a color that lies on the scale between two colors
+ *
+ * @param {IfcColor} startColor
+ * @param {IfcColor} targetColor
+ * @param {number} value
+ * @param {number} min value
+ * @param {number} max value
+ * @return {IfcColor} color on the scale between the start color and max color
+ */
+export function interpolateColors(startColor, targetColor, value, min, max) {
```
pablo-mayrgundter, line 29: Please add unit tests

src/Infrastructure/ColorHelperFunctions.js
```diff
@@ -0,0 +1,52 @@
...
+ * convert value from one value scale to another
+ *
+ * @param {number} value
+ * @param {number} min
+ * @param {number} max
+ * @param {number} targetMin
+ * @param {number} targetMax
+ * @return {number} the new value
+ */
+export function changeValueScale(value, min, max, targetMin, targetMax) {
```
pablo-mayrgundter, line 48: Please add unit tests

src/Infrastructure/ViewRulesCompiler.js
```diff
@@ -0,0 +1,60 @@
...
+/* eslint-disable no-magic-numbers */
+/**
+ * Convert view rules into view settings object
+ *
+ * @param {object} api
+ * @param {number} modelID
+ * @param {number} rules
+ * @return {IfcCustomViewSettings} the view settings
+ */
+export async function compileViewRules(api, modelID, rules) {
```
pablo-mayrgundter, line 17: Please add unit tests

src/Containers/CadView.jsx
```diff
@@ -33,6 +33,7 @@ import {addSceneLayer} from './SceneLayer'
 export const searchIndex = new SearchIndex()
 let count = 0
 
+const useViewRules = window.location.href.match(/(view=CH.SIA380-1.heatmap)/ig)?.length > 0
```
pablo-mayrgundter, line 36: Can you move this to IfcViewerApiExtended?  We're trying to limit any changes to CadView, so should avoid any changes and looks like this doesn't actually have a dep.

src/Infrastructure/ViewRulesCompiler.js
```diff
@@ -0,0 +1,60 @@
...
+  const allPropObjects = []
+  for (let i = 0; i < propRelLines.size(); i++) {
+    allPropObjects.push( await api.GetLine(modelID, propRelLines.get(i), true))
+  }
+  const psetObjects = allPropObjects.filter((x) => x.RelatingPropertyDefinition.type === IFCPROPERTYSET)
+  // Get only sets containing PSet_vyzn.Verlust
+  const objectsAndPropVal = psetObjects.map((a) =>
+    ({o: a.RelatedObjects[0]?.expressID,
+      p: a.RelatingPropertyDefinition.HasProperties?.find(
+          (s) => s.Name.value === 'Verlust' ||
```
pablo-mayrgundter, line 29: I wonder if these ought to be passed by URL param and have the application be stateless..

- on load, look for StateToken encoding param names to be colorized, and a color map
- recolor mesh in scene (unload and reload mesh if need be?)
- no state kept

pablo-mayrgundter (commented):
Hey Zien, I think this is great functionality, but still a little ambivalent about the kind of demo status of this.  I have a proposal inline about passing more params to generalize this.  See what you think

src/Infrastructure/ViewRulesCompiler.js
aozien: Hi @pablo-mayrgundter The current 'ViewRulesCompiler' file is a just dummy implementation that mimics the compilation of a bunch of rules against the IFC model on loading, and returns a result of 'IfcCustomViewSettings' to be used for overriding model colors.

The compiler functionality itself, and how the 'ViewRules' are stored and shared or changed will be explored in details in #475, that way it would be more generic and flexible, but for now it's fixed with the predefined view rules in #603

They're both big topics (compiling and storing/changing the ViewRules) that this PR will get too complicated if it tried to cover both of them as well

this PR's main focus was validating the workflow of having a predefined set of rules, that would be compiled and applied on model load, that's why there's not much focus on the actual compilation of viewRules itself

src/Infrastructure/ViewRulesCompiler.js
aozien: This is currently just a placeholder for the rule compilation functionality

PR #609, opened 2023-02-17
Replace storybook with react-cosmos.  Mostly working.

pablo-mayrgundter:
Still needs some stories moved over, and need to check the static build process.


pablo-mayrgundter:
Hi guys, can you see what you think about this?  Ogali, if you could help me figure out the action to trigger to switch the build.. it's now `yarn build-cosmos` and that will output to docs/sb (which we can change; maybe docs/ui).

Storybook has been pretty broken for awhile.  It's very complex to configure and tons of out-of-date deps.  react-cosmos seems to be pretty good so far, though it does lack theme support.  We weren't using that with SB tho.

If it's reasonable I'll continue moving over the other stories, get the static build working and make the auto-config params work better.

pablo-mayrgundter:
This is building and running for me locally, but deploy preview fails with an MSW error.  Not sure why it is being included tho, as it's not a dep of react-cosmos.. so maybe `/sb/` is falling back to root routing and cosmos is still in development mode

pablo-mayrgundter:
Ok, ready for review!

pablo-mayrgundter:
ping

OlegMoshkovich:
looking over it

pablo-mayrgundter:
> looking over it. many files:)

Note they're 80% or so *deletes*.  Storybook was a beast

oo-bldrs (approved):

Issue #599, opened 2023-02-13
Notes: Placemarks original issue made for Ron


pablo-mayrgundter:
We'd like to have in-scene placemarks, e.g. an icon at a certain location x,y,z.  And also to have these created and referenced from Notes.

## PR 1: In-scene placemarks
1) Add a State Token to the bldrs permalink with the "m:" prefix, e.g. "m:1,2,3".  When it's present and notes panel is opened, show an active placemark mesh that works like sprite, like the red Google Map icon, at that location in the scene.  This should be an svg that its color can be converted between any colors.

Proposed placemark icon: https://thenounproject.com/icon/map-pin-415550/
Proposed Feature State Token for permalink: https://github.com/bldrs-ai/Share/wiki/Design:-URL-Structure:-State-Tokens#placemark-token
PR: https://github.com/bldrs-ai/Share/pull/709
Demo Link: https://deploy-preview-709--bldrs-share.netlify.app/share/v/gh/ronplaymeta/bldrs-ai-Share/main/public/index.ifc?feature=placemark

## PR 2: Store placemark URLs in Notes
1) This should work like Cameras in Notes, where each note may have some metadata encoding placemarks for the scene.  When Notes are loaded from GitHub, scan all notes for Markdown links and extract placemarks to display.  e.g. there may be 10 notes for a model, and each might have a few mark URLs in its markdown.  So there would be a total of 10x3=30 placemarks to display in the scene.
2) Each embedded URL can have only 1 placemark, so that when they're clicked on to make active, only 1 active placemark is visible.  A note can have multiple URLs with placemarks.
3) Display all extracted placemarks in the scene using "inactive" icons, like a small dot.  To start with, these should be a single color like grey.  If it's easy, we should use color of any tag in the issue to set the color of the inactive dot icon.
4) When an inactive placemark is clicked, it becomes active.  This changes the URL to include its State Token (m:x,y,z for the current placemark) and opens the associated Note if there is an issue State Token present (i:1234) in the placemark's link.

PR: https://github.com/bldrs-ai/Share/pull/709
Demo Link: https://deploy-preview-709--bldrs-share.netlify.app/share/v/gh/ronplaymeta/bldrs-ai-Share/main/public/index.ifc?feature=placemark

## PR 3: Create placemarks from notes
1) User clicks placemark icon on the Notes on the sidebar
2) user double click or shift + single click any object to place icon at that location on the object; the precision matters here, so ideally it's placed exactly where they put it
[...]

PR #611, opened 2023-02-17
In-scene placemarks

ronplaymeta:
Hi @pablo-mayrgundter, I updated codes for firs PR of https://github.com/bldrs-ai/Share/issues/599.
Please check the result and give me feedback.
Thank you.


OlegMoshkovich:
@ronplaymeta the preview was failed - some tests are failing please fix.

pablo-mayrgundter (changes requested):
- Per convo, please remove any changes to Notes from this PR and submit those in a subsequent change (and probably after Oleg's work goes in).

- Please move what you can of the changes to CadView into a neighboring file.  My recent PR to introduce the SceneLayer is a good example.

- For the icon, please edit it to remove the attribution.  There should be only a couple paths in there for the main shape.  We should note the attribute though, so please add a README.md in the assets directory listing where it came from, with URL.

pablo-mayrgundter (changes requested):
About the placemark itself, it should always face towards the user. Can you try the sprite material as discussed in the dicord thread.

It also has a white background.  See if you can make this transparent.  The SVG file should just have paths for the shape, with no background.

ronplaymeta:
ptal

src/utils/svg.js
```diff
@@ -0,0 +1,199 @@
...
+      const material = new THREE.MeshBasicMaterial({map: texture, side: THREE.DoubleSide})
+      material.map.minFilter = THREE.LinearFilter
+      const mesh = new THREE.Mesh(geometry, material)
+      resolve(mesh)
+    }
+  })
+}
+
+
+export const getSVGSprite = async ({
```
pablo-mayrgundter, line 156: Please use function with jsdoc for top-level funcs

src/utils/svg.js
```diff
@@ -0,0 +1,199 @@
...
+  group.scale.y = scaleY
+  group.position.x = -width
+  group.position.y = height
+  group.scale.y *= - 1
+  svgGroup.add(group)
+  return svgGroup
+}
+
+
+export const getSVGMesh = async ({
```
pablo-mayrgundter, line 119: Please use function with jsdoc for top-level funcs

src/utils/svg.js
```diff
@@ -0,0 +1,199 @@
...
+import debug from './debug'
+
+
+const svgLoadingManager = new THREE.LoadingManager()
+const svgLoader = new SVGLoader(svgLoadingManager)
+const fileLoadingManager = new THREE.LoadingManager()
+const fileLoader = new THREE.FileLoader(fileLoadingManager)
+
+
+export const getSVGGroup = async ({
```
pablo-mayrgundter, line 13: Please use function with jsdoc for top-level funcs

src/utils/svg.js
```diff
@@ -0,0 +1,199 @@
+import * as THREE from 'three'
```
pablo-mayrgundter, line 1: Please add unit tests for this class

pablo-mayrgundter (changes requested):
Getting close.  Per convo, still need to remove the Notes work (for later) and please test the next svg code.

There's also some commented code.. if that's for debugging please remove.. tho realize it might be prep for follow-up work, in which case it's ok to leave.

src/utils/svg.js
ronplaymeta: done

src/utils/svg.js
ronplaymeta: done

src/utils/svg.js
ronplaymeta: done

src/utils/svg.js
ronplaymeta: done

ronplaymeta:
ptal

ronplaymeta:
ptal

ronplaymeta:
ptal

ronplaymeta:
ptal

pablo-mayrgundter:
Heya, looks like you're still working...or is this ready for review?

ronplaymeta:
It's ready to review, please check. :)
ptal

PR #626, opened 2023-02-22
Flesh out Auth0 user mock

oo-bldrs:
When the initial Auth0 mock user object was created, I specified the bare minimum field set in order to test my UI component.

As additional work progresses this object should be fleshed out to reflect all fields that are returned by Auth0 to support future development.

This PR brings the object to that expected level.


OlegMoshkovich (approved):
LGTM

PR #628, opened 2023-02-22
Explicit build control via environment variables

oo-bldrs:
Originally, there was logic that would enable or disable different parts of the build depending on the value of the `NODE_ENV` environment variable.

However, using this variable is fraught with problems due to the nature of the tooling combination that interprets its values in different ways[1][2].

This PR introduces dedicated environment variables for controlling specific behavior when the build is being performed.

This goes along with the necessary configuration changes within Netlify to effect the build as desired.

1: https://seanconnolly.dev/dont-be-fooled-by-node-env
2: https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-environment

### DIFF
--- a/config/build.js
+++ b/config/build.js
@@ -8,7 +8,7 @@ esbuild
     .build(common.build)
     .then((result) => {
       // Remove development resources from non-development builds
-      if (process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== 'test') {
+      if (process.env.DISABLE_MOCK_SERVICE_WORKER === 'true') {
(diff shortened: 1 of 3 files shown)


oo-bldrs:
Sentry issue: [SHARE-A](https://sentry.io/organizations/bldrs-ai/issues/3887210099/?referrer=github_integration)

pablo-mayrgundter (approved):
Yay, very nice.  I've switched minification off and forgot about it too many times. This will help remember the distinction

PR #633, opened 2023-02-23
Introduce snapshot testing for three.  Removes global three mock to allow more custom handling by tests.  Requires some utilities to remove uuids from snapshots.

pablo-mayrgundter:


pablo-mayrgundter:
Whoops, meant for Ron's fork

PR #636, opened 2023-02-26
App Store Drawer

### LINKED ISSUE #464: VYZN: As a user I can launch the vyzn app from Bldrs
As a user I can launch the vyzn app from Bldrs

**Workflow**

1. Open Bldrs
2. Start vyzn app
3. Vyzn app loads

SSO or loading the proper ifc model is not in the scope here.

**UI** 

When the URL `&appstore=true` is supplied to Bldrs perform the following changes in the UI:

1. Add an app-store icon here in the right panel
2. When clicking the icon the right panel toggles on/off

image

The panel should contain the following content:

image

https://excalidraw.com/#json=8K9AZSF17N8U6J9hht213,LvzAt9zCXlP4OfI8dgPeVg

Scope:
- Initially use a custom HTML page inside the iframe instead of prod.vyzn.tech
- In Bldrs store the current uploaded file in Local Storage of the browser so that the raw file remains available for vyzn later on. Currently the file is discarded by Bldrs.
- When launching the vyzn applet, transmit the **raw file** through the iframe boundary to the vyzn applet
  - This should be triggered by a Matrix Widget API message
  - If there is a github URL available then transmit the URL as well (the data is discarded at the moment)
- The applet show allow the user to
  - chose a project from a dropdown ("Projekt")
  - chose a building from a dropdown ("Gebäude")
  - hit the button "import"
    - this should close the applet and open vyzn in fullscreen (navigated to project > building) 

The iframe concept has already been explored in a prototype: 

- https://github.com/bldrs-ai/Share/pull/379
- https://deploy-preview-379--bldrs-share.netlify.app/share/v/p/index.ifc#c:-111.37,14.94,90.63,-43.48,15.73,-4.34

This issue should be split later on.

aozien:
### This PR should close #464  

~To get this to work you've to first serve this file from visual studio, so that it's served on port 5500, you can also change the AppStoreData.json file to change the action to be any url~ **(update: now it's served from local path)**

### How storing the file data works:
1- a new property loadedFileInfo is added to the store, and gets updated when the model changes.
2- if the model was loaded from a local file we don't have to store the whole file, storing the file info is enough, and the hosted frame will be able to re-read the file from disk.
3- if the model was loaded from bldrs/share or github it will store the 

### How sharing the data with the app works:
Currently this implementation uses the native [MessageChannel](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel) to avoid any overheads or dependencies, so that apps can integrate without having to depend on certain libraries, but that's a topic that can be open for discussion as well

The hosted frame sends an action name, and the host will respond with the same action name with the response data

image

And for local files
image


pablo-mayrgundter:
wowow!

image

pablo-mayrgundter:
Could you please move the AppStore button into the third group, after the separator and before the Theme button.

There's a typo in the Vyzn App description "environemental" -> "environmental".  Could you please also add to the description smth like "- DRAFT"

Also, see the screenshot.. the icon is filled in black.  Is that as intended?  I think something more like this?  https://thenounproject.com/icon/apps-5361891/

@OlegMoshkovich thoughts?

pablo-mayrgundter:
@oo-bldrs could you check this out from a security perspective?  The message submission from form input makes me a little nervous but it's been ages since i've thought about code injection

pablo-mayrgundter:
Might be nice to have the Vyzn App icon from the Store prefix the title when the App is active.  Wdyt?

image

aozien:
@pablo-mayrgundter I believe it makes sense to add the app icon as well, I've updated the branch
Also regarding the Icon, it was just a placeholder, also I'm not sure if I can use the icon you provided now because of its license, so I used the closest one that I could find, let me know what you think.

pablo-mayrgundter:
This looks good.  Not sure we want it live yet though.  Talked with @Adrian62D today and one idea is to add a feature flag to the URL, e.g. feature=apps, and set it in the zustand store on init.  So launch process would be:
   - That way you can check in code now, disabled in zustand by default
   - enabled if feature flag present in URL for testing
   - Later enabled by default in zustand, optionally removed

aozien:
Hi @pablo-mayrgundter , I've updated the branch from main, and added the feature=apps check before displaying the open app store icon, PTAL :)

src/BaseRoutes.jsx
```diff
@@ -69,6 +70,16 @@ export default function BaseRoutes({testElt = null}) {
...
 
+  /**
+   * Store initial query parameters, since they will be cleared by the application on model loading.
+   */
+  useEffect(() => {
+    const initialParameters = new URLSearchParams(window.location.search)
+    const enabledFeature = initialParameters.get('feature')
+    const appStoreEnabled = enabledFeature && enabledFeature.toLowerCase() === 'apps'
+    setIsAppStoreEnabled(appStoreEnabled)
+  }, [])
```
pablo-mayrgundter, line 81: please add dep for setIsAppStoreEnabled

src/Components/PanelTitle.jsx
```diff
@@ -10,7 +10,7 @@ import Typography from '@mui/material/Typography'
  * @param {object} [controlsGroup] Controls Group is placed on the right of the title
```
pablo-mayrgundter, line 10: Please update these jsdocs

src/BaseRoutes.jsx
```diff
@@ -69,6 +70,16 @@ export default function BaseRoutes({testElt = null}) {
     }
   }, [basePath, installPrefix, location, navigation, getAccessTokenSilently, isAuthenticated, isLoading, setAccessToken])
 
+  /**
```
pablo-mayrgundter, line 73: I don't think this needs to be here, and in general let's try to keep the core as light as possible.

We've been using a pattern of having URL read/write logic and the associate Component in a single file, naming the file ThingControl.jsx.  Can you consolidate this code like that?  Also please add a cypress test for enabling via the URL parameter.

It seems <AppStoreDrawer> gets loaded via cadview, no matter what..

pablo-mayrgundter (changes requested):
Ok, I think this is the last set of changes.  Just some code organization.

src/BaseRoutes.jsx
aozien: @pablo-mayrgundter  I see your point as why it shouldn't be in the BaseRoute file, however the reason the urlParameter logic wasn't in the same file as the component is that it will not work for **https://bldrs.ai/?feature=apps** for example,  because the initial model load (index.ifc) will clear the query parameters when calling  React's "navigate" function, testing it in cypress will be as simple as ` cy.visit('/?feature=apps')`

It'll still work on some cases such as **https://bldrs.ai/share/v/p/index.ifc?feature=apps** but I thought that can be confusing.

IMO the permanent solution to this dilemma, is to preserve the query parameters when using React's "navigate" function throughout the application, what do you think?

pablo-mayrgundter: Aha.. I didn't think of that, but I think it's ok to use the expanded path and set it from the component.

Here's why.  The basic structure of the path is to allow us to have multiple top-level pages in the future, e.g. profile page or an editor, etc.. These can be built of the same shared components, e.g. the AppStore.  So I think that means components should make minimal assumptions about what path they may run under and so not necessary to think of the feature flag as being global state.

The cypress test can be a little brittle because of this, but I think that's ok.

pablo-mayrgundter: Keeping track of discussion here.. you've said current cypress can't modify full paths.  Next step is to check

https://www.npmjs.com/package/cypress-react-router

As this will be an ongoing challenge for us.

pablo-mayrgundter:
Let's have some more planning here.. https://discord.com/channels/853953158560743424/1090001703681658890/1098247117685932193

If we can't resolve it, let's get this merged.

pablo-mayrgundter:
I had lost track of this.. @aozien you had a blocking question there.  Let's try in the future to make sure any blocking issues are mentioned here.

This is me doing that now

TODO(pablo): answer Zien here:
https://discord.com/channels/853953158560743424/1090001703681658890/1098247117685932193

pablo-mayrgundter (approved):
Awesome!

PR #640, opened 2023-03-03
File access dialog

OlegMoshkovich:
This PR extends open dialog to include a select components for organization, repository and files.

### DIFF
--- a/src/Components/OpenModelControl.jsx
+++ b/src/Components/OpenModelControl.jsx
@@ -1,4 +1,4 @@
-import React, {useState} from 'react'
+import React, {useState, useEffect} from 'react'
 import {useNavigate} from 'react-router-dom'
 import Box from '@mui/material/Box'
 import MenuItem from '@mui/material/MenuItem'
@@ -6,11 +6,16 @@ import Paper from '@mui/material/Paper'
 import Typography from '@mui/material/Typography'
 import TextField from '@mui/material/TextField'
 import useTheme from '@mui/styles/useTheme'
+import {useAuth0} from '@auth0/auth0-react'
 import Dialog from './Dialog'
 import {TooltipIconButton} from './Buttons'
+import Selector from './Selector'
+import useStore from '../store/useStore'
+import {handleBeforeUnload} from '../utils/event'
+import {getOrganizations, getRepositories, getFiles, getUserRepositories} from '../utils/GitHub'
+import {RectangularButton} from '../Components/Buttons'
 import OpenIcon from '../assets/icons/Open.svg'
 import UploadIcon from '../assets/icons/Upload.svg'
-import {handleBeforeUnload} from '../utils/event'
 
 
 /**
@@ -20,7 +25,25 @@ import {handleBeforeUnload} from '../utils/event'
  */
 export default function OpenModelControl({fileOpen}) {
   const [isDialogDisplayed, setIsDialogDisplayed] = useState(false)
+  const [orgNamesArr, setOrgNamesArray] = useState([''])
+  const {user} = useAuth0()
   const theme = useTheme()
+  const accessToken = useStore((state) => state.accessToken)
+  useEffect(() => {
+    /**
+     * Asynchronously fetch organizations
+     *
+     * @return {Array} organizations
+     */
+    async function fetchOrganizations() {
+      const orgs = await getOrganizations(accessToken)
+      const orgNamesFetched = Object.keys(orgs).map((key) => orgs[key].login)
+      const orgNames = [...orgNamesFetched, user && user.nickname]
+      setOrgNamesArray(orgNames)
+      return orgs
+    }
+    fetchOrganizations()
+  }, [accessToken, user])
 
 
   return (
@@ -47,6 +70,7 @@ export default function OpenModelControl({fileOpen}) {
           isDialogDisplayed={isDialogDisplayed}
           setIsDialogDisplayed={setIsDialogDisplayed}
           fileOpen={fileOpen}
+          orgNamesArr={orgNamesArr}
         />
       }
     </Box>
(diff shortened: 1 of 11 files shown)


pablo-mayrgundter:
Looking good!

May have found a bug.. opening ifctool/index.ifc shows an error in console and leaves the dialog open.. 

image

OlegMoshkovich:
Hmm - I wonder what that is about

src/__mocks__/api-handlers.js
```diff
@@ -137,4 +135,31 @@ export const handlers = [
         ctx.status(httpOk),
     )
   }),
+
```
oo-bldrs, line 138: Mock responses should resemble the real API shape and response as much as possible.

src/utils/GitHub.test.js
```diff
@@ -69,4 +72,18 @@ describe('GitHub', () => {
       expect(res.status).toEqual(httpOK)
     })
   })
+  describe('get models from github', () => {
```
oo-bldrs, line 75: Once the mocks are updated to contain real shape and responses, these test assertions should also be updated.

OlegMoshkovich (commented):
@oo-bldrs PTAL

oo-bldrs:
@OlegMoshkovich Why are you returning `@orgs` instead of `orgs`?

src/Components/OpenModelControl.jsx
```diff
@@ -20,9 +25,27 @@ import {handleBeforeUnload} from '../utils/event'
...
+  const accessToken = useStore((state) => state.accessToken)
+  useEffect(() => {
+    /**
+     * Asynchronously fetch organizations
+     *
+     * @return {Array} organizations
+     */
+    async function fetchOrganizations() {
+      const orgs = await getOrganizations(accessToken)
+      const orgNamesFetched = Object.keys(orgs).map((key) => orgs[key].login)
```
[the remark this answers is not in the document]
OlegMoshkovich: @oo-bldrs perhaps I misunderstood... I tried to replace org with @org per GitHub convention and VS code was complaining about it.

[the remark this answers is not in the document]
pablo-mayrgundter: Per discussion, I think Ogali is suggesting the '@' is part of the display string for users and orgs, as that's common in git, as you can see when you @OlegMoshkovich someone.

I'd say that's more of a display format tho, since we often pass org and usernames without it.  So I'd keep the bare org and user names here, and then when you go to display them below, do smth like `` `@${orgName}` ``

src/Components/OpenModelControl.jsx
```diff
@@ -59,11 +83,50 @@ export default function OpenModelControl({fileOpen}) {
...
+  const accessToken = useStore((state) => state.accessToken)
+
+
   const openFile = () => {
     fileOpen()
     setIsDialogDisplayed(false)
   }
+
+  const selectOrg = async (org) => {
+    setSelectedOrg(org)
```
pablo-mayrgundter, line 105: Would clarify this code to change the react state var from `selectedOrg` to `selectedOrgName` and then this bit would be:

```
const selectOrg = async (orgNdx) => {
  const orgName = orgNamesArr[orgNdx]
  setSelectedOrgName(orgName)
  ...L110:
    repos = await getRepositories(orgName, accessToken)
```
and below, assuming you change the other variables similarly:
```
`/share/v/gh/${orgName}/${repoName}/main/${fileName}`
```

src/Components/Selector.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
+import MenuItem from '@mui/material/MenuItem'
+import Typography from '@mui/material/Typography'
+import TextField from '@mui/material/TextField'
+import useTheme from '@mui/styles/useTheme'
+import {handleBeforeUnload} from '../utils/event'
+
+
+/**
+ * @property {Function} setIsDialogDisplayed callback
```
pablo-mayrgundter, line 10: Please fill out the jsdoc properties

pablo-mayrgundter (changes requested):
Few style things from me and then Ogali's unit test reqs and looks good to go

src/Components/OpenModelControl.jsx
OlegMoshkovich: I tried it, not so straight forward.
If you know how to do it, please take this branch and make the change.
Thank you!

OlegMoshkovich: Sorry about that one, I misunderstood what you were suggesting.

src/Components/OpenModelControl.jsx
```diff
@@ -20,9 +25,27 @@ import {handleBeforeUnload} from '../utils/event'
...
+      const orgs = await getOrganizations(accessToken)
+      const orgNamesFetched = Object.keys(orgs).map((key) => orgs[key].login)
+      const orgNames = [...orgNamesFetched, user ? user.nickname : 'my own repo']
+      setOrgNamesArray(orgNames)
+      console.log('in the fetch Organizations', orgs)
+      return orgs
+    }
+    fetchOrganizations()
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
```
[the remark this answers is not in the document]
OlegMoshkovich: image

Not sure why it is not showing the change.. that is what Pablo was talking about.

src/Components/OpenModelControl.jsx
```diff
@@ -20,9 +25,27 @@ import {handleBeforeUnload} from '../utils/event'
  */
 export default function OpenModelControl({fileOpen}) {
   const [isDialogDisplayed, setIsDialogDisplayed] = useState(false)
+  const [orgNamesArr, setOrgNamesArray] = useState(['loading'])
```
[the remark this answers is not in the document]
OlegMoshkovich: same issue here - it is actually fixed.
image

OlegMoshkovich (commented):
@pablo-mayrgundter you requested changes which are blocking the pr, but all of the comments have been resolved. can you please take a look.

pablo-mayrgundter (approved):

PR #655, opened 2023-03-10
Create placemarks

ronplaymeta:
Hi @pablo-mayrgundter 
It's completed to create place marks from notes, please review and give me feedback.
Thank you.


pablo-mayrgundter:
Hey Ron!

Taking a look now.  Not sure it's working correctly.. 

1) Create a note
2) Click on the placemark icon
3) not sure what to do here.. i double-click on a building element, trying to place the mark..?  But nothing showing up.

Is this the right way?

ronplaymeta:
Here is the steps.
1. Create a note
2. Click on the place mark icon
3. Shift click on the building element, then can see the place mark is placed on that element
4. Select that note, then can see new place mark URL is added.
Thank you for checking current result. :)

pablo-mayrgundter:
Ok, working for me now.

Main comment, this is #599 PR #3, and as discussed we'll need PR #2 to go in first, as this PR presents a partially-functional feature to the user.

Still, good progress and this looks like it will be good after that.

A few issues:

- Placing: The precision of the placing is off.  But maybe that's bc the size of the icon is very large.  See below.
- Placing: the shift-click approach is functional but not discoverable.  Instead, please use the approach we talked about in Discord:
    https://discord.com/channels/853953158560743424/1074784439789420676/1074789861501976688
  The user clicks the placemark button, the cursor change style, and then a double-click places the mark and returns the cursor to normal.
- Size: Currently too big.  I think about 0.5 meters?
- Style: only the current active placemark should be the red active style.  The rest should be dots, say 0.25m.  See #599, PR #2, item 3

ronplaymeta:
I fixed all issues, and will checkout new branch from current branch. :)
PTAL

OlegMoshkovich:
Ron I think there is an error with one of the functions.


image

Issue #666, opened 2023-03-16
FS: Add breadcrumb model path to GItHub


OlegMoshkovich:
<img width="652" alt="image" src="https://user-images.githubusercontent.com/3433606/225645381-e3af8092-c25e-4aaa-98ed-97bad5358940.png">

FS: Add breadcrumb path to GItHub for currently opened model with the links to organization, repository and file

PR #675, opened 2023-03-21
Add a breadcrumbs model path

OlegMoshkovich:
#666


oo-bldrs:
@OlegMoshkovich This PR is titled about breadcrumbs but the contents seems to also include organization-related API changes. Was that intended?

OlegMoshkovich:
@oo-bldrs i think it is because I merged the file system pr into it.

PR #685, opened 2023-03-25
Progressive reveal on NavTree element hiding

pablo-mayrgundter:
using :focus and :hover selectors; lower opacity.


OlegMoshkovich (approved):

Ibrahim5aad:
@pablo-mayrgundter This is making the cypress test to fail. I made a PR in your fork with a small fix that will solve it.

pablo-mayrgundter:
> @pablo-mayrgundter This is making the cypress test to fail. I made a PR in your fork with a small fix that will solve it.

Thanks! Merged

PR #688, opened 2023-03-27
Disallow use of console, alert, and debugger, part 2.  

oo-bldrs:
PR to add ESLint rules that will loudly complain if one attempts to use console.*, alert(), or debugger;.

Repeat PR due to a bad merge.


pablo-mayrgundter (approved):

oo-bldrs:
@pablo-mayrgundter The unit tests are failing from the `console.log` statement you introduced into a Cosmos fixture. How do you want to resolve this?

pablo-mayrgundter:
> @pablo-mayrgundter The unit tests are failing from the `console.log` statement you introduced into a Cosmos fixture. How do you want to resolve this?

I think eslint errors by default are fine but also the console output is permanently useful for fixtures.  Are they suppressible like normal?  If so just suppress the existing ones and we can do so going forward

PR #691, opened 2023-03-27
Don't allow an anonymous user to request user organizations

oo-bldrs:
When the application is loaded, it unconditionally requests a list of organizations from GitHub that the logged in user is a member of. This causes an exception to be thrown every time a non-logged-in user loads the page.

This PR adds logic to reject a request for a user's organizations if we're not logged in, fixes the shape of the mock object returned during testing, and separates the tests for `getOrganizations` into their own suite.


pablo-mayrgundter (approved):
Nice!

PR #693, opened 2023-03-27
Introduce navWith utility for carrying search and hash state.

pablo-mayrgundter:
This is prep work to allow feature flags to be kept in URLs and not require cookies for testing.  Also move canvas npm over to devDeps.. still having trouble with it not being loaded.

For demo, try adding ?foo=bar,baz to preview instance

https://deploy-preview-693--bldrs-share.netlify.app/share/v/p/index.ifc?foo=bah%2Cblaz#c:-150.147,-85.796,167.057,-32.603,17.373,-1.347


src/utils/debug.js
```diff
@@ -5,7 +5,7 @@ const WARN = 2
 const INFO = 1
 const DEBUG = 0
 /* eslint-enable no-unused-vars */
-let DEBUG_LEVEL = OFF
+let DEBUG_LEVEL = DEBUG
```
ronplaymeta, line 8: revert this please.

pablo-mayrgundter: reverted

src/utils/navigate.js
```diff
@@ -0,0 +1,89 @@
...
+ * Helper for calling navigate that will append location.hash to the
+ * given path, if present.
+ *
+ * @param {Function} navigate
+ * @param {string} path
+ */
+export function navWithHashState(navigate, path) {
+  navWith(navigate, path, {
+    hash: location.hash,
+  })
```
aozien, line 53: I'd try to avoid adding unused functions, might cause confusion and lowers maintainability,
this is regarding both `navWithSearchState` and `navWithHashState`

src/utils/navigate.js
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param {Function} navigate
+ * @param {string} path
+ * @param {object} options
+ */
+export function navWith(navigate, path, options = {
+  search: '',
+  hash: '',
+}) {
+  const search = options.search || ''
```
aozien, line 17: I'd prefer the default behaviour to be maintaining current parameters,
```
export function navWith(navigate, path, options = {
  search: location.search,
  hash: location.hash,
}) {
```
 that way you wouldn't need the function `navWithSearchAndHashState` as well, if someone wants to preserve the state (query and hash) he can call navWith with 2 parameters, and if he wants to override any of which he should just pass the options parameters similar to what you did here 

``` 
 navWith(navigate, defaultPath, {
    search: location.search,
    hash: cameraHash,
  })
  ```

What do you think?

**Review by aozien — COMMENTED** [review]
Great job! I believe that would improve the navigation management through out the app.

**Comment by oo-bldrs** [comment]
Forgive me, I'm a little confused.

1. At any point, will this factor into the existing feature flag logic? (see: `src/FeatureFlags.js`)
2. Are multiple camel case functions that wrap a parent function preferred over having an additional object argument with sensible defaults?
3. Would this be a better implemented as a React hook that exposes functions already associated with the location and navigation router objects opposed to making the caller invoke the necessary hooks (`useLocation` and `useNavigate`) and manually pass along the returned objects?

File: src/utils/navigate.js (hunk 2 above)
Comment (pablo-mayrgundter) [reply src/utils/navigate.js]: Done.

File: src/utils/navigate.js (hunk 3 above)
Comment (pablo-mayrgundter) [reply src/utils/navigate.js:17]: Done.

**Review by pablo-mayrgundter — COMMENTED** [review]
Sorry for the delay!  Getting back from some travels.

**Comment by pablo-mayrgundter** [comment]
> Forgive me, I'm a little confused.
> 
> 1. At any point, will this factor into the existing feature flag logic? (see: `src/FeatureFlags.js`)

Thanks for the reminder that exists.  Yeah, we're going to put a few things behind URL-based feature flags and will consolidate there.  This is prep so that feature flags won't be cleared during nav.

> 2. Are multiple camel case functions that wrap a parent function preferred over having an additional object argument with sensible defaults?

I kinda like this but you're second vote against, so removed.

> 3. Would this be a better implemented as a React hook that exposes functions already associated with the location and navigation router objects opposed to making the caller invoke the necessary hooks (`useLocation` and `useNavigate`) and manually pass along the returned objects?

For me hooks are more magical than a wrapper.  I don't have a strong preference otherwise.

File: src/Containers/CadView.jsx [hunk 4]
```diff
@@ -429,7 +430,7 @@ export default function CadView({
     resetState()
     const repoFilePath = modelPath.gitpath ? modelPath.getRepoPath() : modelPath.filepath
     window.removeEventListener('beforeunload', handleBeforeUnload)
-    navigate(`${pathPrefix}${repoFilePath}`)
+    navWith(navigate, `${pathPrefix}${repoFilePath}`)
```
aozien, line 433: we might want to clear both hash and query paramters here?

src/Containers/CadView.jsx
```diff
@@ -452,7 +453,7 @@ export default function CadView({
         const pathIds = getPathIdsForElements(lastId)
         const repoFilePath = modelPath.gitpath ? modelPath.getRepoPath() : modelPath.filepath
         const path = pathIds.join('/')
-        navigate(`${pathPrefix}${repoFilePath}/${path}`)
+        navWith(navigate, `${pathPrefix}${repoFilePath}/${path}`)
```
aozien, line 456: having a camera hash parameter in the url while making a new selection resets camera position to the url parameter, so we might want to clear it somehow to fix this?

aozien (approved):

src/Containers/CadView.jsx
pablo-mayrgundter: Sure, going to try that and removing search as well.

src/Share.jsx
```diff
@@ -47,8 +49,8 @@ export default function Share({installPrefix, appPrefix, pathPrefix}) {
         return
       }
       if (modelPath === null ||
-        (modelPath.filepath && modelPath.filepath !== mp.filepath) ||
-        (modelPath.gitpath && modelPath.gitpath !== mp.gitpath)) {
+          (modelPath.filepath && modelPath.filepath !== mp.filepath) ||
```
ronplaymeta, line 52: not sure whether this tab space is correct.

src/Containers/CadView.jsx
pablo-mayrgundter: Done.

src/Share.jsx
pablo-mayrgundter: It is.  Doesn't look right here but is fine in my client.

src/Containers/CadView.jsx
pablo-mayrgundter: @ronplaymeta I got a merge conflict with how you're using this for placemarks.  I'm going with my version for now, but just keep in mind we should think about how persistent state is when changing selection if a placemark is active

pablo-mayrgundter:
I'm going to merge without approval since it seems Zien&Ron can't at this point.  Will look into that separately.

PR #697, opened 2023-03-28
add file path CadView

OlegMoshkovich:


pablo-mayrgundter:
Heya, so as I was suggesting when we talked in the mtg, what do you think about moving this into the Search component?

As it is, I'm trying to see it as a new user and I wonder if it's a little isolated to convey its meaning.  The size is too compact to really see the full path, and there's not label noting what it is.

My thinking with the merge into the Search component is that we make that more of a general file/open/search component, and as soon as it's hovered or focused, it expands substantially so we have room to show more context, e.g. like GH's searchbar expando

OlegMoshkovich:
I feel putting anything in search is confusing.
I like the idea of expanding the length of the component on hover to see the full path.

OlegMoshkovich:
at some point we will probably need to introduce a footer similar to vscode.
image

OlegMoshkovich:
image
It feel clear and functional.
Let's get it in.

PR #698, opened 2023-03-29
controlsGroup 

OlegMoshkovich:
This PR introduces controlsGroup to control and toggle the visibility of select components on the left of the app.

### DIFF
--- a/src/Components/ControlsGroup.jsx
+++ b/src/Components/ControlsGroup.jsx
@@ -7,9 +7,9 @@ import TreeIcon from '../assets/icons/Tree.svg'
 
 
 /**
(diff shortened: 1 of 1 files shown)


src/Components/BranchesControl.jsx
```diff
@@ -49,6 +52,11 @@ export default function Branches() {
           versionPathsTemp.push(versionPath)
         })
         setVersionPaths(versionPathsTemp)
+        if (branchesData.length > 1) {
```
OlegMoshkovich, line 55: this does not change the previous logic of the branch control, just adds a store value used in control group.

PR #699, opened 2023-03-29
Bring the dialogs up in z value

OlegMoshkovich:


src/Components/Dialog.jsx
oo-bldrs, line 1: Is there a MUI component that will use the browser's native `<dialog>` element to avoid the z-index wars?

https://developer.chrome.com/blog/what-is-the-top-layer/

oo-bldrs (changes requested):
There's a failing end-to-end test as a result of this change.

src/Components/Dialog.jsx
OlegMoshkovich: I think this is the only dialog component
https://mui.com/material-ui/api/dialog/
The tests are failing because of time out, rerunning the tests now.

OlegMoshkovich:
The tests are failing because of the timeout, it happens sometimes. Do you know why?

OlegMoshkovich:
image

PR #706, opened 2023-04-06
Remove unused effect from NoteCard

oo-bldrs:
PR to remove a state and its accompanying logic that are not used in the component.


pablo-mayrgundter:
@OlegMoshkovich can you take this and assign to yourself?

pablo-mayrgundter (approved):

PR #709, opened 2023-04-12
Use feature flag in placemark

ronplaymeta:
Hi @pablo-mayrgundter
Placemark uses feature flag to enable itself, please review and give me feedback.
Here is demo link. https://deploy-preview-709--bldrs-share.netlify.app/share/v/gh/ronplaymeta/bldrs-ai-Share/main/public/index.ifc?feature=placemark
Thank you.


src/hooks/usePlaceMark.js
```diff
@@ -49,30 +39,18 @@ export function usePlaceMark() {
...
-      if (featureValue) {
-        placeMarkEnabled = featureValue.toLowerCase() === 'placemark'
-      }
-    }
-
-    debug().log('usePlaceMark#useEffect: placeMarkEnabled: ', placeMarkEnabled)
+    const placeMarkEnabled = enabledFeature && enabledFeature.toLowerCase() === 'placemark'
     setIsPlaceMarkEnabled(placeMarkEnabled)
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
```
pablo-mayrgundter, line 45: I tested a couple cases.
  - Using the feature flag works, but I believe this condition means it will stay set after the first read.  The design we'd talked about should continuously check for the presence of the flag.  For clarity I think this is preferred.
  - I'm also seeing the feature flag being set when i click on a placemark link in a test comment.  Maybe this is how the link is being created?  I'd like to have the feature flag be manual and not referred to as permanent state in the comment links.

pablo-mayrgundter (changes requested):

src/hooks/usePlaceMark.js
ronplaymeta: - Using the feature flag works, but I believe this condition means it will stay set after the first read. The design we'd talked about should continuously check for the presence of the flag. For clarity I think this is preferred.

yep, feature flag checking works whenever placemark related funcs are called now.

- I'm also seeing the feature flag being set when i click on a placemark link in a test comment. Maybe this is how the link is being created? I'd like to have the feature flag be manual and not referred to as permanent state in the comment links.

when a user drops placemark in scene, then current url with hash is saved to comment as a link.
do we need to remove feature flag when saving?

ronplaymeta: cypress test for placemark seems to be impossible since user can drop placemark in a comment that he created and comment creation require GitHub sign.
so to speak, it's impossible to test GitHub sign locally. 🧐

pablo-mayrgundter (approved):

PR #712, opened 2023-04-20
Use hook for feature flag.

ronplaymeta:
Hi @pablo-mayrgundter 
This PR uses hook for feature flag checking logic, please review and give me feedback.
Thank you.


src/hooks/useExistInFeature.js
```diff
@@ -0,0 +1,40 @@
+import {useEffect, useState} from 'react'
+import {useSearchParams} from 'react-router-dom'
+import debug from '../utils/debug'
+
+
+/**
+ * Feature Hook
```
pablo-mayrgundter, line 7: `This hook checks for the existence of a named feature in the URL SearchParams (via react-router), e.g. `feature=app,placemark`.`

pablo-mayrgundter (changes requested):
Just one comment suggestion.

Also, please add unit tests for both hooks.

src/hooks/useExistInFeature.js
ronplaymeta: done

pablo-mayrgundter:
Per convo, feature flag no longer works for me on the demo.

Please add an e2e test for this.. i'm thinking either a screenshot capture/regression test, or mock three.js methods for adding the mesh or smth earlier than that, just to check that when feature=placemark is enabled and a comment loaded, the placemark call is made

ronplaymeta:
we should login to GitHub to test placemark CRUD, so imho, there is no way to test placemark feature.
anyway, placemark not working issue is fixed, please check again.
ptal

PR #743, opened 2023-06-03
FIX: enhancing IFC isolator perf. by  applying BVH trees when creating the hiding/isolation subsets

Ibrahim5aad:
This PR is a performance enhancement on the IfcIsolator. Applying BVH trees for the created geometry subsets seems to help with navigating the model (after applying isolation or hiding) seamlessly with the previously encountered lags.

### DIFF
--- a/src/Infrastructure/IfcIsolator.js
+++ b/src/Infrastructure/IfcIsolator.js
@@ -124,6 +124,7 @@ export default class IfcIsolator {
       modelID: 0,
       scene: this.context.getScene(),
       ids: includedIds,
+      applyBVH: true,
       removePrevious: true,
       customID: this.subsetCustomId,
     })
@@ -142,6 +143,7 @@ export default class IfcIsolator {
       modelID: 0,
       scene: this.context.getScene(),
       ids: includedIds,
+      applyBVH: true,
       removePrevious: true,
       customID: this.subsetCustomId,
     })
@@ -297,6 +299,7 @@ export default class IfcIsolator {
         modelID: 0,
         scene: this.context.getScene(),
         ids: hidden,
+        applyBVH: true,
         removePrevious: true,
         customID: this.revealSubsetCustomId,
         material: this.hiddenMaterial,


oo-bldrs:
Hi @Ibrahim5aad, what's the performance delta between the version without BVH trees and the one with?

And does it make sense to expose this as a configurable option via the constructor, or is this something we always want to enable by default?

Ibrahim5aad:
> Hi @Ibrahim5aad, what's the performance delta between the version without BVH trees and the one with?
> 
> And does it make sense to expose this as a configurable option via the constructor, or is this something we always want to enable by default?

Hi @oo-bldrs, without setting this flag to true, the hiding/isolation operations cause a great decrease in the FPS ratio for the viewer, causing a very laggy model navigation.

Please check this video where I show by example this drop in FPS and how setting this flag to true is actually helping in better model navigation: https://www.veed.io/view/7aa3df52-7b77-47bb-bbbb-4f38a1003d3c?panel=share

oo-bldrs (approved):
LGTM

pablo-mayrgundter:
@ConorStokes and @nickcastel50 I'm going to start cc'ing you on some work in existing bldrs, esp performance.

Here, you can go to a complex model like:

  https://deploy-preview-743--bldrs-share.netlify.app/share/v/gh/IFCjs/test-ifc-files/main/Schependomlaan/IFC%20Schependomlaan.ifc/56/74/108/311035/408692

And select an item by double-clicking it, then use the 'i' key to isolate it.  I believe this is now faster by enabling BVH in IFCjs/Three.

I'll go ahead and approve and would appreciate an LGTM just to know you're aware of this track of work as we go.

pablo-mayrgundter (approved):

Adrian62D:
@pablo-mayrgundter We experience performance problems at a client with a model spanning 4 buildings. Currently Bldrs is too slow to handle it. Would it be possible to merge this PR so that we can include this in our next release next monday? If needed it could be hidden behind a feature flag.

ConorStokes:
> @ConorStokes and @nickcastel50 I'm going to start cc'ing you on some work in existing bldrs, esp performance.
> 
> Here, you can go to a complex model like:
> 
> https://deploy-preview-743--bldrs-share.netlify.app/share/v/gh/IFCjs/test-ifc-files/main/Schependomlaan/IFC%20Schependomlaan.ifc/56/74/108/311035/408692
> 
> And select an item by double-clicking it, then use the 'i' key to isolate it. I believe this is now faster by enabling BVH in IFCjs/Three.
> 
> I'll go ahead and approve and would appreciate an LGTM just to know you're aware of this track of work as we go.

LGTM

PR #747, opened 2023-06-16
Introduce screenshots, with feature flag, NoteCard button, volatile thumb.

pablo-mayrgundter:
Hi Ogali,

This PR is based off of your canvas-screenshot branch, to address https://github.com/bldrs-ai/Share/issues/733.

I found there was a confusing naming clash in our zustand viewer variable, so most of this is a refactor of that.

I also thought this was a small enough bit of functionality to just fit in the IfcApiExtended.

Working on cypress.

PR #749, opened 2023-06-16
Rename view and model variables in store to follow convention.

pablo-mayrgundter:
(This is in prep for a restarted version of https://github.com/bldrs-ai/Share/pull/747.)

I had changed the suffix for these to use *Store to help clarify these are not local React state variables, e.g. for useModel vs useModelStore in CadView.

However, it looked like this was causing confusion: there was now redundant methods in the store: setViewer and setViewerStore.  The better way to handle this is use uniform naming in the store, and use different local names as needed.


nickcastel50 (approved):
LGTM

Issue #748, opened 2023-06-16
cypress tests failing


pablo-mayrgundter:
Hi Ibrahim and Zien,

I just checked out from main and see broken cypress tests and seems to be related to Vyzn code.  Can you take a look and advise?

Thanks!
Pablo

```
> git fetch upstream
> git checkout -b ro upstream/main
> yarn cypress

<img width="1226" alt="image" src="https://github.com/bldrs-ai/Share/assets/2480879/5acbaf2d-efc9-4e1f-91c8-a1503629980f">

pablo-mayrgundter:
Couldn't reproduce.

Issue #733, opened 2023-05-31
Create screenshot of model


pablo-mayrgundter:
Ogali has this started here:

  https://github.com/bldrs-ai/Share/compare/main...oo-bldrs:Share:canvas-screenshot

- Needs a Share path endpoint to call for generating the screen
- Cypress test

PR #750, opened 2023-06-16
Introduce screenshot API, feature flag and use it in NoteCard.

pablo-mayrgundter:
Follows https://github.com/bldrs-ai/Share/pull/749, address FR https://github.com/bldrs-ai/Share/issues/733.

Demo:

  /share/v/p/index.ifc?feature=screenshot

Then open Notes, and should see screenshot button in note footer.  Click on it and should see screen added next to button.  It is not stored, will be overwritten on subsequent clicks.

Note: I added a cypress test which passes, but there are failing cy tests in main, from which I branched. https://github.com/bldrs-ai/Share/issues/748


nickcastel50 (approved):
LGTM

PR #772, opened 2023-08-04
Tabs

OlegMoshkovich:
This PR introduces a styled tabs component for the component library.

https://deploy-preview-772--bldrs-share.netlify.app/cosmos/?fixtureId=%7B%22path%22%3A%22src%2FComponents%2FLibrary%2FBldrsTabs.fixture.jsx%22%7D

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -15,6 +15,7 @@
   <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400&amp;display=swap" />
   <link href="https://fonts.googleapis.com/css2?family=Inter:wght@100;200;300;400;500;600&amp;display=swap"
     rel="stylesheet">
+  <link href="https://api.fontshare.com/v2/css?f[]=satoshi@700,500,300,400&display=swap" rel="stylesheet">
   <link rel="preconnect" href="https://fonts.googleapis.com" />
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 
--- a/public/index.html
+++ b/public/index.html
@@ -15,7 +15,6 @@
   <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400&amp;display=swap" />
   <link href="https://fonts.googleapis.com/css2?family=Inter:wght@100;200;300;400;500;600&amp;display=swap"
     rel="stylesheet">
-  <link href="https://api.fontshare.com/v2/css?f[]=satoshi@700,500,300,400&display=swap" rel="stylesheet">
   <link rel="preconnect" href="https://fonts.googleapis.com" />
   <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
 
--- a/src/Components/Library/BldrsTabs.jsx
+++ b/src/Components/Library/BldrsTabs.jsx
@@ -44,7 +44,7 @@ const CustomTab = styled((props) => <Tab disableRipple {...props}/>)(({theme}) =
 /**
  * Styled Tabs component.
  *
- * @property {Array} tabs array of tabs
+ * @property {Array<string>} tabs array of tabs
  * @return {React.Component}
  */
 export default function BldrsTabs({tabNames, actionCb}) {
--- a/src/Components/Library/BldrsTabs.jsx
+++ b/src/Components/Library/BldrsTabs.jsx
@@ -52,8 +52,8 @@ export default function BldrsTabs({tabNames, actionCb}) {
 
   const handleChange = (event, newValue) => {
     setValue(newValue)
+    actionCb(newValue)
   }
-
   return (
     <Box sx={{width: '100%'}}>
       <Box sx={{bgcolor: '#fff'}}>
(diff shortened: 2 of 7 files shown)


public/index.html
```diff
@@ -15,6 +15,7 @@
   <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto:wght@100;300;400&amp;display=swap" />
   <link href="https://fonts.googleapis.com/css2?family=Inter:wght@100;200;300;400;500;600&amp;display=swap"
     rel="stylesheet">
+  <link href="https://api.fontshare.com/v2/css?f[]=satoshi@700,500,300,400&display=swap" rel="stylesheet">
```
oo-bldrs, line 18: This font family is added to the index, meaning it's loaded by the browser each time the application is loaded.

However, this font family is not referenced by any stylesheets. Is it supposed to be in use, or can this be removed?

src/Components/Library/BldrsTabs.jsx
oo-bldrs, line 1: There should be one component per file.

src/Components/Library/BldrsTabs.jsx
```diff
@@ -0,0 +1,71 @@
...
+  '@media (max-width: 700px)': {
+    fontSize: '.7em',
+  },
+}))
+
+
+/**
+ * Styled Tabs component.
+ *
+ * @property {Array} tabs array of tabs
```
oo-bldrs, line 47: JSDoc type should be explicit that this is an array of strings, i.e. `Array<string>`.

src/Components/Library/BldrsTabs.jsx
```diff
@@ -0,0 +1,71 @@
...
+ * @return {React.Component}
+ */
+export default function BldrsTabs({tabNames, actionCb}) {
```
oo-bldrs, line 50: The `actionCb` argument is not used and should be removed.

src/Components/Library/BldrsTabs.jsx
```diff
@@ -0,0 +1,71 @@
...
+  const [value, setValue] = useState(0)
+
+  const handleChange = (event, newValue) => {
+    setValue(newValue)
+  }
+
+  return (
+    <Box sx={{width: '100%'}}>
+      <Box sx={{bgcolor: '#fff'}}>
+        <CustomTabs value={value} onChange={handleChange} aria-label="ant example">
```
oo-bldrs, line 60: The ARIA label should be taken as an argument (or left off altogether) rather than hard-coded to a non-unique value.

public/index.html
OlegMoshkovich: addressed.

src/Components/Library/BldrsTabs.jsx
OlegMoshkovich: Pulling a tab into its own file introduces extra logic responsible for selection of the current tab. 
I can see the way to pass in the value for the tab, but cannot figure out how to change currently selected tab.
https://mui.com/material-ui/api/tab/#Tab-prop-value
Seems to be ok for these components to be in the same file, since one would never use Tab individually.
lmk

src/Components/Library/BldrsTabs.jsx
OlegMoshkovich: I foresee the need to pass in the callback to the tabs that exposes the value of the selected tab to other components.
For example to drive the title of the dialog. 
The call back needs to accept the value and needs to be triggered when the tab is selected.

OlegMoshkovich (commented):
@oo-bldrs PTAL

src/Components/Library/BldrsTabs.jsx
oo-bldrs: I think that can be added at the time it's needed in order to give you some more room for designing the interface.

Specifically, it's unclear what value would be passed back. The tab index? The tab name? Both?

pablo-mayrgundter:
Looking good!

- Please add this link to PR description:

https://deploy-preview-772--bldrs-share.netlify.app/cosmos/?fixtureId=%7B%22path%22%3A%22src%2FComponents%2FLibrary%2FBldrsTabs.fixture.jsx%22%7D

- I think we ought to keep name simple, so just Tabs.jsx.  All of our components are ours..
- For the action, I think more convenient to use console.log instead of alert.  Should be easy to click through lots of components quickly

OlegMoshkovich:
Pablo, 
wanted to put all new library components in one place for easy of discovery. 
Will subtract the Bldrs from the name of the file, but will keep it in the name of the component,
 since the mui name is the same, or maybe you can suggest another way to deal with the conflict in the name.

OlegMoshkovich:
@oo-bldrs Please resolve all of the comments that are addressed.

src/Components/Library/BldrsTabs.jsx
oo-bldrs: The MUI documentation shows that there should be an attributed named `selected`, and its value should be `true` for the selected tab.

Has that worked?

pablo-mayrgundter:
heya, are you moving forward with this?

PR #787, opened 2023-09-13
Icon replace

OlegMoshkovich:
This PR replaces all of the custom icons with MUI icons.

### DIFF
--- a/src/Components/OpenModelControl.jsx
+++ b/src/Components/OpenModelControl.jsx
@@ -1,10 +1,11 @@
 import React, {useState, useEffect} from 'react'
 import {useNavigate} from 'react-router-dom'
 import Box from '@mui/material/Box'
+import Link from '@mui/material/Link'
 import MenuItem from '@mui/material/MenuItem'
-import Paper from '@mui/material/Paper'
-import Typography from '@mui/material/Typography'
+import Stack from '@mui/material/Stack'
 import TextField from '@mui/material/TextField'
+import Typography from '@mui/material/Typography'
 import useTheme from '@mui/styles/useTheme'
 import {useAuth0} from '@auth0/auth0-react'
 import Dialog from './Dialog'
@@ -27,7 +28,6 @@ export default function OpenModelControl({fileOpen}) {
   const [isDialogDisplayed, setIsDialogDisplayed] = useState(false)
   const [orgNamesArr, setOrgNamesArray] = useState([''])
   const {user} = useAuth0()
-  const theme = useTheme()
   const accessToken = useStore((state) => state.accessToken)
   useEffect(() => {
     /**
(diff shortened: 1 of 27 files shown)


src/theme/Colors.js
```diff
@@ -12,7 +12,7 @@ export function getColors() {
       light: grey[300],
       medium: '#C1C1C1',
       dark: '#444444',
-      darkest: grey[900],
```
pablo-mayrgundter, line 15: Haven't looked at these in a while, but I'm reminded that their palettes are supposed to be balanced, so arguably we should just replace all hard-coded numbers with grey[] selectors.

Would the greys here work for you?

https://mui.com/material-ui/customization/color/#color-palette

Besdies grey, only exception I'd make now is keeping our logo green on purpose to make it stick out of pallette :)

src/theme/Palette.js
```diff
@@ -7,13 +7,13 @@ const colors = getColors()
 export const day = {
   mode: 'light',
   primary: {
-    main: colors.grey.medium,
+    main: '#6D8752',
```
pablo-mayrgundter, line 10: Same here.. I think we should avoid deviation from the palette combos in Material if possible

src/theme/Palette.js
```diff
@@ -29,7 +29,7 @@ export const day = {
 export const night = {
   mode: 'dark',
   primary: {
-    main: colors.grey.dark,
+    main: '#6D8752',
```
pablo-mayrgundter, line 32: Again

src/theme/Theme.jsx
```diff
@@ -53,9 +52,8 @@ function loadTheme(mode, setMode, themeChangeListeners) {
   // https://mui.com/customization/dark-mode/
   const activePalette = mode === Themes.Day ? day : night
   const theme = {
-    components: getComponentOverrides(activePalette, getTypography()),
-    typography: getTypography(),
-    shape: {borderRadius: 8},
+    components: getComponentOverrides(activePalette),
+    shape: {borderRadius: 0},
```
pablo-mayrgundter, line 56: In your previous PR I question removing border radius.. let's figure that out and do whatever we decide here

pablo-mayrgundter (changes requested):
Lots of simplification!  There's some other topics addressed in the PR too.. so think we should discuss and then ready shortly

src/Components/OpenModelControl.jsx
```diff
@@ -1,11 +1,14 @@
 import React, {useState, useEffect} from 'react'
 import {useNavigate} from 'react-router-dom'
 import Box from '@mui/material/Box'
+import CreateNewFolderOutlinedIcon from '@mui/icons-material/CreateNewFolderOutlined'
+import InputLabel from '@mui/material/InputLabel'
+import Link from '@mui/material/Link'
 import MenuItem from '@mui/material/MenuItem'
-import Paper from '@mui/material/Paper'
```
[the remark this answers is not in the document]
OlegMoshkovich: Oh - that is coming from the other style pr - since I branched off of that.

src/theme/Colors.js
OlegMoshkovich: Yeah totally

src/theme/Palette.js
OlegMoshkovich: This one is primary green - mui greens are not so nice

PR #792, opened 2023-09-23
Upgrade Cypress

oo-bldrs:
PR to upgrade Cypress to the latest version (v13) and upgrade the Cypress GitHub action to its latest version (v6).

This resolves an issue where Cypress <= 12.15 is used with Chrome >= 117, and all tests fail.


oo-bldrs:
@OlegMoshkovich This will address the issue that's blocking your builds.

nickcastel50 (approved):
LGTM

PR #796, opened 2023-10-01
Notes mobile fix

OlegMoshkovich:


OlegMoshkovich:
@oo-bldrs in this pr I subtracted expand on element select from nav panel. It feels distracting, particularly on mobile. I think it broke one of the cypress tests.
image
I re-ran the job several times.
Can you please take a look when you have a chance.

oo-bldrs:
@OlegMoshkovich This is the test that is failing.

- https://github.com/OlegMoshkovich/Share/blob/ea01f3f7867effc3d48bd0cbadabd06b60deb243/cypress/e2e/integration/bldrs-inside-iframe.cy.js#L84

Does your change affect the text that is displayed in that scenario?

OlegMoshkovich:
I don't think so.

pablo-mayrgundter (approved):

PR #799, opened 2023-10-04
Update workflow names and artifact IDs

oo-bldrs:
PR to update the workflow names and artifact IDs to avoid collisions between multiple builds.

@nickcastel50 This copies your changes from your branch to minimize the size of the conflict when you go to merge.


nickcastel50 (approved):
LGTM

PR #801, opened 2023-10-06
Version History Panel

OlegMoshkovich:
This PR adds the Version History Panel component.
- Displays versions in a chronological timeline format.
- Highlights specific version details, including the author, commit date, and commit message.


pablo-mayrgundter (approved):
Per convo, disable changing the selection so this is read-only.

PR #811, opened 2023-10-13
Geometry Coverage

nickcastel50:
Mainly a rollup of conway/65 and conway-geom/66 for gaps with web-ifc@0.0.44.


pablo-mayrgundter (approved):
I think this just needs a merge of that API typo.

I was going to suggest changing the version number, but I'll follow-up with my changes from the auto-version script and update issues referring to that version.

PR #812, opened 2023-10-14
re order the operations group

OlegMoshkovich:
This PR introduces a button group that becomes active when an element is selected. The button group contains  properties, isolation, and clear actions.


src/Containers/CadView.test.jsx
pablo-mayrgundter, line 1: Love the change, but should increase test coverage

pablo-mayrgundter (changes requested):

PR #819, opened 2023-10-20
Project root dir cleanup

pablo-mayrgundter:
- Move configs for esbuild, jest and tsc to `tools` subdir.
- Remove unused files and dirs.

Manually tested build, lint, serve, test, conway variants

Based on #817


oo-bldrs:
FYI, moving configuration files will break any IDEs or their plugins that rely on standard search paths for Jest and TypeScript.

PR #820, opened 2023-10-20
Esbuild 0.17.19

pablo-mayrgundter:
This was a breaking change that required some API changes to the esbuild.serve call, now `ctx = esbuild.context` followed by `ctx.serve`.

Based on #819


pablo-mayrgundter:
Done for the evening.. here's where I got to.  I see this both with `yarn serve` and `cd docs; npx http-server`

image

oo-bldrs (approved):

PR #824, opened 2023-10-25
Remove global Google Analytics and simplify privacy package.

pablo-mayrgundter:
- Don't even initialize GA when analytics user-pref is disabled.
- Changed gtags to react-ga4 npm.  Update to use our GA4 ID.
- Simplify privacy package: use js-cookie npm and use simple wrappers for cookies we use: isAnalyticsEnabled, isFirstTime and theme preferences.

Tested locally by disabling msw and watching traffic in GA4 dashboard.  Not showing up in netlify, but I do see netlify as a suggested domain in GA4 dashboard, so perhaps GA is just not turning itself on unless it's in a configured domain.

I suggest pushing this live and I verify it's live there.


src/__mocks__/api-handlers.js
```diff
@@ -13,202 +13,240 @@ const httpCreated = 201
...
-    const ref = req.url.searchParams.get('ref')
-
-    if (org !== 'bldrs-ai' || repo !== 'Share' || path !== 'README.md') {
+
+/**
+ * Initialize API handlers, including Google Analytics and GitHub.
+ *
+ * @return {Array<object>} handlers
+ */
+export function initHandlers() {
```
oo-bldrs, line 22: Why do these need to be functions?

Is there a reason the spread operator can't be used here for concatenating two arrays?

src/privacy/analytics.js
```diff
@@ -1,21 +1,24 @@
-import gtag from '../utils/gtag'
+import ReactGA from 'react-ga4'
 import * as Functional from './functional'
 import * as Privacy from './Privacy'
 import {assertDefined} from '../utils/assert'
 
 
+ReactGA.initialize('G-GRLNVMZRGW')
```
oo-bldrs, line 7: Does the initialization need to be unconditional?

Can the production value be injected via an environment variable (if present), rather than hard-coded?

src/__mocks__/api-handlers.js
pablo-mayrgundter: They were defined at root scope, so to use spreading (my first impulse too) would mean joining them at the end of the file, which I thought was less readable.

src/privacy/analytics.js
pablo-mayrgundter: Actually, I take it back.  I rewrote the privacy system.  The old setup assumed a global GA enable, which was for separate usage vs. social, which we didn't make use of.  Functional cookies are now just set directly in `firstTime` and `preferences`, and `analytics` does a lazy init only if enabled, which is true by default or until the PrivacyControl toggle.

pablo-mayrgundter:
PTAL

pablo-mayrgundter:
@OlegMoshkovich fyi too.  I apologize to both of you for the ugly cookie/privacy system before! 😬

oo-bldrs (approved):

PR #829, opened 2023-10-30
Remove network fonts.

pablo-mayrgundter:
I had a draft (https://github.com/bldrs-ai/Share/pull/825) setup for this change, but it got out of sync.

I don't see any significant diffs in fonts and the network load for them makes me sad.  We can revisit later if we want to use something visually distinct.

Using screens from Nick and I just testing on it.  

Draft PR on Mac:
image

Draft PR on Windows:
image

Prod win:
image

Prod Mac:
image

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -3,20 +3,13 @@
 
 <head>
   <meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


nickcastel50 (approved):
LGTM

PR #831, opened 2023-11-03
Add tests for file loader function

OlegMoshkovich:
This PR introduces a Jest test suite for the file loader utility. The tests validate the expected behavior when loading local models.

### DIFF
--- a/src/utils/loader.js
+++ b/src/utils/loader.js
@@ -9,7 +9,7 @@ import {assertDefined} from '../utils/assert'
  * @param {string} appPrefix
  * @param {Function} handleBeforeUnload
  */
-export function loadLocalFile(navigate, appPrefix, handleBeforeUnload) {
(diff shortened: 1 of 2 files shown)


src/utils/loader.js
```diff
@@ -9,7 +9,7 @@ import {assertDefined} from '../utils/assert'
  * @param {string} appPrefix
  * @param {Function} handleBeforeUnload
  */
-export function loadLocalFile(navigate, appPrefix, handleBeforeUnload) {
+export function loadLocalFile(navigate, appPrefix, handleBeforeUnload, skipAutoRemove = false) {
```
pablo-mayrgundter, line 12: if that's there for only for the test, then I'd prefix it like testing_skipAutoRemove

pablo-mayrgundter (approved):

PR #832, opened 2023-11-04
Add Element group for selected element

OlegMoshkovich:
This PR introduces a new "Element Group" feature at the bottom of the screen. 
The group encompasses controls for the following functionalities:

Cutplane
Isolating selected element
Hiding selected element
Deselecting an element
Unhiding all of the hidden elements

Behavioral Notes:
The majority of these controls become active only when an element is selected.
In its neutral or default state, only the cutplane control is visible and accessible.


pablo-mayrgundter:
w00t!

- Love isolate
- Feeling a little bare.. maybe we should start with nav open and root elt selected so that props are visible?  And maybe no in mobile if it's too crowded there
- Props and info have same icon.  Can we go back to usual hamburger props icon?  it's very recognizable to me
- Clear is both in the Section menu and in the Element button group, but not when just Sectioned.  Can you have it always be part of Element group when clear is valid?
- Did Help tooltips get cleaned up?  They look nice!

Did a first scan of code.  Looking good.

pablo-mayrgundter:
Draft or no?

pablo-mayrgundter:
Per convo, starting with Nav open we're going to punt.

pablo-mayrgundter (approved):

PR #857, opened 2023-11-14
Disable GA as a workaround for prod freeze bug.

pablo-mayrgundter:
Discord #ops diagnosis: 
https://discord.com/channels/853953158560743424/1171855282205380608/1174108348749205624

```
turning off shields in brave fixes it for me
that may be related to the cookie refactor i did
and somehow tickled by the About component reorg
yes indeed. once i disable shield, page loads, then i toggle cookie, and now after re-enabling shield, no longer seeing GA blocked and also can spin again
so right, after the cookie refactor, none of the GA tracking, incl functional, are enabled until About is shown

Maybe in the way the About code is structured that cookie is touched somehow, even tho the dialog isn't visible
aha, and GA is disabled via msw in local, which is why we couldn't repro it locally
ta-da
ok, so fix is probably to disable GA in prod, get it repro'd locally, go from there
```


OlegMoshkovich (approved):
LGTM

PR #872, opened 2023-11-27
Add run-script-os package via yarn + fix windows compilation 

nickcastel50:
* This PR address fixing windows compilation, the syntax we were using for setting the USE_WEBIFC_SHIM environment variable wasn't valid Windows syntax.


pablo-mayrgundter (commented):
LGTM but will handoff to @oo-bldrs

pablo-mayrgundter (approved):
Talked w/nick, he's confident so we'll go ahead

PR #878, opened 2023-12-04
Panel update

OlegMoshkovich:
This pr unifies styles of the panel with the rest of the components.


pablo-mayrgundter (approved):
Would prefer em to px

PR #917, opened 2024-01-09
Refactor fixtures to be used in tests.

pablo-mayrgundter:
This was work from before holiday break to improve our fixture and unit test system.

- Remove kitchen-sink FixtureContext in favor of fine-grained contexts.
- Use Fixtures to define default component-under-test in unit tests.
- Many more fixtures.
- Simpler unit tests.

PR #925, opened 2024-01-16
Replace FixtureContext with more fine-grained contexts.

pablo-mayrgundter:
First PR pulling work from https://github.com/bldrs-ai/Share/pull/917


OlegMoshkovich:
image
There is this message - which I think is fine, but just want to bring it to your attention.
image
About dialog is broken.
The rest is good!

PR #936, opened 2024-01-16
New fixture and unit test for Logo.  From #917.

pablo-mayrgundter:


pablo-mayrgundter:
Hmm, not sure what's up here.. i fixed the cypress test that's failing and pushed it up but don't see the changes

PR #960, opened 2024-02-02
Timeline load - fixing disappearing future

pablo-mayrgundter:
We had been discovering commits with `/commits?sha=<hash>` which gets commits up to that sha.

This PR switches to use `/commits?path=<path to file>` to get all commits related to that file.

It also adds some logic to compare the current URL's file ref to the commit ref in the Timeline to highlight the active commit within the list.


pablo-mayrgundter:
All set!

OlegMoshkovich (approved):

PR #974, opened 2024-02-07
Comment Create

OlegMoshkovich:
This PR adds create comment card when a note is selected.


OlegMoshkovich:
image

Opening share dialog opens side drawer.

OlegMoshkovich:
image

Buttons are hidden.

OlegMoshkovich:
image


Pablo, as you cherry pick can you also take at the octakit caching issue, this Comment stuff is affected by it I think.
I have been struggling with it the whole week, sometimes it goes through and sometimes it does not, but requests always go through.

PR #975, opened 2024-02-08
Fix msw config and disable for NODE_ENV=production

pablo-mayrgundter:
Default is to use MSW to isolate dev environment for API calls.  Model loads are still live.
```
yarn serve
```
image

Setting to production disables MSW.
```
NODE_ENV=production yarn serve
```
image


nickcastel50 (approved):
LGTM

PR #987, opened 2024-02-18
Remove canvas from jest and pkgs.

pablo-mayrgundter:
I was setting up a new user on my machine, new brew install etc and didn't have canvas system libs installed.  Didn't know this until I went to run `yarn test`, which failed without them.  That reminded me that during the work on headless-three, I think I figured out how not to use it and must have fixed that in Share since, as I'm able to remove canvas now and server + test both work.

This removes a pretty heavy req for others too.


nickcastel50 (approved):
LGTM

PR #989, opened 2024-02-20
Auth0 End To End Cypress Tests + screenshot / video generation for failed tests

nickcastel50:
This PR has been updated for the following:

Auth0 End To End Cypress Tests
Screenshot / Video artifact generation for failed tests
One Save IFC test.

[Profile: 100 Login Test](https://github.com/bldrs-ai/Share/issues/1052)


cypress/e2e/saveModelControl/save-model.cy.js
```diff
@@ -155,7 +163,7 @@ describe('save model', () => {
         cy.wait('@authorizeRequest').its('response.statusCode').should('eq', STATUS_OK)
         cy.wait('@tokenRequest').its('response.statusCode').should('eq', STATUS_OK)
         // Commented out for now
-        // cy.findByTestId('Save IFC', {timeout: 60000}).should('exist')
+        cy.findByTestId('Save IFC', {timeout: 60000}).should('exist')
```
pablo-mayrgundter: I feel like these should be short, e.g. 5s.  If we don't have the popup almost immediately, it's bc it's not going to work.  Or that's how it's been for me with the dialogs i was testing recently

if so, the length for these timeouts should be smth like "to make tests that are going to fail, fail faster, but no so short that we get flaky behavior"

src/store/AppSlice.js
```diff
@@ -1,3 +1,6 @@
+const isOpfsEnabled = (process.env.OPFS_IS_ENABLED || 'false').toLowerCase() === 'true'
```
pablo-mayrgundter: I think you need to merge?  There's a new condition here based on our convo yesterday https://github.com/bldrs-ai/Share/blob/main/src/store/AppSlice.js

.gitignore
```diff
@@ -15,3 +15,7 @@ cypress/downloads
 *.log
 *~
 \#*
+
+#do not check in cypress videos or screenshots
+cypress/videos/*
+cypress/screenshots/*
```
pablo-mayrgundter: per convo, pls remove

pablo-mayrgundter (changes requested):

.gitignore
nickcastel50: PTAL

src/store/AppSlice.js
nickcastel50: PTAL

cypress/e2e/saveModelControl/save-model.cy.js
nickcastel50: PTAL

PR #993, opened 2024-02-22
Lock packages versions, update yarn.lock.  Remove null console output…

pablo-mayrgundter:
… during esbuild unlink of msw.


nickcastel50 (approved):
LGTM

PR #995, opened 2024-02-22
Fix for broken model load on private LFS from break in OPFS, also Ver…

pablo-mayrgundter:
…sions would have been broken for using finalUrl (LFS URL) for branch path composition.  Fix Notes/Octokit race on non/authed init.


nickcastel50 (approved):
LGTM

PR #1009, opened 2024-02-28
Address private model load with no AuthToken

nickcastel50:
Added try catch + octokit request timeout for getLatestCommitHash to address private model load with no authToken issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.0.906",
+  "version": "1.0.907",
   "main": "src/index.jsx",
   "license": "MIT",
   "homepage": "https://github.com/bldrs-ai/Share",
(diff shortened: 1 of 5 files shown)


pablo-mayrgundter (changes requested):
Per convo, I'd move the new functions in GitHub.js to use the helpers (getGitHub, postGitHub, ...) and instead of doing local error handling, let the exception bubble up and catch it at the call site in CadView

nickcastel50:
> Per convo, I'd move the new functions in GitHub.js to use the helpers (getGitHub, postGitHub, ...) and instead of doing local error handling, let the exception bubble up and catch it at the call site in CadView

PTAL: Updated getLatestCommitHash to use the getGithub function, bubbled error up to CadView, and wrote a failure case test that expects an error thrown (passing 'failurecaseowner' / 'failurecaserepo' can trigger a throw)

pablo-mayrgundter (approved):

Issue #1013, opened 2024-02-29
OPFS sha is null in at least one test private model load


pablo-mayrgundter:
Seeing this for many models

eg..
https://github.com/bldrs-ai/test-models/blob/main/ifc/misc/haus.ifc

<img width="990" alt="image" src="https://github.com/bldrs-ai/Share/assets/2480879/6ee6a608-f1f1-4af8-8abc-5b32475b38f9">

PR #1015, opened 2024-02-29
ISSUE_1003 Fix bug with parameter set up in getLatestCommitHash

nickcastel50:
Fixes [ISSUE 1003](https://github.com/bldrs-ai/Share/issues/1013)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.0.908",
+  "version": "1.0.909",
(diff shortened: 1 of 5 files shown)


pablo-mayrgundter:
Per convo, we should throw on "200 not found" this seems to be signaling, and should also harden the path over to where a null was being accepted as a sha value, e.g. with assertDefined there.

Also add unit tests at both sites.

pablo-mayrgundter (approved):

Issue #1017, opened 2024-03-04
Enhance OPFS availability detection


nickcastel50:
Enhance OPFS availability detection - specifically around private browser's lack of OPFS support in many browsers.

PR #1018, opened 2024-03-04
ISSUE1017 - Enhance OPFS Availability Detection

nickcastel50:
Satisfies [Issue 1017](https://github.com/bldrs-ai/Share/issues/1017)

### DIFF
--- a/src/Containers/CadView.jsx
+++ b/src/Containers/CadView.jsx
@@ -74,8 +74,6 @@ export default function CadView({
 }) {
   assertDefined(...arguments)
 
-  const isOPFSAvailable = checkOPFSAvailability()
-
   const {setFile} = useContext(FileContext) // Consume the context
   debug().log('CadView#init: count: ', count++)
   // React router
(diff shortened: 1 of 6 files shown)


src/Containers/CadView.jsx
```diff
@@ -301,6 +312,9 @@ export default function CadView({
 
   /** When viewer is ready, load IFC model. */
   async function onViewer() {
+    if (isOPFSAvailable === null) {
```
pablo-mayrgundter, line 315: assertDefined instead

src/OPFS/utils.js
```diff
@@ -303,10 +303,15 @@ export function loadLocalFileDragAndDrop(
...
-export function checkOPFSAvailability() {
-  // Check for FileSystemDirectoryHandle availability
+export async function checkOPFSAvailability() {
   if ('FileSystemDirectoryHandle' in window) {
-    return true
+    try {
+      await navigator.storage.getDirectory()
+      return true
+    } catch (error) {
+      debug().error(`OPFS error: ${ error}`)
```
pablo-mayrgundter, line 312: Add a comment:
```
// Expected for browsers XX ...

pablo-mayrgundter (changes requested):
Per convo, move OPFS check state into store so no need to route it with params, and also initialize it in BaseRoutes

src/Containers/CadView.jsx
nickcastel50: I don't think we want to assertDefined there, as onViewer can be called possibly after isOPFSAvailable resolves, in which case it will still be null.

src/OPFS/utils.js
nickcastel50: Updated, PTAL

src/Containers/CadView.jsx
pablo-mayrgundter: Ok, then return

pablo-mayrgundter (approved):
Made two small changes:
  - onViewer returns early if isOPFSAvailable is null
  - Lint fix for isOPFSAvailable in BaseRoutes useEffect deps

PR #1021, opened 2024-03-07
Prepare for new web-ifc-viewer build

pablo-mayrgundter:
Remove custom setup for screesshot leftover from AEC hackathon.  Use existing IFCjs method directly.

Test: click imagine button and verified screenshot working


nickcastel50 (approved):
LGTM

Issue #1031, opened 2024-03-11
View 100: Initial model load and view


pablo-mayrgundter:
The screen setup on the initial model view should show the model, include all control groups, with an About dialog for first time users and not after that.

# Tasks
In the following, list items are Context/Actions, checkboxes are Expects. Nesting denotes dependency and items on the same level are independent.  "Screen" for screenshot in percy

* Returning user
  * Visits homepage
    - [x] See logo model, model title and all main controls - Screen
  * Visits about permalink
    - [x] See AboutDialog - Screen
 
* First-time user
  * Visits homepage
    - [x] See AboutDialog - Screen
    * Closes AboutDialog
      - [x] AboutDialog not visible, model visible - Screen

# Tests
[Cypress](https://github.com/bldrs-ai/Share/blob/main/cypress/e2e/view-100/initial-model-load-and-view.cy.js)
[Screens](https://percy.io/8fe2b2f1/share/builds/33552754/changed/1838718781)

PR #1050, opened 2024-03-12
1031 initial model load and view

pablo-mayrgundter:
Tests for #1031, mostly by screenshot.

- Replace cypress e2e github action script with basic one from template.  Just builds Share to ./docs then runs cypress as configured in our cypress.config.js
- Take screenshot test and add tool for diffing (tools/imgdiff.sh).  Will need to put extra setup commands on wiki.
- CadView: wait on cameraControls 'rest' event to signal model loaded


nickcastel50 (approved):
LGTM

PR #1065, opened 2024-03-22
percy: convert a view-100 test case to percy screenshot

pablo-mayrgundter:
This PR introduces the percy.io screen testing library.

Percy run for this PR: https://percy.io/8fe2b2f1/share/builds/33552754

There's an important change to three.js (in web-ifc-viewer) to show the rendered model 3d buffer: preserveDrawingBuffer enabled to fix screen capture.  Also needed to include some build flags to enable it only in cypress, as it's a potential performance hit if left on all the time.


nickcastel50 (approved):
LGTM

PR #1094, opened 2024-04-05
Remove confusing testId property

pablo-mayrgundter:
We are using testing-library's special `data-testid` prop for finding elements in unit tests, as in https://testing-library.com/docs/queries/bytestid/

Over time a non-standard `testId` has crept into mixed usage.  Sometimes we were using it to pass the value for the data-testid in our new components, but then this was being confused with being a valid prop on a react component.

Testing-library seems to know about this as it emits the warning below if you do this.

This PR replaces this usage with the following cases.  testId for passing a props is now replaced by:
  - prop drilling of data-testid if value is just passed to a sole subcomponent
  - namedTestId if the id is targeted to one subcomponent in a complex case (e.g. as buttonTestId the Button in TooltipIconButton)

image

### DIFF
--- a/src/Components/Buttons.jsx
+++ b/src/Components/Buttons.jsx
@@ -14,7 +14,8 @@ import BackIcon from '../assets/icons/Back.svg'
 
 
 /**
- * An icon button with a tooltip.
+ * An icon button with a tooltip.  THe button will use a given buttonTestId or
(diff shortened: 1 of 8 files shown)


nickcastel50 (approved):
LGTM

PR #1098, opened 2024-04-08
gha: remove percy from cypress run

pablo-mayrgundter:
I was testing this setup and inadvertently checked it in.

Percy should be run locally only for now


nickcastel50 (approved):
LGTM

Issue #1096, opened 2024-04-08
Permalink to individual notes broken


OlegMoshkovich:
When a user accesses a link that includes a note hash ID, intending to open a specific individual note, share's side drawer behavior is not as expected. 
Instead of navigating directly to the individual note view associated with the hash ID, the side drawer opens to the general list view. 

Refer to #1072  for the UX flow

https://bldrs.ai/share/v/gh/Swiss-Property-AG/Schneestock-Public/main/ZGRAGGEN.ifc#1340217041:undefined;c:80.66,11.66,-94.06,6.32,2.93,-8.72;i:

When accessing a shared note the URL changes to include undefined in the string and the number of id is subtracted.

OlegMoshkovich:
I implemented a useEffect in the notes controller that detects an ID if it's appended to the notes hash. However, when the Share feature identifies a link containing a hash, it incorrectly removes the number from the hash.

PR #1102, opened 2024-04-09
Sidedrawer navigation problem when accessing individual notes via hash ID #1096

OlegMoshkovich:
#1096


pablo-mayrgundter:
Addressed in my notes fixes yesterday, except NoteCard footer share link is still [Object object].. will handle separately

PR #1104, opened 2024-04-10
Refactor view-100 initial model load and view

pablo-mayrgundter:
- closer match with story issue https://github.com/bldrs-ai/Share/issues/1031
- add expects for all control buttons on main screen
  - merge in the controbutton pr to get their data-testids
- start reorganizing cypress/e2e/support/utils

# Screen diffs
https://percy.io/8fe2b2f1/share/builds/33611784

I think these are just name changes.. I'm playing around with using context more and having the `it` methods just have expects


nickcastel50 (approved):
LGTM

PR #1105, opened 2024-04-11
Rename contexts and expects to match story tasks

pablo-mayrgundter:
Now matches https://github.com/bldrs-ai/Share/issues/1031

https://percy.io/8fe2b2f1/share/builds/33644960


OlegMoshkovich:
@pablo-mayrgundter format looks clear. Percy is failing.

Issue #1072, opened 2024-03-26
Notes 100 - Access shared note


OlegMoshkovich:
As a user, I want to access the notes shared with me via the link so that I can see the information and the 3d model view of the project.

# Tasks
    
  * Access the shared link containing the path to the general notes
    - [x] Correct project to be loaded into the viewport and side drawer to be open - Screen
      - [x] Panel title to contain 'NOTES' string
      - [x] List of notes to be visible
  * Access the shared link containing the path to a note index
    - [x] Correct project to be loaded into the viewport and side drawer to be open - Screen
      - [x] Panel title to contain 'NOTE' string
      - [x] Panel nav to contain back button
      - [x] Shared note to be visible 
      - [x] Notes comments to be visible
  

# Tests
[Cypress](https://github.com/bldrs-ai/Share/blob/main/cypress/e2e/notes-100/access-shared-note.cy.js)
[Percy](https://percy.io/8fe2b2f1/share/builds/33846447/changed/1853375510)

PR #1119, opened 2024-04-15
1072 - Notes 100 - Access shared note

OlegMoshkovich:
#1072


cypress/e2e/notes-100/access-shared-note.cy.js
```diff
@@ -1,16 +1,39 @@
 import {waitForModel, homepageSetup} from '../../support/utils'
 
-
-describe('select-a-note', () => {
-  context('Open the link with a shared note', () => {
+// From https://github.com/bldrs-ai/Share/issues/1072
+describe('Notes 100 - Access shared note', () => {
```
pablo-mayrgundter, line 4: - Use utils.js#setIsReturningUser instead of setCookie (setCookie semantics are hard to understand.. the 1 is a serial, not a bool, and will change). 

- Since you're doing both homepageSetup and setIsReturningUser, for all tests here, move them together under describe
```
describe('Notes 100 - Access shared note', () => {
  beforeEach(() => {
    homepageSetup()
    setIsReturningUser()
  })
```

cypress/e2e/notes-100/access-shared-note.cy.js
```diff
@@ -1,16 +1,39 @@
...
+  context('Access shared link containing the path to the general notes', () => {
     beforeEach(() => {
```
pablo-mayrgundter, line 6: now this one should just be the visit and wait:
```
beforeEach(() => {
  cy.visit('/share/v/p/index.ifc#c:-133.022,131.828,161.85,-38.078,22.64,-2.314;i:')
  waitForModel()
})
```

cypress/e2e/notes-100/access-shared-note.cy.js
```diff
@@ -1,16 +1,39 @@
...
+    it('Panel title to contain NOTES string', () => {
       cy.get('[data-testid="panelTitle"]').contains('NOTES')
     })
+    it('Notes list to be visible', () => {
+      cy.get('.MuiList-root').should('exist')
+    })
+  })
+  context('Access the shared link containing the path to a note index', () => {
+    beforeEach(() => {
+      homepageSetup()
```
pablo-mayrgundter, line 21: remove homepage and setcookie

pablo-mayrgundter (changes requested):
Story issue looks good.  Comments below are mostly about using the helpers consistently with what I've vetted between local & GHA.

cypress/e2e/notes-100/access-shared-note.cy.js
```diff
@@ -1,16 +1,39 @@
...
+  context('Access shared link containing the path to the general notes', () => {
     beforeEach(() => {
       homepageSetup()
-    })
-    it('Visit notes permalink, side drawer shall be opened', () => {
       cy.setCookie('isFirstTime', '1')
       cy.visit('/share/v/p/index.ifc#c:-133.022,131.828,161.85,-38.078,22.64,-2.314;i:')
       waitForModel()
+    })
+    it('Panel title to contain NOTES string', () => {
```
pablo-mayrgundter, line 12: Still needs percySnapshot.

On mine I've been noticing they're running slow.

What do you think about this instead?

```
it('Correct project to be loaded into the viewport and side drawer to be open - Screen', () => {
  // Panel title to contain 'NOTES' string
  cy.get('[data-testid="panelTitle"]').contains('NOTES')

  // List of notes to be visible
  cy.get('.MuiList-root').should('exist')
  
  cy.percySnapshot()
})

File: cypress/e2e/notes-100/access-shared-note.cy.js [hunk 5]
```diff
@@ -1,16 +1,39 @@
 import {waitForModel, homepageSetup} from '../../support/utils'
 
-
-describe('select-a-note', () => {
-  context('Open the link with a shared note', () => {
+// From https://github.com/bldrs-ai/Share/issues/1072
```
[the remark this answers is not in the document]
OlegMoshkovich: addressed.

cypress/e2e/notes-100/access-shared-note.cy.js
OlegMoshkovich: I see. Cool. Addressed.

cypress/e2e/notes-100/access-shared-note.cy.js
OlegMoshkovich: Ah ok, makes sense to split into two different methods.

cypress/e2e/notes-100/access-shared-note.cy.js
OlegMoshkovich: That's cool, I was thinking we should have 'IT' for each expect, but it makes sense to combine these two ITs together since it points to the setup and the screenshot should capture the baseline state.

OlegMoshkovich: I like this format, I suppose it is judgement call, also we should include comments in this format?

cypress/e2e/notes-100/access-shared-note.cy.js
OlegMoshkovich: But we need it to setup for the context, no?

cypress/e2e/notes-100/access-shared-note.cy.js
```diff
@@ -1,16 +1,39 @@
...
     })
+    it('Notes list to be visible', () => {
+      cy.get('.MuiList-root').should('exist')
+    })
+  })
+  context('Access the shared link containing the path to a note index', () => {
+    beforeEach(() => {
+      homepageSetup()
+      cy.setCookie('isFirstTime', '1')
+      cy.visit('http://localhost:8080/share/v/p/index.ifc#c:-26.91,28.84,112.47,-22,16.21,-3.48;i:2')
```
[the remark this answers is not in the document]
OlegMoshkovich: addressed, in the seprated methods, let me know if this is correct. sorry if I misundertoon.

cypress/e2e/notes-100/access-shared-note.cy.js
```diff
@@ -1,16 +1,39 @@
...
+      waitForModel()
+    })
+      it('Panel title to contain NOTE string', () => {
+        cy.get('[data-testid="panelTitle"]').contains('NOTE')
```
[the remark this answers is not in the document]
OlegMoshkovich: Well it is the same test ID panelTitle, the title should change to NOTE instead of NOTES. 
Maybe a different cypress method should be utilized here.

src/Components/Notes/NotesControl.jsx
```diff
@@ -73,6 +74,18 @@ export default function NotesControl() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [model, isCreateNoteVisible])
 
+  useEffect(() => {
+    const noteNumberStr = getHashParams(location, NOTES_PREFIX) // Assuming this returns something like "i:123"
```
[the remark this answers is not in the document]
OlegMoshkovich: I did not add it.. I was also a bit surprised it appeared here, it was a result of the rebase.
Maybe something did not connect properly. Will delete
Sorry for that.

cypress/e2e/notes-100/access-shared-note.cy.js
pablo-mayrgundter: The way you have it here is necessary, but the idea was to have a common beforeEach for homepage+returning use for both contexts

In general, you shouldn't have 2 beforeEach's in a row, like this

cypress/e2e/notes-100/access-shared-note.cy.js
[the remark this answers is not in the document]
pablo-mayrgundter: Let's talk about this.. catch you on discord

pablo-mayrgundter (approved):

cypress/e2e/notes-100/access-shared-note.cy.js
pablo-mayrgundter: Comments help when the selectors are generic, like .MuiList-root is ambiguous to me without going and reading the code under test and Mui api to remember how this works.  e2e tests should be more black-box, which is why i've been moving towards data-testids that reflect use-perceivable types

PR #1131, opened 2024-04-17
Fix 1130-select-element-doesnt-open-nav

pablo-mayrgundter:
e2e test coming in next pr

# NavTreeControl
- Add useEffect[selectedElements]
# CadView
- Move useEffect[selectedElements] later in initialization (and down file).  This is necessary to have elementsById setup correctly.
- Trigger selectElementBasedOnFilePath on rootElement instead of model to avoid race
# Theme
- re-enable Mui-Selected coloring for NavTreeItem
# Util
- Add assertNumber and test

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.0.984",
+  "version": "1.0.985",
   "main": "src/index.jsx",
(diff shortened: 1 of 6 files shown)


nickcastel50 (approved):
LGTM

PR #1133, opened 2024-04-19
Introduce cy-parallel pkg script

pablo-mayrgundter:
Also tune cypress pageLoadTimeout from 60s to 2.5s to speedup broken tests.

Sample run:
image


nickcastel50 (approved):
LGTM

Issue #1166, opened 2024-05-02
Caching on GitHub API is causing issues saving models within 60 seconds


nickcastel50:
Caching on the GitHub API is causing issues saving models within 60 seconds

PR #1167, opened 2024-05-02
Network Caching System + Octokit Version Upgrade + Node 18 Upgrade

nickcastel50:
Satisfies #1166 

This PR adds a full network caching system using etags. This ended up being a bit long-winded, because there is a bug in OctoKit where, when using the If-None-Match header and passing an etag, if we are not logged in OctoKit has a bug when returning the response and will crash. Bug is here: https://github.com/octokit/request.js/issues/659

Because of this, I updated to the latest OctoKit which fixed the issue. The caveat here is I also had to upgrade us to Node 18 to use the latest OctoKit, which means we should thoroughly test this before deployment. The latest OctoKit also does not have a native fetch function anymore, so node-fetch was installed as a dev dependency in order to get the jest tests working again, as the GitHub specific tests were failing.


pablo-mayrgundter:
Heya, just checking that this is paused pending the new etag caching?

pablo-mayrgundter (approved):

PR #1173, opened 2024-05-06
conway 0.1.510-1 -> 0.1.513

pablo-mayrgundter:
Partly addresses https://github.com/bldrs-ai/test-models-private/issues/37

Screen included in private issue

i'm skipping percy on just geom fixes in preference for manually testing all samples, since that covers more than what we're doing in percy currently.


nickcastel50 (approved):
LGTM

PR #1197, opened 2024-05-30
conway-0.1.543

pablo-mayrgundter:
We did QA & Perf testing.

In QA we are seeing all the SP Portfolio models are the same.
There's a new CSG issue in thatopen, but significant other private fixes, so going to release.

We'll do more Performance tests later, currently a little slower than before on big boy model.


nickcastel50 (approved):
LGTM

PR #1198, opened 2024-05-31
open-200: support .stp and .step URLs

pablo-mayrgundter:
Demo:

https://deploy-preview-1198--bldrs-share.netlify.app/share/v/gh/bldrs-ai/test-models/main/step/zoo.dev/a-gear.step

Percy
https://percy.io/8fe2b2f1/share/builds/34513539

Diffs:
- First 2 gear ones are from new test case names, one each for .stp and .step
- Not sure why Open 100 has no prior.  I recently changed the way we get base images, so maybe let's just note and approve these for a bit
- There's jitter in a few below, plus a little change in notes that should have been picked up in that pr


nickcastel50 (approved):
LGTM

PR #1208, opened 2024-06-04
Minor UI fixes: Search + Versions panel

OlegMoshkovich:
This PR:
- Sets the width of the search bar to match the width of other panels.
- Adds an adornment to the TextField component of the search bar to clear the input, ensuring the search string does not overlap with the clear icon.
- Sets the search TextField to multiline to accommodate long GitHub strings.
- Moves the versions panel to the same location as the nav tree panel.

# Updated
[percy](https://percy.io/8fe2b2f1/share/builds/34616270)


pablo-mayrgundter (changes requested):
Code ok, let's see percy for UI change comparison

pablo-mayrgundter:
I don't see any changes to search in Percy.  Shouldn't we see them in the search-100 tests?

pablo-mayrgundter:
Oh whoops, I do see it.  But also a flake.  Can you add waits for those to deflake?

This is the intended change:
image

These are the flakes:
image

OlegMoshkovich:
@pablo-mayrgundter notes tests should not be a part of this PR. 
it should be addressed separately.

Updated percy run - PTAL

nickcastel50 (approved):
LGTM

OlegMoshkovich (commented):
Created a ticket to deal with notes percy flakes.

PR #1211, opened 2024-06-07
Fix social icons on About dialog

OlegMoshkovich:
This PR fixes the alignment of the social icons in the about dialog and removes the Twitter and LinkedIn icons since we are not yet active on those platforms. It leaves only the Discord and GitHub links, as those are our primary channels of communication and engagement.


nickcastel50 (approved):
LGTM

PR #1225, opened 2024-07-02
notes: reduce gh api usage by delay of gh issues fetch until notes are open

pablo-mayrgundter:
Just add a check for isNotesVisible in the NotesControl loader useEffect to delay load until Notes are visible


nickcastel50 (approved):
LGTM

PR #1228, opened 2024-07-05
open: reduce gh api usage by delay of gh orgs fetch until open/save dialog is open

pablo-mayrgundter:
Similar to last PR for open, just condition the useEffect on the component visibility.

Added tests for visible and not.


nickcastel50 (approved):
LGTM

Issue #1238, opened 2024-07-17
Reduce Latency Associated with Element Selection by Keeping Navigation Panel Deactivated


OlegMoshkovich:
When an element is selected, the navigation panel is automatically activated, causing significant latency during element selection.

Proposed Solution:
Keep the navigation panel deactivated by default unless explicitly accessed by the user.

STORY:
As a user, I want to click on the elements with out any UI changes.

# Tasks
* User visits permalink to selected element
  - [ ] The Nav panel remains closed

# Tests
[Cypress]()
[Percy](https://percy.io/8fe2b2f1/share/builds/35432989/)

PR #1239, opened 2024-07-17
Disable nav tree activation on the element select

OlegMoshkovich:
#1238 

[percy](https://percy.io/8fe2b2f1/share/builds/35432989/)

### DIFF
--- a/cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
+++ b/cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
@@ -47,6 +47,7 @@ describe('View 100: Synchronized View and NavTree', () => {
     })
 
     it('Item highlighted in scene - Screen', () => {
+      cy.get('[data-testid="nav-tree-root"]').should('not.exist')
       cy.percySnapshot()
     })
   })
--- a/cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
+++ b/cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
@@ -46,7 +46,7 @@ describe('View 100: Synchronized View and NavTree', () => {
       waitForModelReady('twoLevelSelect')
     })
 
-    it('Item highlighted in tree and scene - Screen', () => {
+    it('Item highlighted in scene - Screen', () => {
       cy.percySnapshot()
     })
   })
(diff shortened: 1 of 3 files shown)


OlegMoshkovich:
@pablo-mayrgundter PTAL

OlegMoshkovich:
..In the process of running Percy

OlegMoshkovich:
@pablo-mayrgundter PTAL

OlegMoshkovich:
This change is testable in Percy; however, it is difficult to test since it involves synchronization of the model interaction and the tree panel response. The current test involves accessing the app through the permalink and SCREEN testing whether all of the elements respond accordingly. The current PR changes the interaction with the tree, keeping it closed when an element is selected; therefore, a Percy screenshot has been adjusted.

**Unit Testing:**
Here, I don't know exactly which component to test because this PR subtracts the logic that sets the panel visibility store value to true, keeping it the same. Maybe that could be tested by setting up the store variables between selected elements and the tree and testing if those store variables remain unchanged when setting the selected elements to a number.

**End-to-End Test:**
Here, to test the interaction, an element needs to be clicked. We need to figure out how to select an element. One way is to navigate to the app with the element selected in the URL and test if the tree remains closed. Yes, that test could be adjusted.

image

OlegMoshkovich:
@pablo-mayrgundter PTAL

cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
```diff
@@ -46,7 +46,7 @@ describe('View 100: Synchronized View and NavTree', () => {
       waitForModelReady('twoLevelSelect')
     })
 
-    it('Item highlighted in tree and scene - Screen', () => {
+    it('Item highlighted in scene - Screen', () => {
       cy.percySnapshot()
```
pablo-mayrgundter, line 50: Please add check that tree is not visible

pablo-mayrgundter (changes requested):

cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
OlegMoshkovich: done.

OlegMoshkovich (commented):
Added.

pablo-mayrgundter (approved):

cypress/e2e/view-100/synchronized-view-and-navtree.cy.js
OlegMoshkovich: done

PR #1248, opened 2024-07-25
Dynamic octokit instantiation based on auth status

nickcastel50:
* Dynamic octokit instantiation based on auth status, we probably don't want to use the proxy for unearthed requests due to rate limiting.


pablo-mayrgundter (changes requested):
Let's figure out a way to test this.  Should need some new hooks in msw?  for new non-auth'd paths

src/net/github/OctokitExport.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+export function initializeOctoKit(authed) {
+    if (authed) {
+        octokit = new Octokit({
+        baseUrl: GITHUB_BASE_URL_AUTHED,
+        userAgent: `bldrs/${PkgJson.version}`,
+      })
+    } else {
+        octokit = new Octokit({
+            baseUrl: 'https://api.github.com',
```
pablo-mayrgundter, line 25: Pretty sure this should be from process.env

nickcastel50: updated

src/net/github/OctokitExport.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Initialize an instance of Octokit depending on auth status.
+ */
+export function initializeOctoKit(authed) {
+    octokit = new Octokit({
+    baseUrl: authed ? GITHUB_BASE_URL_AUTHED : GITHUB_BASE_URL_UNAUTHED,
+    userAgent: `bldrs/${PkgJson.version}`,
+    })
+}
```
pablo-mayrgundter, line 22: Please turn this into two functions, initializeOctoKitAuthed and initializeOctoKitUnauthed

pablo-mayrgundter (approved):

src/net/github/OctokitExport.js
nickcastel50: PTAL

PR #1256, opened 2024-08-22
Optimized Model Loading

nickcastel50:
This PR reworks model load + OPFS storage structure, with the goal of reducing an API call to GitHub while unauthenticated.


tools/esbuild/vars.prod.js
```diff
@@ -12,6 +12,9 @@ export default {
   GITHUB_API_TOKEN: null,
   GITHUB_BASE_URL: 'https://git.bldrs.dev/p/gh',
   GITHUB_BASE_URL_UNAUTHENTICATED: 'https://api.github.com',
+  // RAW_GIT_PROXY_URL_NEW: 'http://localhost:8083/model',
+  // RAW_GIT_PROXY_URL: 'http:localhost:8083/r',
+  RAW_GIT_PROXY_URL_NEW: 'https://rawgit.bldrs.dev/model',
```
pablo-mayrgundter, line 17: Please document both this and the old one for how/when each is used.

pablo-mayrgundter (approved):

PR #1271, opened 2024-11-30
UI Refresh: Apps, LeftDrawer, themed Typography, reorg; Hot-reload esbuild and more

pablo-mayrgundter:
[Percy](https://percy.io/8fe2b2f1/web/share/builds/37682929)

# Main updates:
- AppsDrawer
- LeftDrawer: NavTree & Versions now on left like Notes & Properties on right
- Move most Component groups to Containers dir
- Serve Roboto font (WIP)
- More uniform use of data-testid
- Typography from theme
- Icons & Typography sized together

# Tools
- Hot reloading in dev
- React StrictMode in dev

## Broken in Percy:
- Search 100: Permalink Returning user visits homepage, Open Search > Enters "together" Search box with query visible, "Together" items highlighted in tree and scene 
- Notes 100: Access shared note Returning user accessing permalink with notes panel open Notes open - Screen
- view 100: Cutplanes View model Click CutPlaneControl Select all cut-planes Clear all cut-planes No cutplanes visible

PR #1274, opened 2024-12-05
cutplanes: fix regression that broke Clear All in #1271

pablo-mayrgundter:
We caught this in Percy.  Fix just needed to line up some positional params

This looks correct to me:
https://percy.io/8fe2b2f1/web/share/builds/37686044/changed/2050599839


nickcastel50 (approved):
LGTM

PR #1297, opened 2024-12-11
versions: fix mobile VersionPanel display. percy

pablo-mayrgundter:
also try rewriting percy call to explicitly set viewport width to trigger mobile width correctly.

The percy fix isn't working.  I have a support ticket in with BrowserStack.


nickcastel50 (approved):
LGTM

PR #1302, opened 2024-12-11
Fix: 1300-sharing-removes-cut-plane

### LINKED ISSUE #1300: Sharing removes cut-plane
1) Setup cutplane
2) Click Share control
3) Cutplane is removed

Expect cutplane to stay active, and also be part of the Share dialog, and that it's active when loaded as permalink

pablo-mayrgundter:
Fixes #1300.

Reverted back to pre UI-refresh and fixed up a couple diffs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.0.1196",
+  "version": "1.0.1198",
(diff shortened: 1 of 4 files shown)


nickcastel50 (approved):
LGTM

Issue #1280, opened 2024-12-06
SnackBar fixes


OlegMoshkovich:
- There's "[Object object]" text rendered sometimes.. maybe that's the progress object not being unpacked
- The recent theme changes have clobberd colors, so can't read text anymore

<img width="460" alt="image" src="https://github.com/user-attachments/assets/d872d676-8a9b-4e41-9884-a1e1b21c42e0">

PR #1310, opened 2024-12-13
1281-properties-ux

### LINKED ISSUE #1281: Properties UX
Each PropertySet should have a visual indicator that it is expandable.  The margins and such should be more compact.

image

pablo-mayrgundter:
Fixes #1281.  Partly addresses #1280.

- Props: use carat open/close icon.
- NavTree: replace custom open/close icon with mui carat.
- Snack: fix colors.
- Removed some old config from theme/Components, and converted all theme.palette to just use the scoped palette


nickcastel50 (approved):
LGTM

PR #1313, opened 2024-12-16
Fix sample models chip padding

pablo-mayrgundter:
[Percy TBD](https://percy.io/8fe2b2f1/web/share/builds/37842324)
  - Backgrounds aren't right theme color, will fix in subsequent PR
  - Not sure about the bottom button behavior. not seeing local testing and could be related to fixed viewport size percy is using, as in the mobile screenshot issue


nickcastel50 (approved):
LGTM

PR #1340, opened 2025-03-07
Downgrade conway to stable version

nickcastel50:
- noticed a regression in 0.16.849, downgrading to last version


ConorStokes (approved):
LGTM

PR #1347, opened 2025-03-13
Disable http caching on /contents endpoint

nickcastel50:
- Due to an issue with the ETAG system with this endpoint, we have to disable http caching here. 
- More info here: [#67](https://github.com/bldrs-ai/test-models-private/issues/67#issuecomment-2719426602)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.0.1288",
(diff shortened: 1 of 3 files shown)


pablo-mayrgundter (approved):
Please add the comment, but also a big TODO that says we'll be directly downloading small content responses and also avoiding requests when we have a cache file that matches the etag hash.

PR #1348, opened 2025-03-14
Add OPFS does file exist check + bubble up cache hit on http cache

nickcastel50:
- Add OPFS does file exist check + bubble up cache hit on http cache
- Satisfies new flow, does not handle using base 64 content for small files (yet)

```
first time
   /contents
   For small files, maybe just use the content
   download OTURL, story <sha>.ext
      Browser context keeps cache API to store etag data
Next time
   REQ /contents w/etag
   304 Not Modified, no body, but we have a cached RSP
      Use cached SHA to see if file exists in OPFS (now we're safe with clearOPFSCache); if so, use it no DL
      If not GET /contents w/o ETAG
         For small files, maybe just use the content
```

### DIFF
--- a/src/OPFS/OPFS.worker.js
+++ b/src/OPFS/OPFS.worker.js
@@ -165,6 +165,7 @@ async function fetchRGHUC(modelUrl) {
     return modelResponse
   } catch (error) {
     console.error('Error:', error)
+    return null
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/OPFS/OPFS.worker.js
```diff
@@ -165,6 +165,7 @@ async function fetchRGHUC(modelUrl) {
     return modelResponse
   } catch (error) {
     console.error('Error:', error)
```
pablo-mayrgundter, line 167: I think we should add Sentry tracking for cases like this, either by letting the error bubble or explicitly logging it with Sentry.captureException(err)

src/loader/urls.js
```diff
@@ -110,10 +110,11 @@ export function parseCoords(url) {
  * @param {string} urlStr
  * @param {string} accessToken
  * @param {boolean} isOpfsAvailable
+ * @param {boolean} useCache
  * @return {Array<string>} A tuple of urlStr (changed to our proxy if
```
pablo-mayrgundter, line 114: Please update

src/loader/urls.js
```diff
@@ -135,32 +136,34 @@ export async function dereferenceAndProxyDownloadUrl(urlStr, accessToken, isOpfs
...
+      return [urlStr, '', false]
   }
 }
 
 
 /**
  * @param {string} urlStr
  * @param {string} accessToken
+ * @param {boolean} useCache
  * @return {Array<string>} Pair of [downloadUrl, sha]
```
pablo-mayrgundter, line 154: Please update

src/loader/urls.js
```diff
@@ -135,32 +136,34 @@ export async function dereferenceAndProxyDownloadUrl(urlStr, accessToken, isOpfs
...
  */
-async function getGitHubPathContents(urlStr, accessToken) {
+async function getGitHubPathContents(urlStr, accessToken, useCache) {
   const repo = parseGitHubRepositoryUrl(urlStr)
-  const [downloadUrl, sha] = await getPathContents(
+  const [downloadUrl, sha, cacheHit] = await getPathContents(
```
pablo-mayrgundter, line 158: isCacheHit

src/net/github/Http.js
pablo-mayrgundter, line 1: Add test

pablo-mayrgundter (changes requested):

src/loader/urls.js
nickcastel50: done

src/loader/urls.js
nickcastel50: done

src/loader/urls.js
nickcastel50: done

pablo-mayrgundter (approved):

PR #1377, opened 2025-05-20
Fix progress display for file load

pablo-mayrgundter:
## Summary
- handle object progress events so the snackbar shows bytes instead of `[object Object]`

## Testing
- `yarn test` *(fails: "Internal Error... This package doesn't seem to be present in your lockfile")*


pablo-mayrgundter:
Finishes fix for https://github.com/bldrs-ai/Share/issues/1280

nickcastel50 (approved):
LGTM

PR #1393, opened 2025-07-28
Analytics tweaks

pablo-mayrgundter:
- analytics, CadView, Loader: rename gtag to gtagEvent, combine model load and stats into single event, with stats_ prefixed keys.
- utils/objects: add object merge with prefix
- OPFS.worker: fix global.navigator setup style
- Also cleans up a bunch of pending lint


nickcastel50 (approved):
LGTM

PR #1396, opened 2025-07-29
cypress: fix-up parallel script and use for GHA intead of serial.

pablo-mayrgundter:
Tested locally a bunch of times to get the tuning value of 6 parallel.  We'll see how it goes with GHA


nickcastel50 (approved):
LGTM

PR #1489, opened 2026-04-25
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Critical` | **File**: `netlify/functions/create-portal-session.js:L20`

The Netlify function trusts a client-provided `stripeCustomerId` from `event.body` and creates a billing portal session for that customer without any authentication or authorization checks. An attacker can submit arbitrary customer IDs and obtain valid portal links for other users' Stripe accounts.

## Solution

Require authenticated requests (e.g., verify JWT/session), derive the Stripe customer ID server-side from the authenticated user record, and reject any client-supplied customer ID mismatch. Also validate HTTP method and add audit logging for portal session creation.

## Changes

- `netlify/functions/create-portal-session.js` (modified)
- `public/mod.js` (modified)

### DIFF
--- a/netlify/functions/create-portal-session.js
+++ b/netlify/functions/create-portal-session.js
@@ -13,15 +13,41 @@ Sentry.AWSLambda.init({
   to a page on Stripe that lets them configure their sub.
  */
 exports.handler = Sentry.AWSLambda.wrapHandler(async (event) => {
+  if (event.httpMethod && event.httpMethod !== 'POST') {
+    return {statusCode: 405, body: JSON.stringify({error: 'Method Not Allowed'})}
+  }
+
+  const user = event.clientContext && event.clientContext.user
+  if (!user) {
+    return {statusCode: 401, body: JSON.stringify({error: 'Unauthorized'})}
+  }
+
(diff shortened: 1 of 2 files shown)


pablo-mayrgundter:
Thanks @tomaioo — you flagged a real bug. The Netlify function did trust a client-supplied `stripeCustomerId`, and any caller with a `cus_…` ID could mint a portal URL for that customer.

We've shipped a fix in #1492 with a slightly different approach: instead of validating the client-supplied ID, we drop it from the request entirely and derive the customer ID server-side from the authenticated Auth0 user (Bearer token → `/userinfo` → Management API → `app_metadata.stripeCustomerId`). This mirrors the auth pattern we already use in `unlink-identity.js` and removes the trust on the client value altogether.

Going to close this PR in favor of #1492. Two notes on the diff here that didn't make the cut:
- `public/mod.js` — that's our Matrix-widget host demo, unrelated to billing; we'd want any changes there scoped to a separate PR.
- The body validation pattern still passed through a client-supplied customer ID after checks; we preferred to remove the input entirely.

Genuinely appreciate the report — please keep them coming.

PR #1505, opened 2026-05-07
sharing: provider scaffolding + Drive adapter (PR1 of multi-user sharing)

pablo-mayrgundter:
Implements PR1 of [`design/new/multi-user-sharing.md`](https://github.com/bldrs-ai/Share/blob/main/design/new/multi-user-sharing.md): the provider-side scaffolding and Drive adapter for in-app sharing. Pure-additive, no UI, behind a `sharing` feature flag (default off).

## What's in PR1

The design splits the sharing work into six sequenced PRs. PR1 is everything you can ship without touching UI:

- **`ConnectionProvider` extension** — five optional methods: `listGrants`, `shareWith`, `revokeGrant`, `getVisibility`, `setVisibility`. Optional so adapters can declare any subset; the GitHub side comes in PR3 once `identity-decoupling.md` lands.
- **Sharing types** — `ResourceRef` (discriminated by provider/kind), `Visibility` (`private | org | public`), `GrantRequest`, `Grant`. Provider-neutral on top, provider-shaped at the principal level.
- **Typed errors** — `GrantFailedError`, `InsufficientPermissionError` join the existing `NeedsReconnectError`. The same vocabulary the PR2 dialog will route to typed `AlertDialog` variants.
- **Drive adapter** (`GoogleDriveSharing.ts`) — pure transport over `drive.permissions.*`, factored out of the provider object the way `GoogleDriveBrowser.ts` is. The provider methods just acquire a token and delegate. `drive.file` scope is sufficient — no auth changes.
- **`sharing` feature flag** — present but off; PR2 gates the Share dialog on it.

## Notable design decisions

- **`getVisibility` returns `null` honestly** when the connection has no email metadata and a domain permission exists — we genuinely can't tell `org` from `private` without a workspace anchor. UI shouldn't render a chip in that case.
- **`setVisibility('org')` without an inferable domain** throws `GrantFailedError(cause: 'domain_unknown')` rather than guessing. PR2 will prompt for a domain.
- **`setVisibility` is idempotent** by reading current state first — calling \"public\" on an already-public resource is a no-op, not a duplicate-grant error.
- **`'public'` wins over `'org'`** in the visibility derivation: a link-shared resource is effectively public regardless of additional domain grants.
- **401 → `NeedsReconnectError`** maps the case where our token cache says \"valid\" but Drive says \"revoked.\" 403 → `InsufficientPermissionError`. Other 4xx/5xx → `GrantFailedError(cause: 'http_<status>')`.
- **`'private'` preserves user/group grants** — \"private\" in Drive means \"no link, no domain\"; named collaborators are not sharing in the visibility sense.

## Tests

25 new unit tests in `GoogleDriveSharing.test.ts`, stubbing `fetch` directly (matching the pattern in `GoogleDriveProvider.test.ts`):

- Permission-list mapping (user/group/domain/anyone shapes, all roles)
- URL targeting (`supportsAllDrives=true`, the right path for files vs folders)
- Cross-provider rejection (`drive*` helpers refuse `github-*` resources with `cause: 'wrong_provider'`)
- Principal-shape validation (user/group rejected without `principalId`)
- All four visibility derivations including the null and \"public-wins\" cases
- Idempotent `setVisibility('public')` and `setVisibility('private')` selectivity
- Error mapping for 401/403/500

All 1035 pre-existing tests still pass.

## Out of scope (explicitly)

- Share dialog UI — PR2.
- GitHub adapter — PR3, blocked on `identity-decoupling.md`.
- Folder-scoped routes — PR4.
- GitHub token-health parity — PR5.
- Comments + versions stretch — multi-quarter follow-on.

### DIFF
--- a/src/FeatureFlags.js
+++ b/src/FeatureFlags.js
@@ -4,4 +4,8 @@ export const flags = [
   },
   {name: 'googleOAuth2', isActive: true},
   {name: 'googleDrive', isActive: true},
+  // Multi-user sharing UI (Share dialog, visibility chip). Provider scaffolding
+  // ships unconditionally; this flag gates the consumer surface in PR2+.
+  // See design/new/multi-user-sharing.md.
+  {name: 'sharing', isActive: false},
 ]
--- a/src/connections/errors.ts
+++ b/src/connections/errors.ts
@@ -30,3 +30,60 @@ export class NeedsReconnectError extends Error {
     Object.setPrototypeOf(this, NeedsReconnectError.prototype)
   }
 }
+
+
+/**
+ * Thrown when a sharing-capability call is refused by the provider for
+ * authorization reasons — the connection is healthy, but the acting user
+ * cannot perform the operation on this resource (e.g. listing collaborators
+ * on a repo they don't admin, or sharing a Drive file they don't own).
+ *
+ * Distinct from `NeedsReconnectError`: there's nothing to reconnect, the
+ * user simply lacks the privilege. UI should surface this as "you don't
+ * have permission" rather than as a sign-in prompt.
+ */
+export class InsufficientPermissionError extends Error {
+  readonly connection: Connection
+  readonly cause?: string
+
+  /**
+   * @param connection The connection that was acting on the resource
+   * @param cause Short machine-readable tag, e.g. 'forbidden'
+   * @param message Human-readable message for logs/dialogs
+   */
+  constructor(connection: Connection, cause?: string, message?: string) {
+    super(message ?? `Insufficient permission for ${connection.label}`)
+    this.name = 'InsufficientPermissionError'
+    this.connection = connection
+    this.cause = cause
+    Object.setPrototypeOf(this, InsufficientPermissionError.prototype)
+  }
+}
+
+
+/**
+ * Thrown for any non-auth, non-permission failure during a sharing-capability
+ * call: 4xx that isn't 401/403, 5xx, network failure, malformed response,
+ * or a request the adapter refused outright (e.g. a `ResourceRef` shape the
+ * adapter doesn't own).
+ *
+ * The `cause` field carries a short machine-readable tag so callers and
+ * dialogs can branch without parsing `message`.
+ */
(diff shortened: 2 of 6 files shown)


pablo-mayrgundter:
Pushed `6f56504` addressing the four review findings:

1. **Security — free-email guard.** `workspaceDomain` now returns `null` for personal-Google and other free-email providers (gmail.com, outlook.com, icloud.com, …). Prevents `setVisibility('org')` from attempting `domain: 'gmail.com'` (which would share with every Gmail user worldwide). Throws `GrantFailedError(cause: 'domain_unknown')` instead.
2. **Type safety — Shared Drive roles.** New `normalizeRole` helper downcasts `'organizer'` and `'fileOrganizer'` to `'writer'` (both grant write access; nearest cross-provider equivalent), so `Grant.role` no longer carries values its type union excludes.
3. **Partial-failure aggregation.** `setVisibility('private')` switched from sequential awaits to `Promise.allSettled`. If all revokes fail with the same auth-class typed error (`NeedsReconnect` / `InsufficientPermission`), that error propagates directly so UI routes cleanly. Otherwise wraps as `GrantFailedError(cause: 'partial_revoke')` with the surviving ids in the message.
4. **Case-insensitive domain matching.** `workspaceDomain` lowercases the email domain; `permissionToGrant` lowercases Drive-returned domains. Robust to user-typed casing.

Tests grow 25 → 35:
- Shared Drive role downcast (organizer/fileOrganizer → writer)
- Lowercase domain principalId mapping
- Empty-`permissions`-field response
- `notify: false` explicit forward (`sendNotificationEmail=false`, message dropped)
- `getVisibility` mixed-case-email match → `'org'`
- `getVisibility` for gmail.com → `null` (free-email branch)
- `setVisibility('org')` for gmail.com → `domain_unknown`
- `setVisibility('org')` already-org idempotency
- Partial revoke → `GrantFailedError(partial_revoke)` with survivor ids
- All-same-class revoke failures → propagates `NeedsReconnectError`

1045 tests pass; lint clean.

PR #1588, opened 2026-07-14
Lazy properties step 1: load spatial tree in 'names' mode + release entity cache

pablo-mayrgundter:
## What

First step of the lazy/incremental property-load track. The eager full-property visit on model load is vestigial from the web-ifc days — conway's parse-time index already backs on-demand access, so the load path only needs the node names.

- Bumps `@bldrs-ai/conway` to **1.373.1180**, which includes [conway#373](https://github.com/bldrs-ai/conway/pull/373): `getSpatialStructure(modelID, 'names')` returns per-node `Name`/`LongName`/`GlobalId` value handles only (no recursive property inlining), plus `IfcAPI.ReleaseEntityCache(modelID)` to drop the entity-descriptor cache after load.
- `CadView.jsx`: `getSpatialStructure(0, true)` → `getSpatialStructure(0, 'names')` on both load branches, then calls `ReleaseEntityCache(0)` once the element-types map is built.
- `conwayDirectIfcLoader.js`: the `getSpatialStructure` argument-forwarding closure now passes `'names'` through instead of boolean-coercing it (the old `!!args[1]` would have silently degraded `'names'` → `true`, keeping the eager visit).

## Why

On large models the eager visit materializes every entity record at load just to inline properties the UI may never open. With 'names' mode the tree carries only what NavTree/search need; everything else stays behind conway's on-demand API (Properties panel, ifclib deref per opened branch — unchanged). Combined with the cache release this removes the largest post-load JS heap resident and cuts load time.

Consumer audit (all load-path consumers of the spatial tree verified against the names shape): `setupLookupAndParentLinks`, `SearchIndex`, `reifyName`, `groupElementsByTypes`, NavTree, `IfcIsolator`, cache-hit closure capture, GLB writers, `getByFloor` (deliberately untouched — still requests full props on demand since it needs `Elevation`).

## Testing

- Full Share suite green: 208 suites, 1830 passed.
- conway#373 carries its own parity tests (names ↔ full-mode value agreement, ReleaseEntityCache round-trip) and shipped with byte-identical geometry regression green.
- Needs a manual smoke test on a large model (NavTree names, search, Properties panel on-demand, floor isolation).

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@babel/plugin-syntax-import-assertions": "7.18.6",
     "@babel/preset-env": "7.18.10",
     "@babel/preset-react": "7.18.6",
-    "@bldrs-ai/conway": "1.372.1179",
+    "@bldrs-ai/conway": "1.373.1180",
     "@bldrs-ai/ifclib": "5.3.3",
     "@emotion/react": "11.10.0",
     "@emotion/styled": "11.10.0",
--- a/src/Components/Properties/Properties.jsx
+++ b/src/Components/Properties/Properties.jsx
@@ -54,10 +54,10 @@ export default function Properties() {
       if (model && element) {
         // Resolve `element` to the full IFC entity before rendering.
         // `selectedElement` is a spatial-tree node — for cache-miss
-        // IFC it carries full IFC properties (wit-three's
-        // `getSpatialStructure(0, true)` inlines them); for cache-hit
-        // GLB it's the slim whitelist {expressID, type, Name,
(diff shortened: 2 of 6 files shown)


pablo-mayrgundter:
📐 **Design context:** this work is now captured in the lazy-properties-memory design doc — **#1590** (`design/new/lazy-properties-memory.md`), which lays out the property-memory sinks, the merge-order rationale (this landed after #1589, rebased on top), and what's next. Engine-side companion: **bldrs-ai/conway#375** (`design/new/memory-residency.md`).

---

PR #1589, opened 2026-07-14
Stream the GLB element-properties capture (fixed-memory writer)

pablo-mayrgundter:
## What

Step 2 of the lazy/incremental properties track (follows #1588): make the GLB cache writer fixed-memory instead of manifesting the whole property tree before serializing.

The fast capture path in `bldrsElementProperties.js` materialized **every** parsed entity into one big `itemProperties` object (millions of records on 100MB-class IFCs — multi-GB transient JS heap) and only then filtered down to the reachable-from-IfcRoot closure (~50k entities on Snowdon). This replaces it with a streaming sweep that gzips the wire JSON incrementally:

- **Sweep 1** scans entities via sync `proxy.getLine`, serializes IfcRoot seeds (GlobalId-bearing) straight into a pako streaming `Deflate`, builds the product→psetIds index inline, and queues non-geometric refs. Records not kept are dropped on the spot.
- **Sweep 2** drains the ref closure by id — fetch, serialize, release.
- conway entity memoization is disabled during the sweep (restored after, throw-safe), and the entity/descriptor cache is released every 200k materialisations plus once at the end (`proxy.releaseEntityCache`, conway ≥1.373 — optional-chained no-op on older versions; this branch stays on 1.372 and works either way).
- Peak retained memory: **O(reachable ids + pset index + one record + compressed output)** instead of O(all parsed records).

## Wire compatibility

The on-disk format is unchanged — same gzipped `{itemProperties, propertySets}` JSON the existing reader already decodes (key order differs from the legacy one-shot stringify; decode-identical). No reader changes, no cache-schema bump, old GLBs keep working.

Plumbing: the capture now returns a discriminated union — `{compressedBytes}` from the streaming path, the legacy object from the slow fallback (tests/non-Conway loaders) — and `injectGlbExtensions` accepts a `precompressed` payload it embeds verbatim with `compressed: true`, so neither the main thread nor the GlbWriter worker re-materializes/re-compresses the payload.

## Testing

- Adapter-path suite migrated to decode the streamed bytes and assert the same content (seeds, pset index, geometric pruning, error tolerance).
- New pins: memoization off during sweep + restored after (including on throw), cache release invoked, `precompressed` inject round-trip, and a full streaming capture → inject → reader decode end-to-end.
- Full Share suite green (208 suites, 1830+ passed) via the pre-push hook.

Related: conway#372 (SoA descriptors), conway#373 ('names' spatial tree), #1588 (load-path flip).

---

### DIFF
--- a/src/loader/GlbWriterService.js
+++ b/src/loader/GlbWriterService.js
@@ -86,7 +86,11 @@ function getWorker() {
  *   the underlying ArrayBuffer
  * @param {string|null} args.mode container mode tag — draco/meshopt/null
  * @param {Array<object>} args.extensions extension payloads to inject;
- *   each entry is `{name, data, compress?}` matching `injectGlbExtensions`
+ *   each entry is `{name, data, compress?, precompressed?}` matching
+ *   `injectGlbExtensions`. A `precompressed` Uint8Array rides through
+ *   the structured clone intact (deliberately NOT in the transfer
+ *   list — the caller's inline fallback after a worker failure must
+ *   still see an attached buffer)
  * @param {object} [args.sceneExtras] optional small string-keyed
  *   metadata merged into `scenes[0].extras` in the same inject pass
  *   (see `injectGlbExtensions`)
--- a/src/loader/bldrsElementProperties.js
+++ b/src/loader/bldrsElementProperties.js
@@ -186,72 +186,122 @@ function collectSpatialTreeIds(root) {
 
 /**
  * Capture the element-properties closure for a model from an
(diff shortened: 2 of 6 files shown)


pablo-mayrgundter:
📐 **Design context:** captured in the lazy-properties-memory design doc — **#1590** (`design/new/lazy-properties-memory.md`). This PR (streaming fixed-memory GLB property writer) is sink #3 in the map there, with the measured PSB context and the "memory-safety, not speedup" framing. Engine-side companion: **bldrs-ai/conway#375** (`design/new/memory-residency.md`).

---

PR #1618, opened 2026-07-22
Bump @bldrs-ai/conway 1.434.1300 → 1.432.1305: STEP surface-fidelity release

pablo-mayrgundter:
Ships the full STEP surface-fidelity pass for bldrs-ai/test-models#47 (bldrs-ai/conway#432 → conway-geom#150–#154). Note the version: releases are `1.&lt;PR&gt;.&lt;run&gt;`, so `1.432.1305` was published **after** and supersedes `1.434.1300` (the #432 merge includes #434&#39;s changes).

What this picks up, all on the Jetenginestep model and friends:

- **Revolution surfaces** (conway-geom#150/#151): adaptive ring count + CDT-trimmed boundary loops — round shafts and spinner.
- **Linear-extrusion surfaces** (conway#432 + conway-geom#152): extractor builds native profiles from AP214&#39;s bare swept curves; exact developable unwrap — correct compressor drum skins and blades (1,370 faces that previously fell to the flat fallback).
- **Cylindrical + conical surfaces** (conway-geom#153): trimmed (θ, z) unwrap with CDT hole nesting replaces earcut chord projections — smooth blade/join/shaft rings between compressor sections, and the jet drops 4.9M → 3.5M triangles (driver_board 3.7M → 0.4M), so GLBs get lighter across the board.
- **Toroidal surfaces** (conway-geom#154): interior Steiner seeding before CDT — the fuel-manifold tube renders as a round pipe instead of a flat creased ribbon.

Conway CI on #432: no regression failures, two previously-failing ap242 models now pass, visual diff clean across all 24 rendered models.

Also converts two `{@link}` references to method names into plain code spans in `incrementalBatchedBuilder.js` — `jsdoc/no-undefined-types` was failing `eslint --max-warnings 0` on main, blocking any commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01J3A79iKimn7UheESvnumj7

---
_Generated by [Claude Code](https://claude.ai/code/session_01J3A79iKimn7UheESvnumj7)_

PR #1626, opened 2026-07-24
Auto-color colorless STEP/CAD assemblies by product (Onshape-style)

pablo-mayrgundter:
## What

Colorless multi-part STEP/CAD models now render with a distinct color per **part** instead of a uniform grey blob — the way Onshape (and most MCAD viewers) auto-assign an appearance per component on import.

Motivating case: the GrabCAD **Jetenginestep** model. It's an AP203 `CONFIG_CONTROL_DESIGN` export containing pure geometry + assembly structure and **zero** presentation entities (grepped: 0 × `COLOUR_RGB`, 0 × `STYLED_ITEM`, no materials). AP203e1 can't even encode appearance, so the colors seen in Onshape are Onshape's own default per-component scheme, not data in the file. Conway correctly hands every placement the same fallback grey; this gives the viewer the same fallback Onshape has.

## How

- **`productPalette.js`** (new, pure): a curated 16-color palette + `hashKey` (FNV-1a) → `productPaletteRgb(key)`, plus `isDefaultColor` and `applyProductPalette(batches)`.
- **Keyed by geometry (part), not occurrence.** A STEP assembly instances one part definition many times; each instance is a distinct NAUO occurrence with its own product express id but shares the ONE geometry buffer the part was defined with. Probing the jet's FlatMesh stream: **147 occurrences → 9 distinct `geometryExpressID`s** (every turbine blade is geometry `271914`). So the palette keys by `instanceGeometryIds` — all instances of a part get one color, a different part (the shaft, the casing) its own — falling back to the occurrence id only if a batch lacks the geometry table. (`hashKey` also takes a string, so a future by-name key is a one-line change.)
- **Detection is whole-model and conservative**: fires *only* when every instance across every batch is the fallback grey. The instant any real color is present — an IFC material, a colored STEP part, even one — the file has presentation intent, so we honor it and skip the palette (colorless siblings in a mixed model stay grey, as today).
- **Wired at `assembleBatchedModel`**, the single chokepoint both batched paths (incremental streaming + `?feature=batchedMesh`) funnel through, *before* the pick-table stamp so the `instanceColors` restore table `batchedHighlight` reads already carries the palette color — selection/hover restore, isolate, and batched→merged export all follow. Original per-instance alpha preserved.
- **Feature flag `autoColorParts`, default-on** (only changes models that had zero color).

## Tests

- `productPalette.test.js` — hash determinism / range / spread, palette membership, no entry is grey, default-grey epsilon; and `applyProductPalette`: **colors by geometry so instances of one part share a color**, parent-id fallback when no geometry table, buffer + restore-table repaint, cross-batch keying, alpha preservation, no-op on any real color (mixed model), no-op single-part, bail on missing tables.
- `buildBatchedConwayModel.test.js` — integration through a **real** `THREE.BatchedMesh` + `flatMeshToBatchedModel` + `assembleBatchedModel`, colors read back via `getColorAt`: **three blade occurrences sharing geometry get one color, the shaft another**; repaint on/off; honors a real color in a mixed model.
- Full `viewer/ifc` suite (187) green; precommit gate passed.

## Known nuance

Colors are a hash into 16 slots, so two unrelated parts can occasionally collide on the same color (birthday-paradox; ~2 collisions across the jet's 9 parts). Instanced parts are always internally consistent — this only ever affects whether two *different* parts happen to match. A collision-free variant (dense-index the distinct parts, or a larger palette) is a small follow-up if desired.

## Scope

Covers the batched render paths (the jet's actual path). The rare merged-geometry fallback (`buildConwayIfcModel`, used only if batched assembly throws) is not palettized. Independent of the conway release bump #1618; based on `main`.

https://claude.ai/code/session_01J3A79iKimn7UheESvnumj7

### DIFF
--- a/src/FeatureFlags.js
+++ b/src/FeatureFlags.js
@@ -105,6 +105,16 @@ export const flags = [
   // follow-ups. Flip on via `?feature=batchedMesh`.
   // Design: design/new/viewer-replacement.md §3b.iv.
   {name: 'batchedMesh', isActive: false},
+  // Synthetic per-product coloring for STEP/CAD models that carry no
+  // presentation data. When a batched model comes back entirely
+  // default-grey (no COLOUR_RGB / STYLED_ITEM, e.g. the Jetenginestep
+  // AP203 export), each product is repainted from a curated palette by a
+  // deterministic hash of its express id — Onshape-style, so a multi-part
+  // assembly is legible instead of a grey blob. Strictly no-op the moment
+  // any real color is present, so IFC and colored STEP are untouched.
+  // Default-on; it only changes models that had zero color to begin with.
+  // See src/viewer/ifc/productPalette.js.
+  {name: 'autoColorParts', isActive: true},
 ]
 
 
--- a/src/viewer/ifc/buildBatchedConwayModel.js
+++ b/src/viewer/ifc/buildBatchedConwayModel.js
@@ -1,10 +1,12 @@
 import {Group} from 'three'
+import {isFeatureEnabled} from '../../FeatureFlags'
 import {attachBatchedSubsets} from './batchedSubset'
 import {flatMeshToBatchedModel} from './flatMeshToBatchedModel'
 import {
   attachConwayDirectModelMethods,
   makeConwayDirectIfcManager,
 } from './conwayDirectIfcLoader'
+import {applyProductPalette} from './productPalette'
 import {occurrencePathKey} from '../../utils/occurrencePaths'
 
 
@@ -111,6 +113,16 @@ export function assembleBatchedModel(batches, ifcAPI, modelID, opts = {}) {
     // Degenerate (no renderable geometry). Throw so ShareIfcLoader falls
     // back to the merged path rather than adding an empty model.
     throw new Error('buildBatchedConwayModel: no renderable geometry')
+  }
+
+  // Colorless-model fallback: a STEP/CAD file with no presentation data
+  // comes back entirely default-grey. Repaint each product from a palette
+  // (Onshape-style) so a multi-part assembly is legible. Strictly no-op the
+  // moment any real color is present, so IFC and colored STEP are untouched.
+  // Runs before the pick-table stamp below so the `instanceColors` restore
+  // table `batchedHighlight` reads already carries the palette color.
+  if (isFeatureEnabled('autoColorParts')) {
+    applyProductPalette(batches)
   }
 
   // Each batch carries its own pick tables; CadView reads them off the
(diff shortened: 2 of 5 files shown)


pablo-mayrgundter:
## Review — LGTM, no changes required

Reviewed the full diff (5 files: flag, wiring, `productPalette.js`, 2 test files). CI green (build + both Playwright + Netlify); I also re-derived the correctness-sensitive bits against the surrounding code.

### What it does
Colorless multi-part STEP/CAD (no `COLOUR_RGB`/`STYLED_ITEM` — the jet is an AP203 export with zero presentation data) gets a synthetic per-part color instead of a grey blob, keyed by **geometry** id so all instances of a part share a color. Fires only when the whole model is default-grey; any authored color skips it entirely.

### Correctness (verified, not just read)
- **Slot replacement is alias-safe.** `flatMeshToBatchedModel` does `instanceColors[batchId] = placement.color`, and the `?? DEFAULT_COLOR` fallback can hand the *same* constant object to many slots. `applyProductPalette` **replaces** each slot (`instanceColors[i] = {…}`) rather than mutating in place, so a shared fallback object can't be corrupted. ✅
- **Restore-table + buffer stay in sync.** Both `setColorAt` (GPU) and `instanceColors[i]` (the table `batchedHighlight.paint()` restores from) are updated, so selection/hover un-highlight to the palette color, and `batchedToMergedMesh` carries it. Runs before the pick-table stamp, so the assignment `mesh.instanceColors = batch.instanceColors` sees the repainted array. ✅
- **Geometry-keying is the right fix** for the blade-scatter bug: 147 occurrences → 9 geometries on the jet; keyed by occurrence they rainbowed, by geometry they group per part. Integration test pins it through a real `BatchedMesh`.
- **Gate + fallback consistent.** `keyOf` uses the same geometry-or-parent choice in both the classify and repaint passes; `parts.size < 2` correctly no-ops single-part models.

### Style / conventions
Matches the batched-path idioms (module-scratch `Vector4` like `batchedHighlight`, per-batch tables, load-bearing JSDoc). Default-on flag is justified (no-op unless the model had zero color). Perf is linear and early-bails on the first authored color.

### Non-blocking nuances (documented, no action)
1. **Hash collisions**: 16-slot hash → two *unrelated* parts can share a color (~2 across the jet's 9). Instanced parts are always internally consistent; only affects whether two different parts happen to match. Collision-free variant (dense-index distinct parts, or larger palette) is a small follow-up if wanted.
2. **Single product, multiple solids**: a lone colorless part with N distinct solids would get N colors (keyed by geometry). Rare, non-harmful, and arguably still legible; noting for the record.
3. **No URL opt-out**: default-on flags can't be disabled via `?feature=` (the mechanism only adds). Flip in source to compare against honest grey. Fine as-is.

Ready to merge.

---

Issue #1637, opened 2026-07-24
Tune camera near/far depth budget for large models (currently masked by logarithmic depth)


pablo-mayrgundter:
Follow-up to the #1632 retrospective — the **near-plane configuration** finding. Preventive: no confirmed user-visible artifact today, but the budget is measurably poor and only papered over.

## Measured (ISSUE_129, ~74 m radius, fit distance ~194 m)

`orbit-control.js:81-82` sets `near = max(minDistance·0.5, 0.1)` and `far = (maxDistance + radius)·1.5` → **near ≈ 0.97 m, far ≈ 3,022 m** (ratio 3.1×10³). During orbit the model surfaces span 120–268 m from the camera — in fixed-point terms that band uses **~75k of 16.7M depth codes (~0.45%)**. A near/far bracket tight to the fitted sphere would be ~175× better.

Today `logarithmicDepthBuffer: true` (renderer.js) masks this — and #1630's `noLogDepth` A/B confirmed log-depth on/off makes no visible difference to the jitter class. But log-depth has real costs (per-fragment `gl_FragDepth` writes disable early-Z on some GPUs) and hides rather than fixes the budget.

## Options

1. **Dynamic near from the fitted sphere** — e.g. `near = max((cameraDistance − radius)·k, radius·0.001)` updated with the fit/zoom range, keeping log-depth as belt.
2. Tighten near/far enough to **drop log-depth entirely** (perf win) — needs the recenter (#1631) as a prerequisite, which is now in.
3. Leave as-is, documented — log-depth is working.

Related setters to keep coherent: `ShareViewer.js` shadow-camera `diag·SHADOW_NEAR/FAR_FACTOR`, and `ProgressiveLoadSession`'s monotonic `camera.far` growth during load.

Parent: #1632

pablo-mayrgundter:
[...]

Issue #1632, opened 2026-07-24
Retrospective: georeferenced (large-coordinate) models jitter/tear in the demand render path (#1614 → #1631)


pablo-mayrgundter:
## Summary

Georeferenced IFC models — coordinates authored on national grids (Swiss LV95 etc.) that place geometry at 10⁶–10⁷ m — rendered with torn-looking geometry that visibly **shook on camera rotation but was stable on zoom**, and only on real GPUs. Reported on `test-models-private` vyzn/`388_4 (EDITED).ifc` and reproduced on public [`ISSUE_129_N1540…V17.ifc`](https://github.com/bldrs-ai/test-models/blob/main/ifc/thatopen/ifcfiles/ISSUE_129_N1540_17_EXE_MOD_448200_02_09_11SMC_IGC_V17.ifc).

**Regression window (user bisect): #1611 clean → #1614 broken.** This was blocking validation of the demand/tiled memory architecture on exactly the models it exists for.

## Root cause

#1614 switched cache-miss loads to the streamed/deferred open (`OpenModelStreamed` + `DEFER_GEOMETRY`, default-on). In the **browser**, that open does not honor `COORDINATE_TO_ORIGIN` — it hands Share raw source-world placements (`first conway flatTransform trans=(1661282.9, -15.5, -8178626.0)`, confirmed by live `[DEBUG-COORD]` instrumentation on deploy-preview-1630). The classic `OpenModel` path (#1611 and earlier) recenters to ~70 m.

At ~8.3×10⁶ m, float32 — the GPU vertex format and the `BatchedMesh` instance-matrix texture — resolves to **~1 m**. Every vertex snaps to a metre-scale grid that shifts with the view transform: whole-model shake on rotate, stable on zoom.

**The investigation's biggest trap:** the *node* build of the same conway versions (1.432.1305 and 1.436.1311, same settings) recenters correctly on the streamed open — so every headless data-parity check (geometry, world AABBs, coordination matrix, builder output, float32 simulation) came back clean, and several wrong render-level theories got chased first.

## Fix

[...]

PR #1653, opened 2026-07-24
Request a stencil buffer: restore 24-bit depth precision (fixes haus.ifc z-fighting)

pablo-mayrgundter:
## The mechanism, finally

**three r163 changed `WebGLRenderer`'s default from `stencil: true` to `stencil: false`. Without a stencil buffer, ANGLE/Chrome may allocate a 16-bit depth buffer instead of the packed 24-bit D24S8.** Under our log-depth config (near 0.1, far ~3000 m), D16 gives ~**2.4 mm** depth resolution at 15 m viewing distance vs ~9 µm at 24-bit. The FZK-Haus roof interfaces (soffit, rake, ridge) sit 0–1.6 µm apart — coplanar for either format, but at D16 *everything within millimetres* z-fights, which is the speckle. The fix: request `stencil: true` explicitly on both renderer constructions (main + screenshot).

## How it was found

Preview bisection (probes #1644–#1651) narrowed Pablo's deploy-preview bisect to #1509's compression commits, but every static-analysis suspect there (keepNames, eager meshopt WASM init) was disproven by clean-cache probe tests. The breakthrough came from **building the good (`eafc88a0`) and bad (`e6f13bf8`) trees locally and diffing complete runtime renderer state headless** (via a `__THREE_DEVTOOLS__` hook — stable across reruns):

| | good tree | bad tree |
|---|---|---|
| granted context `stencil` | **true** | **false** |
| renderer `outputColorSpace` | *(property absent)* | `srgb-linear` |

The absent `outputColorSpace` (added in three ~r152) is the tell: the good tree's renderer was constructed by the **vendored web-ifc-viewer's three 0.135** (stencil default *true* → D24S8), while the compression commits' dependency reshuffle flipped module resolution so the same code constructed with **three 0.184** (stencil default *false* → D16). Nobody changed renderer code in #1509 — the *default* changed underneath it via which three copy resolved.

Today's `IfcRenderer` builds with 0.184 defaults directly, so the same D16 downgrade has been live on prod ever since — one continuous mechanism from May to now, and why the artifact survived the keepNames (#1644), draw-order (#1643), and lazy-WASM (#1651) probes untouched.

## Side effects and related issues

- Memory: the stencil8 attachment is the cost (the D24 half of D24S8 is typically allocated anyway on desktop GPUs); it's what Share shipped with for years pre-#1509.
- #1637's depth-budget analysis assumed 24-bit codes; at D16 its numbers are 256× worse — this fix restores the analysis's premise. #1637 stays open for near-plane tuning proper.
- #1643 (stable opaque draw order) remains correct and worth keeping: exactly-0-separation ties (ridge miter, rafters) resolve by draw order at *any* depth precision.
- Headless couldn't show the artifact because SwiftShader grants a float depth buffer regardless — the #1632 method note strikes again.

**Test:** haus.ifc (clean cache) on this preview vs prod — roof soffit/rake should be stable here. The diagnostic probe PRs #1644–#1651 can all be closed once this confirms.

Fixes the FZK-Haus z-fighting report (bisected to #1509 by Pablo).

---

### DIFF
--- a/src/viewer/three/context/renderer/renderer.js
+++ b/src/viewer/three/context/renderer/renderer.js
@@ -22,6 +22,13 @@ export class IfcRenderer extends IfcComponent {
       antialias: true,
       logarithmicDepthBuffer: true,
       preserveDrawingBuffer: pdbEnabled,
+      // three r163 flipped the default to false, and without a stencil
+      // ANGLE/Chrome may allocate a 16-bit depth buffer (~2.4mm depth
+      // resolution at 15m under log-depth, vs ~9µm with the packed
+      // 24-bit D24S8) — coplanar BIM interfaces (soffits, rakes,
+      // ridges) z-fight visibly. Explicit so a default change can't
+      // silently downgrade depth precision again.
+      stencil: true,
     })
     // For debugger tracing
     this.renderer.preserveDrawingBufferENABLED = pdbEnabled
@@ -73,6 +80,9 @@ export class IfcRenderer extends IfcComponent {
         antialias: true,
         logarithmicDepthBuffer: true,
         preserveDrawingBuffer: process.env.THREE_PDB_IS_ENABLED || false,
+        // Same depth-precision requirement as the main renderer above:
+        // screenshots must not z-fight where the live view doesn't.
+        stencil: true,
       })
       this.tempRenderer.localClippingEnabled = true
     }


pablo-mayrgundter:
Closing — theory falsified by direct measurement (prod already grants 24-bit depth without a stencil; forcing `stencil: true` changed nothing on a real GPU). The actual mechanism turned out to be the postprocessing composer's depth attachment widening from D16 to D24 at the three upgrade; fix in #1655.

---

PR #1654, opened 2026-07-26
PROBE: z-fight A/B lab — bakeXforms / logDepth / nearMin flags (+ stencil & draw order)

pablo-mayrgundter:
**Diagnostic probe — do not merge.** Updated after the depth-format theory was falsified by Pablo's real-GPU measurement (prod already grants `depthBits 24` without stencil) and this preview still fought with stencil + stable draw order combined.

This round adds URL-flag A/B toggles — all renderer-side, none alter authored geometry:

| Flag | What it tests |
|---|---|
| `?bakeXforms=1` | Bakes each placement transform into a per-instance geometry copy (CPU double precision, identity instance matrices) in the incremental batched builder — reproducing the classic merged-path vertex math inside the batched architecture. Hypothesis: BatchedMesh's float32 vertex-shader instance transform (1 ulp ≈ 1–4 µm at 10–50 m world coordinates) is the tie-breaking noise behind the post-1.438 speckle; classic-era CPU-double baking is why it used to be stable. |
| `?logDepth=0` | Disables the logarithmic depth buffer (conventional D24). |
| `?nearMin=N` | Raises the near-plane floor (e.g. `nearMin=1`); only meaningful with `logDepth=0` — conventional depth resolution scales with near, log depth doesn't. |

The `[zfight-probe]` console line now also reports `logDepth=`. Headless-verified: flags activate on model routes, bake produces identity instance matrices with per-instance geometry, unit tests pass.

**Important:** flags only survive on a full model-route URL (bare `/share` redirects to the default model and strips the query). Test like:

```
…/share/v/gh/…/haus.ifc?bakeXforms=1
```

**Test sequence on haus.ifc roof (each vs. plain preview URL):**
1. `?bakeXforms=1` — if stable, the float32 batching-matrix noise theory is confirmed, and the production fix direction is precision-aware instance transforms (bake for small models / relative-to-anchor matrices), no geometry change.
2. `?logDepth=0&nearMin=1` — conventional-depth A/B for the depth-strategy redesign discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.1642.7c1f9521",
+  "version": "1.1639.4efa7182",
   "main": "src/index.jsx",
   "license": "AGPL-3.0",
   "homepage": "https://github.com/bldrs-ai/Share",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.1639.4efa7182",
+  "version": "1.1639.163c0618",
   "main": "src/index.jsx",
   "license": "AGPL-3.0",
   "homepage": "https://github.com/bldrs-ai/Share",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.1533.772dc90",
+  "version": "1.1642.7c1f9521",
   "main": "src/index.jsx",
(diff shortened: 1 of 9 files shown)


pablo-mayrgundter:
Closing — this probe lab did its job: the `?depth16=1` A/B here confirmed the mechanism on a real GPU, and the production fix merged in #1655. Durable engine-side follow-up: bldrs-ai/conway#439.

---

PR #1683, opened 2026-07-27
Open local recents by OPFS storage id, not display filename

### LINKED ISSUE #1682: Re-opening a local model from Open &gt; Local Recents fails ("Loader could not read model")
## Repro

1. Load a local model (e.g. [`box.ifc`](https://github.com/bldrs-ai/test-models/blob/main/ifc/misc/box.ifc)) into Share via drag-and-drop or Open > Local, then close the page.
2. Re-visit bldrs.ai, Open > Local — the model is listed under Recents. Select it.
3. Error dialog: `Error: Loader could not read model`.

URL is `bldrs.ai/share/v/new/box.ifc`. Console:

```
Model: box.ifc — IFC, 0.0 MB
No type detected when constructing model
Error: parseIfcWithConway: OpenModel returned -1
    at conwayDirectIfcLoader.js:153:13
Error: Loader could not read model
    at Loader.js:1444:11
```

Reproduces with `haus.ifc` too. `index.ifc` re-loads *fine* — see "why index.ifc works" below.

## Root cause

`RecentFileEntry` for a local upload has two distinct name-ish fields, by design:

* `id` — the **OPFS storage id**, `<blob-uuid>.<ext>` (what `saveDnDFileToOpfs` / `opfsWriteModel` write under)
* `name` — the **original user-facing filename**, e.g. `box.ifc`

`Loader.js#uploadedDisplayFileName` follows that contract (looks up `entry.id === storageId`, returns `entry.name` for display). But the recents click handler does not:

```js
// src/Components/Open/OpenModelDialog.jsx
const handleOpenLocalRecent = (entry) => {
  disablePageReloadApprovalCheck()
  navigateToModel(`${appPrefix}/v/new/${entry.name}`, navigate)   // ← display name, not storage id
  setIsDialogDisplayed(false)
}
```

So we navigate to `/share/v/new/box.ifc`. In `Loader#load`, uploads are recognised by `testUuid(path)` — `box.ifc` is not a UUID, so the load falls out of the upload branch and is classified `isLocallyHostedFile`, which fetches `/box.ifc` from the origin. The SPA's catch-all returns `index.html`, and Conway is handed HTML: `No type detected when constructing model` → `OpenModel returned -1` → `Loader could not read model`.

**Why `index.ifc` works:** it is genuinely served at the site root (the default sample model), so the accidental origin fetch returns a real IFC. It never touches OPFS at all — a coincidence, not a working code path.

## Secondary defects in the same flow

1. **Open > Local (file picker) records the storage id as the display name.** `OpenModelDialog#openFile`'s `onLoad` only receives the storage id, and sets both `id` and `name` to it, so that recent renders as `A1B2C3D4-….ifc` instead of `box.ifc`. (Drag-and-drop gets this right.)
2. **Drag-and-drop writes `lastModifiedUtc` as an ISO string** (`new Date(file.lastModified).toISOString()`) where the interface and `RecentFilesList#formatRelativeTime` expect epoch ms — so the Last-modified column renders `NaNm ago` for dropped files.

## Fix

* Navigate local recents by storage id, not display name; record an explicit `sharePath` on local entries at creation so the nav target is stored rather than re-derived (and fall back to `entry.id` for entries already in localStorage).
* Thread the picked file's original name through `loadLocalFile` so the picker records the same `id`/`name` split drag-and-drop does.
* Store `lastModifiedUtc` as epoch ms in the drop handler.

pablo-mayrgundter:
Fixes #1682.

Re-opening a local upload from Open > Local > Recents crashed with `Loader could not read model`.

## Root cause

A local `RecentFileEntry` carries two distinct name-ish fields, by design:

* `id` — the **OPFS storage id** the upload was written under, i.e. the blob UUID plus the type suffix (`ADD77535-….ifc`)
* `name` — the user's **original filename**, e.g. `box.ifc`

`Loader#uploadedDisplayFileName` already honours that split (looks up by `id`, displays `name`). The recents click handler did not — it navigated to `/v/new/${entry.name}`.

`Loader#load` recognises an upload by `testUuid(path)`. A display name like `box.ifc` fails that test, so the load fell out of the upload branch, was classified `isLocallyHostedFile`, and fetched `/box.ifc` from the origin — where the SPA catch-all returns `index.html`. Conway was handed HTML:

```
No type detected when constructing model
Error: parseIfcWithConway: OpenModel returned -1
Error: Loader could not read model
```

`index.ifc` re-loaded fine only because it is genuinely served at the site root (the default sample model) — the accidental origin fetch returned a real IFC. It never touched OPFS at all.

## Changes

* **`OpenModelDialog#handleOpenLocalRecent`** navigates by storage id. Local entries now record an explicit `sharePath` at creation so the nav target is stored rather than re-derived; the `entry.id` fallback covers entries already in localStorage.
* **`OpenModelDialog#openFile`** records the picked file's original name as the display `name` instead of the storage id — the picker's recents row rendered a raw UUID where drag-and-drop rendered `box.ifc`. `loadLocalFile`/`loadLocalFileFallback` now pass the filename as a third `onLoad` arg.
* **`handleFileDrop`** stores `lastModifiedUtc` as epoch ms. It was writing an ISO string where `RecentFileEntry` and `RecentFilesList`'s `Date.now() - utcMs` expect a number, so dropped files showed `NaNm ago` in the Last-modified column.
* **`RecentFilesList#formatRelativeTime`** coerces on read, so entries already carrying an ISO string from before this fix still render instead of showing `NaNm ago` until they roll out of the per-source top-4.
* `RecentFileEntry`'s `id` / `sharePath` / `lastModifiedUtc` docs now spell out the storage-id and epoch-ms contracts for local entries.

## Testing

`yarn lint` and the full Jest suite pass (2067 tests, via the precommit hook).

New Jest cover in `OpenModelDialog.test.jsx` (Local tab): navigation by storage id for a legacy entry with no `sharePath`, navigation via a stored `sharePath`, display name in the row, and the picker's id/name/sharePath recording. Verified these fail against the pre-fix handler (3 of 5 red).

`dragAndDrop.test.js` covers the id/name split and epoch-ms `lastModifiedUtc`; `loader.test.js` covers the new `onLoad` filename arg; `RecentFilesList.test.jsx` covers the legacy ISO-string and unparseable-value cases.

`OpenModelDialog.spec.ts` gains an e2e that seeds a realistic recent (uuid id, `box.ifc` name), clicks it, and asserts the URL lands on the storage id. The existing e2e seeded `id === name === 'box.ifc'`, which could not have caught this; it now uses the realistic shape too. Both pass locally.

https://claude.ai/code/session_01LF3ZSRFpcXJmg5DXvGmD5Y

### DIFF
--- a/src/Components/Connections/RecentFilesList.jsx
+++ b/src/Components/Connections/RecentFilesList.jsx
@@ -38,11 +38,21 @@ function FileTypeIcon({mimeType}) {
 /**
  * Formats a UTC timestamp (ms) as a relative or absolute date string.
  *
- * @param {number|null|undefined} utcMs
+ * Tolerates an ISO-8601 string as well as epoch ms: the drag-and-drop
+ * handler stored ISO strings until #1682, and those entries are still
+ * sitting in users' localStorage — arithmetic on them yielded NaN and
+ * rendered "NaNm ago". Anything unparseable returns null, same as a
+ * missing timestamp.
+ *
+ * @param {number|string|null|undefined} utc
  * @return {string|null}
  */
-function formatRelativeTime(utcMs) {
-  if (!utcMs) {
+function formatRelativeTime(utc) {
+  if (!utc) {
+    return null
+  }
+  const utcMs = typeof utc === 'number' ? utc : Date.parse(utc)
+  if (isNaN(utcMs)) {
     return null
   }
   const diffMs = Date.now() - utcMs
--- a/src/Components/Connections/RecentFilesList.test.jsx
+++ b/src/Components/Connections/RecentFilesList.test.jsx
@@ -91,6 +91,24 @@ describe('RecentFilesList', () => {
     expect(screen.getByText('2d ago')).toBeInTheDocument()
   })
 
+  // Drag-and-drop stored ISO strings until #1682; those entries are
+  // still in users' localStorage and must not render "NaNm ago".
+  it('renders a legacy ISO-string timestamp as relative time', () => {
+    const legacy = [{id: 'file-4', source: 'local', name: 'Old.ifc',
+      lastModifiedUtc: new Date(Date.now() - MS_PER_MINUTE).toISOString()}]
+    render(<RecentFilesList files={legacy} onOpen={jest.fn()}/>, {wrapper: ThemeCtx})
+
+    expect(screen.getByText('1m ago')).toBeInTheDocument()
+  })
+
+  it('renders no timestamp for an unparseable value', () => {
+    const bad = [{id: 'file-5', source: 'local', name: 'Bad.ifc', lastModifiedUtc: 'not-a-date'}]
+    render(<RecentFilesList files={bad} onOpen={jest.fn()}/>, {wrapper: ThemeCtx})
+
+    expect(screen.getByText('Bad.ifc')).toBeInTheDocument()
+    expect(screen.queryByText(/ago/)).not.toBeInTheDocument()
+  })
+
   it('shows column headers', () => {
     render(<RecentFilesList files={mockFiles} onOpen={jest.fn()}/>, {wrapper: ThemeCtx})
 
--- a/src/Components/Open/OpenModelDialog.jsx
+++ b/src/Components/Open/OpenModelDialog.jsx
@@ -206,16 +206,27 @@ export default function OpenModelDialog({
   }
 
   const openFile = () => {
-    const onLoad = (filename, lastModifiedUtc) => {
+    /**
+     * @param {string} storageId OPFS storage id, `<blob-uuid>.<ext>` — the
+     *   `/v/new/` path segment the Loader resolves against OPFS.
+     * @param {number} lastModifiedUtc Epoch ms from the picked File.
+     * @param {string} [originalName] The user's filename, e.g. `box.ifc`.
+     */
+    const onLoad = (storageId, lastModifiedUtc, originalName) => {
       disablePageReloadApprovalCheck()
-      navigateToModel(`${appPrefix}/v/new/${filename}`, navigate)
+      const sharePath = `${appPrefix}/v/new/${storageId}`
+      navigateToModel(sharePath, navigate)
       addRecentFileEntry({
-        id: filename,
+        id: storageId,
         source: 'local',
-        name: filename,
+        // `name` is the display name; `id` is what the Loader resolves.
+        // Keeping them distinct is what lets the recents row read
+        // "box.ifc" while navigation still targets the OPFS entry.
+        name: originalName || storageId,
         lastModifiedUtc: lastModifiedUtc || null,
+        sharePath,
       })
-      setPendingModelNameUpdate(filename)
+      setPendingModelNameUpdate(storageId)
     }
     if (isOpfsAvailable) {
       loadLocalFile(onLoad, false)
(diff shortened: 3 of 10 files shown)


pablo-mayrgundter (commented):
## Review

### Overview

Fixes the #1682 crash by correcting an `id`/`name` conflation in local `RecentFileEntry` handling. `handleOpenLocalRecent` navigated to `/v/new/${entry.name}` — the user's filename — where the Loader's upload detection (`testUuid`) needs the OPFS storage id. The wrong branch then fetched the filename from the origin, got the SPA catch-all's `index.html`, and handed HTML to Conway. The diff routes navigation through the storage id, records `sharePath` at entry creation, and cleans up two adjacent defects (picker recording the UUID as display name; ISO-string `lastModifiedUtc`).

The diagnosis holds up against the code: `Loader.js:119` gates on `testUuid(path)`, `Loader.js:127` (`uploadedDisplayFileName`) already looks entries up by `id` and displays `name`, and the recents handler was the sole violator of that contract. The `index.ifc`-works asymmetry is explained by that file genuinely existing at the site root, not by any working code path.

### Correctness

- The nav-target change is right, and the `entry.sharePath || ${appPrefix}/v/new/${entry.id}` fallback correctly covers entries already in localStorage — both prior creation paths (DnD, picker) stored the storage id in `id`.
- Threading `file.name` as a third `onLoad` arg is applied to all three call sites in `utils/loader.js` (worker `write`, worker `read`, and the `testingDisableWebWorker` branch) plus the fallback — no path left behind.
- `lastModifiedUtc: uploadedFile.lastModified || null` matches every other writer in the codebase (`useGithubLastModified` back-fills with `.getTime()`, the Drive picker with `LAST_EDITED_UTC`).

### Issue found and fixed

**Legacy ISO-string timestamps still rendered `NaNm ago`.** Correcting the *writer* doesn't help users who already dropped a file — those entries sit in localStorage until they roll out of the per-source top-4, and `formatRelativeTime` did `Date.now() - utcMs` on a string. Fixed in `15151a3` by coercing on read (`Date.parse` for strings, null for unparseable), with cover for both the legacy and garbage cases, plus a note on the `RecentFileEntry` field that nothing should write a string going forward.

### Observations (no change made)

- **`sharePath` on local entries is redundant with `id` + current `appPrefix`.** It buys consistency with the github/drive entries and matches what `RecentFileEntry` already documented, but the derived fallback has to exist forever for legacy entries anyway, so it's a second source of truth. Staleness would need the app prefix to change within one origin, which doesn't happen today — noting it rather than churning.
- **`navigateToModel(entry.sharePath, …)` passes a localStorage value to `window.location.assign`.** Pre-existing pattern (`handleGithubOpenById` does the same) and it presupposes same-origin write access, i.e. an attacker who already has XSS. Not a regression from this diff.
- **`checkOPFSAvailability` is `async`, so `isOpfsAvailable` in `OpenModelDialog` is always a truthy Promise** — production always takes the `loadLocalFile` branch, never `loadLocalFileFallback`. The new unit tests therefore drive the fallback (the module mock returns `false`) while production drives the other branch. Behaviourally equivalent here since both receive the same `onLoad` closure, but the always-truthy check is a separate latent bug worth its own issue.
- **`loadLocalFileFallback` returns a blob UUID with no extension**, so a recent created on that path navigates to a `/v/new/<uuid>` whose blob URL was already revoked. Broken before this PR, not made worse by it, out of scope.

### Test coverage

Good. Five new Jest cases on the Local tab cover both the legacy-entry and stored-`sharePath` nav paths, the display name, and the picker's recording; 3 of 5 were verified to go red against the pre-fix handler, so they pin the bug rather than the implementation. The e2e's fixture previously had `id === name === 'box.ifc'`, which structurally could not have caught this — reshaping it to a realistic UUID/filename split is the right call, and the added URL assertion closes the loop end to end.

### Style

Matches the repo's conventions — comments carry the load-bearing "why" (the `testUuid` classification, the epoch-ms arithmetic) rather than restating the code, and the JSDoc on the changed `onLoad` signatures makes the id/name distinction explicit at every boundary a future caller would touch.

**Verdict:** correct, well scoped, well covered. CI green on `15151a3` (build, playwright, playwright-webifc).

---

PR #1687, opened 2026-07-27
Write OPFS bytes via sync access handle, not createWritable (Safari GLB cache)

### LINKED ISSUE #1686: GLB cache never writes on Safari — OPFS writes use createWritable() instead of createSyncAccessHandle()
## Symptom

On Safari, every load is a GLB cache MISS. The model loads fine, but the post-parse GLB writer always fails, so the next load re-parses from source instead of hitting the cache. Chrome reaches the cache-hit path normally.

Console on Safari (`Share v1.1683.9e1ed5c`, FZK-Haus, 2.4 MB):

```
[glb] "reader: upload cache MISS, will export after parse"
...
[glb] "writer: upload source, key=BldrsLocalStorage/V1/Projects/73d7ebf5-…"
[glb] "writer: skipped (threw); reader will fall back to source on next load"
      Error: writeBytesByPath: i[1].createWritable is not a function.
      (In 'i[1].createWritable()', 'i[1].createWritable' is undefined) — utils.js:404
[glb] "source spill: released resident source buffer (2526544B); property reads now page 4MiB windows from OPFS on demand"
```

Note the last line: OPFS itself is healthy on Safari — reads work, and the source spill pages windows out of OPFS successfully. This is not an "OPFS unavailable" case, so `checkOPFSAvailability()` correctly returns true and no fallback kicks in.

## Root cause

Safari implements OPFS writes **only** through `FileSystemSyncAccessHandle` (worker-only). `FileSystemFileHandle.createWritable()` is not implemented there.

`src/OPFS/OPFS.worker.js` already knows this — every write in it goes through `createSyncAccessHandle()` (lines 294, 566, 1348, 1705, 1801), and there is a dedicated Safari hardening helper, `openSyncAccessHandleWithRetry()` (line 387), whose doc comment describes working around Safari's `InvalidStateError` on stale internal handle refs. `renameFileInOPFS()` similarly documents Safari's two-arg `move()` signature.

Two later additions break that convention and call `createWritable()` instead:

1. **`writeBytesByPathToOPFS()` (line 1517)** — the GLB cache writer. This is the reported failure: it throws on every Safari write, and because the writer is deliberately fail-soft (`writer: skipped (threw); reader will fall back to source on next load`) there is no error dialog — just a permanently cold cache.
2. **`renameFileInOPFS()` (line 1424)** — the stream-copy leg of the rename fallback. Latent rather than active: the preferred `fileHandle.move()` path does work on Safari, so this only runs if `move()` fails. It would fail the same way if it ever ran.

## Why tests didn't catch it

The in-memory `FileHandle` stub in `src/OPFS/OPFS.worker.test.js` implements **both** `createSyncAccessHandle()` (line 102) and `createWritable()` (line 162). It models Chrome, so a Safari-only API gap is invisible to the suite.

## Fix

* Route both call sites through `createSyncAccessHandle()`, reusing `openSyncAccessHandleWithRetry()` so they pick up the existing Safari `InvalidStateError` retry.
* Keep the rename fallback's chunked/streaming memory profile — it exists to avoid a large `ArrayBuffer` on iOS, so the sync-handle version must slice the source rather than buffer it whole.
* Add `truncate()`/`flush()` to the stub's sync handle (both exist on the real API), and add a Safari-shaped handle stub with **no** `createWritable`, so the write paths are covered against the browser that actually lacks it.

## Scan

Grepped `src/` for the rest of this class of hazard. All OPFS filesystem access is confined to `OPFS.worker.js` (nothing in `Loader.js`, `OPFSService.js`, or `GlbWriter.worker.js` touches handles directly), so the two sites above are the complete `createWritable` surface. `requestIdleCallback` is feature-detected (`Loader.js:85`), and `fileHandle.move()` already has cross-browser fallbacks. `crypto.randomUUID()` is used unguarded in the two OAuth providers — Safari 15.4+, secure-context only, so it's within a reasonable baseline and left alone.

Follow-up to #1682 / #1683 (unrelated defect, same "local model doesn't come back" family of symptoms).

pablo-mayrgundter:
Fixes #1686.

On Safari every load was a GLB cache MISS — the model rendered fine, but the post-parse writer always threw, so the next load re-parsed from source. Chrome reached the cache-hit path normally.

```
[glb] "writer: skipped (threw); reader will fall back to source on next load"
      Error: writeBytesByPath: i[1].createWritable is not a function — utils.js:404
[glb] "source spill: released resident source buffer (2526544B); property reads now page 4MiB windows from OPFS on demand"
```

That last line matters: OPFS itself is healthy on Safari — reads work and the source spill pages windows out of it successfully. This is not an "OPFS unavailable" case, so `checkOPFSAvailability()` correctly returns true and no fallback engages.

## Root cause

Safari implements OPFS writes **only** through `FileSystemSyncAccessHandle` (worker-only). `FileSystemFileHandle.createWritable()` is undefined there.

`OPFS.worker.js` already handles this correctly nearly everywhere: every other write goes through `createSyncAccessHandle()`, and `openSyncAccessHandleWithRetry()` exists specifically to absorb Safari's `InvalidStateError` on stale internal handle refs. `renameFileInOPFS` even documents Safari's two-arg `move()` signature.

Two sites broke that convention:

1. **`writeBytesByPathToOPFS`** — the GLB cache writer. The live break. Silent, because the writer is deliberately fail-soft, so the only symptom is a permanently cold cache.
2. **`renameFileInOPFS`'s copy fallback** — latent. The preferred `fileHandle.move()` does work on Safari, so the stream-copy leg only runs if that fails.

## Changes

* Both sites now go through `openSyncAccessHandleWithRetry()`, inheriting the existing Safari retry.
* `writeBytesByPathToOPFS` truncates after writing — unlike a writable stream, a sync handle doesn't implicitly discard prior contents, so overwriting an entry with a shorter payload would otherwise leave a stale tail.
* The rename fallback keeps its memory profile. It exists to avoid a large `ArrayBuffer` on iOS, and a sync handle has no stream interface, so the source is copied in 4MiB slices rather than read whole.

## Why this shipped green

The in-memory `FileHandle` stub in `OPFS.worker.test.js` implements **both** `createSyncAccessHandle()` and `createWritable()` — it models Chrome, so a Safari-only API gap was invisible.

The stub now also carries `truncate()`/`flush()` (both real API), and a new describe block deletes `createWritable` from it to exercise both write paths against the browser that actually lacks it. Verified those tests fail against the old code with the exact production error (`createWritable is not a function`), each reverted site independently.

Coverage added: bytes land correctly; a shorter re-write truncates rather than leaving a tail; the rename copy fallback works; and a >4MiB payload spans multiple chunks so the slice loop is actually exercised.

## Scan

Per review request, swept `src/` for the rest of this class of hazard. All OPFS filesystem access is confined to `OPFS.worker.js` — nothing in `Loader.js`, `OPFSService.js`, or `GlbWriter.worker.js` touches handles directly — so these two were the complete `createWritable` surface. `requestIdleCallback` is feature-detected (`Loader.js:85`) and `fileHandle.move()` already has cross-browser fallbacks. `crypto.randomUUID()` is unguarded in the two OAuth providers, but that's Safari 15.4+ in a secure context, within a reasonable baseline; left alone.

## Testing

`yarn lint` clean; full Jest suite passes (2071 tests) via the precommit hook.

Not verified on a real Safari — I don't have one in this environment. The fix moves both sites onto the exact API the rest of this worker already uses successfully on Safari, but a spot-check that the second load now logs a cache HIT would be worth doing before merge.

https://claude.ai/code/session_01LF3ZSRFpcXJmg5DXvGmD5Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01LF3ZSRFpcXJmg5DXvGmD5Y)_

### DIFF
--- a/src/OPFS/OPFS.worker.js
+++ b/src/OPFS/OPFS.worker.js
@@ -1434,9 +1434,14 @@ export async function renameFileInOPFS(parentDirectory, fileHandle, newFileName,
   // interface, so we slice the source ourselves to hold the same memory
   // profile.
   const srcFile = await fileHandle.getFile()
-  const destHandle = await parentDirectory.getFileHandle(newFileName, {create: true})
-  const {handle: destAccess} = await openSyncAccessHandleWithRetry(
-    parentDirectory, destHandle, newFileName)
+  const openedDest = await parentDirectory.getFileHandle(newFileName, {create: true})
+  // Take the handle the helper hands back, not the one we passed in: on
+  // the Safari `InvalidStateError` retry it removes and recreates the
+  // entry, so `openedDest` is stale from that point on and the bytes
+  // land in the fresh handle. Returning the stale one would hand the
+  // caller a handle to a file it never wrote.
+  const {handle: destAccess, fileHandle: destHandle} = await openSyncAccessHandleWithRetry(
+    parentDirectory, openedDest, newFileName)
   try {
     for (let at = 0; at < srcFile.size; at += COPY_CHUNK_BYTES) {
       const slice = await srcFile.slice(at, Math.min(at + COPY_CHUNK_BYTES, srcFile.size)).arrayBuffer()
--- a/src/OPFS/OPFS.worker.js
+++ b/src/OPFS/OPFS.worker.js
@@ -13,6 +13,14 @@ let GITHUB_BASE_URL_AUTHENTICATED = null
 let GITHUB_BASE_URL_UNAUTHENTICATED = null
 
 
+// Slice size for the chunked file copy in `renameFileInOPFS`'s fallback.
+// 4MiB matches the source-spill window size Conway pages from OPFS, and
+// keeps peak memory bounded on iOS where the whole point of copying in
+// pieces is to avoid materializing a large ArrayBuffer.
+// eslint-disable-next-line no-magic-numbers
+const COPY_CHUNK_BYTES = 4 * 1024 * 1024
+
+
 /**
  * Pick the GitHub API base URL: use the OAuth-injecting bldrs proxy when a
  * user token is present, otherwise hit GitHub directly.
@@ -1417,12 +1425,30 @@ export async function renameFileInOPFS(parentDirectory, fileHandle, newFileName,
     }
   }
 
-  // Fallback: stream copy + delete (avoids large ArrayBuffer on iOS).
+  // Fallback: chunked copy + delete (avoids large ArrayBuffer on iOS).
+  //
+  // Copied in slices rather than read whole: keeping peak memory to one
+  // chunk is the entire point of this fallback on iOS. This used to be a
+  // `stream().pipeTo(createWritable())`, but Safari has no
+  // `createWritable` (#1686) — and a sync access handle has no stream
+  // interface, so we slice the source ourselves to hold the same memory
+  // profile.
   const srcFile = await fileHandle.getFile()
   const destHandle = await parentDirectory.getFileHandle(newFileName, {create: true})
-  const readable = srcFile.stream()
-  const writable = await destHandle.createWritable()
-  await readable.pipeTo(writable) // closes writable on completion
+  const {handle: destAccess} = await openSyncAccessHandleWithRetry(
+    parentDirectory, destHandle, newFileName)
+  try {
+    for (let at = 0; at < srcFile.size; at += COPY_CHUNK_BYTES) {
+      const slice = await srcFile.slice(at, Math.min(at + COPY_CHUNK_BYTES, srcFile.size)).arrayBuffer()
+      await destAccess.write(new Uint8Array(slice), {at})
+    }
+    await destAccess.truncate(srcFile.size)
+    await destAccess.flush()
+  } finally {
+    try {
+      await destAccess.close()
+    } catch (_) {/* idempotent */}
+  }
 
   try {
     if (typeof fileHandle.remove === 'function') {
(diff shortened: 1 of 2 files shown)


pablo-mayrgundter (commented):
## Review

### Overview

Moves the two OPFS writes that used `FileSystemFileHandle.createWritable()` onto `createSyncAccessHandle()`, via the existing `openSyncAccessHandleWithRetry()` helper. Safari implements OPFS writes only through sync access handles (worker-only), so `createWritable` is undefined there — which silently disabled the GLB cache on Safari, since the writer is fail-soft and only logs `writer: skipped (threw)`.

Confirmed working on real Safari by the reporter (second load now hits the cache).

### Correctness

- The diagnosis is well-supported by the surrounding code rather than just the stack trace: every other write in this worker (lines 294, 566, 1348, 1705, 1801) already uses `createSyncAccessHandle()`, and `openSyncAccessHandleWithRetry` exists specifically to absorb Safari's `InvalidStateError`. The two converted sites were the outliers, not the convention.
- `truncate(payload.byteLength)` after the write is a genuine requirement, not belt-and-braces: a writable stream implicitly replaces contents, a sync handle does not, so overwriting an entry with a shorter payload would otherwise leave a stale tail. Covered by a dedicated test.
- The chunked copy preserves the fallback's reason for existing (bounded peak memory on iOS). `getFile()` returns a lazy disk-backed `File` and each `slice().arrayBuffer()` materializes exactly one chunk, so the memory profile matches the old `stream().pipeTo()`.
- Zero-length source is fine: the loop doesn't execute, `truncate(0)` + `flush()` still run.

### Issue found and fixed

**The chunked copy returned a stale handle on the retry path.** `openSyncAccessHandleWithRetry` recovers from `InvalidStateError` by removing and *recreating* the entry — which is exactly why its contract returns an updated `fileHandle` ("different object if we recreated"). The copy destructured only `{handle}` and then returned the pre-retry object, so whenever the Safari retry fired, the bytes went into the fresh handle while the caller received a handle to the just-recreated (empty) entry.

Narrow — it needs Safari *and* a stale internal handle ref — but it's precisely the case the helper exists to survive, so failing there defeats the point. Fixed in `5691aec`, with a test that forces the first `createSyncAccessHandle` to raise `InvalidStateError`. Verified it fails against the previous commit, with the stale handle showing `data: []`.

The same destructure in `writeBytesByPathToOPFS` discards the updated handle too, but nothing reads it after the write there, so it's harmless — left as-is rather than adding an unused binding.

### Test coverage

The interesting part of this PR. The reason the bug shipped is that the in-memory `FileHandle` stub implements **both** `createSyncAccessHandle` and `createWritable` — it models Chrome, so a Safari-only API gap was structurally invisible. Deleting `createWritable` from the prototype for a describe block is a clean way to pin that, and it generalizes: any future write that reaches for the Chrome-only API now fails in CI.

Both converted sites were verified to fail against the old code with the exact production error, reverted independently. Coverage beyond the happy path is appropriate — truncation of a longer prior entry, a >4MiB payload so the slice loop actually iterates, and now the `InvalidStateError` retry.

Worth noting the stub gained `truncate`/`flush` because the real API has them; that keeps it honest rather than papering over the new calls.

### Style

Consistent with the file. The `no-magic-numbers` disable on `COPY_CHUNK_BYTES` matches how `Loader.js:1665` and `loadProgress.js:53` handle byte math. Comments explain the non-obvious constraints — why a sync handle needs an explicit truncate, why the copy is sliced rather than streamed — which is what this codebase's comments are for.

### Risk

Low, and the blast radius is well understood. `writeBytesByPathToOPFS` is Chrome-unaffected in behaviour (same bytes, different API) and was already 100% broken on Safari, so there is no regression surface there. The rename fallback is the one path where Chrome behaviour changes, and it only runs when `move()` fails — already rare — with existing tests plus three new ones covering it.

**Verdict:** correct and well covered. CI green on `94edd83`; `5691aec` adds the retry fix.

---

PR #1723, opened 2026-08-06
Capture [glb] diagnostics in tests instead of printing them

pablo-mayrgundter:
Follow-up #1 of 3 from the test-log cleanup (#1722), which left the `[glb]` pipeline diagnostics as noted-remaining. This diverts them so a jest run leaves a quiet console (conway #301 §6) and asserts the error-path ones.

## Change

- **`glbLog.js`**: `glbInfo` / `glbVerbose` (and a new **`glbWarn`**) route through a **swappable sink** — console by default. `setGlbLogSink()` lets a harness capture instead of print.
- Moved the pipeline's stray direct `console.warn('[glb] …')` calls onto `glbWarn` so they share the one channel: `glbExport.js` (face_ids-capture skip) and `Loader.js` ×3 (reader face_ids failures / compressed-no-coverage skip).
- **`tools/jest/glbLogCapture.js`** + `setupTests.js`: install a capturing sink for the whole suite (buffer, no console output), cleared before each test. Specs read it via `getGlbLogs()`.
- **Asserted** the diagnostic fires in the three out-of-range-bufferView error-path tests (`BLDRS_face_ids` / `_element_properties` / `_spatial_tree`) — turning would-be noise into positive coverage.

## Result

Full jest suite green (**230 suites, 2173 passed**); **0 `[glb]` console lines** (was ~115). No production behavior change — the console sink is the default; only tests swap it.

## Stacked follow-ups

This is the base of a 3-PR stack: **#2 CadView/ViewerContainer `act()` warnings** targets this branch, **#3 single-`three`-instance jest resolution** targets #2.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HLsefWrsH8W6PWY9qh9xaU)_

### DIFF
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -45,7 +45,7 @@ import glbToThree from './glb'
 import {glbCacheKey} from './glbCacheKey'
 import {activeGlbCompressionMode, activeSchemaVersion} from './glbCompress'
 import {isBldrsGlbContainer, unpackGlbContainer} from './glbContainer'
-import {glbInfo, glbVerbose} from './glbLog'
+import {glbInfo, glbVerbose, glbWarn} from './glbLog'
 import {spillModelSource} from './opfsSourceByteStore'
 import {
   externalCacheKey,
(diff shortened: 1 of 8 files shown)


pablo-mayrgundter (commented):
I approve!

PR #1744, opened 2026-08-08
Derive the camera near plane from model bounds, not an absolute floor

### LINKED ISSUE #1742: Small STEP parts clip and vanish when zooming in: MIN_NEAR is an absolute 0.1 m floor
## Summary

`orbit-control.js` floors the camera near plane at an absolute constant:

```js
// src/viewer/three/context/camera/controls/orbit-control.js:22,139
const MIN_NEAR = 0.1
camera.near = Math.max(controls.minDistance * HALF, MIN_NEAR)
```

Everything else in the camera rig is derived from the model's own bounds (`minDistance`, `maxDistance`, `far`, the ground/shadow frustum), so it scales across model sizes. `MIN_NEAR` does not. For a part smaller than a couple of metres, the near plane sits *inside* the part: zooming in clips it, and past 0.1 m it disappears entirely.

## Why it matters now

This has been latent for STEP models because conway was scaling millimetre STEP files by the reciprocal of the correct unit factor — a 50 mm part loaded as 50 000 m, far above the floor. bldrs-ai/conway#458 (PR bldrs-ai/conway#460) fixes that, so millimetre STEP models now arrive at their true size. A `create-a-tube.step` that used to span 50 km now correctly spans 0.05 m, which is **below** `MIN_NEAR` — so correcting the units is what exposes this.

IFC models were never affected in practice: building-scale geometry is comfortably above the floor.

Repro after conway#460 lands: open any small STEP part (`data/create-a-tube.step`, 50 × 50 × 100 mm) and zoom in. `minDistance` computes to roughly 0.0024 m for it, while `near` is pinned at 0.1 m.

## Suggested fix

Derive the floor from the model like every other camera parameter — e.g. a small fraction of the bounding-sphere radius, clamped to a tiny absolute epsilon only to avoid a zero/denormal near plane:

```js
camera.near = Math.max(controls.minDistance * HALF, sphere.radius * NEAR_FRACTION, ABSOLUTE_MIN_NEAR)
```

Worth a regression check at both extremes — a millimetre-scale STEP part and a georeferenced building — since near/far ratio drives depth precision.

Found while verifying bldrs-ai/conway#460 against the Create E0 round trip.

pablo-mayrgundter:
Fixes #1742. Bumps conway twice: to 1.460 (bldrs-ai/conway#460, true-scale units — the trigger for everything here) and then 1.465 (bldrs-ai/conway#465, a preview-placement fix found while testing this PR).

## The bug

`fitCameraLimitsToModel` derives `minDistance`, `maxDistance` and `far` from the model&#39;s own bounds, but floored `near` at a flat constant:

```js
const MIN_NEAR = 0.1
camera.near = Math.max(controls.minDistance * HALF, MIN_NEAR)
```

For a part smaller than a couple of metres that floor sits *inside* the part — zooming in clips it, and past 0.1 it&#39;s gone. IFC was never affected; building-scale geometry sits well above the floor.

## The fix

`orbit-control.js` — replace the constant with the model-derived terms, so the whole frustum is scale-invariant:

```js
camera.near = Math.max(
  controls.minDistance * HALF,
  sphere.radius * NEAR_RADIUS_FRACTION,
  ABSOLUTE_MIN_NEAR)
```

Worth knowing on review: **`controls.minDistance * HALF` is what actually binds.** `minDistance = fitDistance * 0.01` and `fitDistance = radius / sin(½·limitingFOV) ≥ radius`, so the derived near is always ≥ `0.005·radius`. At `NEAR_RADIUS_FRACTION = 0.001` the middle term never fires today — it&#39;s a guard that only matters if `MIN_DISTANCE_FACTOR` is retuned tighter, and it sits deliberately below the derived value so it can&#39;t invert the `near &lt; minDistance` invariant. `ABSOLUTE_MIN_NEAR = 1e-6` only keeps `near` off zero/denormal for a degenerate sphere.

Depth precision: on the normal path `far/near` is ~3e3, unchanged at building scale and now identical at every model scale. It is *not* unconditional — when `maxDistance` comes from `currentDistance` (permalink restore, issue-card nav) rather than the 10x headroom, a millimetre part can reach a much higher ratio. `logarithmicDepthBuffer: true` covers that, and it beats the old behavior where such parts were invisible outright.

This derivation now lives in `three/cameraLimits.js`, shared by the end-of-load fit and the streaming camera follow — see the follow-on section for why that unification mattered.

## conway 1.450.1353 → 1.460.1363

The delta against 1.450.1353 is exactly conway#460 (`1 / sourceUnitInM` → `sourceUnitInM` on the root unit transform), its own test, and the version stamp. **No wasm change.**

Measured through Share, a millimetre STEP model&#39;s camera rig moves by exactly 1e6:

| | conway 1.450.1353 | conway 1.460.1363 |
|---|---|---|
| `gear.step` world size | 3300 m | **0.0033 m** |
| `minDistance` | 92 | 9.2e-5 |
| `near` (pre-fix) | 46 | 0.1 ← clips the 3.3 mm part |

### conway 1.460.1363 → 1.465.1375

Carries conway#465, root-caused during this PR&#39;s Snowdon testing: the parse-time preview channel extracted a product whose placement was still a dangling forward reference in the parse prefix as **unplaced** — on georeferenced Snowdon, door `#5014` emitted 88 preview payloads ~425 km from the building and the camera follow framed their union, rendering the model as a sub-pixel speck for the whole load. conway now defers such products to the durable pump. Preview-only change; durable geometry is byte-identical (conway regression: zero visual diffs), hence **no second GLB schema bump**.

## GLB cache invalidation

`BLDRS_GLB_SCHEMA_VERSION` 0.14.0 → 0.15.0. Geometry is baked into the GLB in **world coordinates** and nothing in the cache key identifies the engine — so without this, anyone holding a pre-1.460 artifact would keep loading 1e6x-oversized geometry, never pick up this fix, and disagree with cache-miss users on every shared camera / cut-plane link. d1a74ea is the precedent for pairing an engine bump with a schema bump.

## Testing

`yarn precommit` (eslint + typecheck + Jest) green on every commit via the husky hook; the suite has grown with the follow-on fixes below. Playwright E2E green at every push since the CI-workers fix.

**Unit** (`orbit-control.test.js`): a 50 mm part keeps `0 &lt; near &lt; minDistance`; every derived limit holds the same ratio across a 20000x size difference. Tolerance is 6 digits because `BoxGeometry` positions are `Float32Array`, so the two cubes aren&#39;t exactly 20000x apart to begin with.

**E2E** (`smallPartNearPlane.spec.ts`, desktop + mobile via `describeMobileAndDesktop`): loads `gear.step` and asserts `near &lt; extent` and `near &lt; minDistance`. Verified it **fails against the pre-fix floor** (`near` 0.1 vs a 0.0033 m part) on both form factors and passes with the fix.

Two things that made this spec non-obvious, both commented in place:
- Model bounds are read with `Box3.expandByObject`, not a hand walk over `geometry.boundingBox` — conway emits STEP as an `InstancedMesh` whose per-instance matrices carry the root unit scale, so a hand walk reports raw file coordinates (3.3, not 0.0033) and misses exactly what&#39;s being measured.
- The model&#39;s world size is asserted *before* the near-plane invariant, so a conway unit regression fails as itself instead of softening the real assertion into a tautology.

## CI workers

main moved these jobs to a 4-vCPU/8GB runner (6df8de7) but left `--workers=6` — six Chromiums (each rendering WebGL with the conway wasm heap resident) starved the box, which presented as flake: click/visibility timeouts and `#viewer-container` never mounting. Measured: 6 workers → 53 failed / 81 passed (main&#39;s own state); 3 → 4 failed / 134 passed; **2 → green in 11.7m** — faster in wall-clock than 3, which was spending its time on 30s timeouts and retries. Settled on 2 (Playwright&#39;s own vCPU/2 heuristic), with `timeout-minutes` 20 → 35 as headroom for the degraded case, and comments tying both numbers to `runs-on`.

## Follow-on fixes (same PR)

The conway rescale exposed three more absolute-scale assumptions of the same class as #1742 — originally listed here as known-not-fixed, now all fixed, one commit each:

- **Cut-plane arrow** (`MeshClipper.js`, `0a5b2e5`) — the absolute `[2, 40]` clamp is gone; the gizmo scales purely with model radius. The old MIN pinned it at ~600x a true-scale gear, drawn `depthTest:false` over the model.
- **Camera permalinks** (`58adbc0`) — `roundCoordComponent` rounds relative to scale (fixed decimals or significant digits, whichever lands closer), the read side no longer re-rounds through `floatStrTrim`, and the URL grammar accepts exponential notation. Placemarks got the same fix for free — on a millimetre part every placemark previously serialized to `0,0,0,0,0,0` and collided as a single map key.
- **Cut-plane offsets** (`4a17538`) — no longer quantized to 3 absolute decimals at any of their three hops (write, read, scene placement).

Streaming-load fixes that fell out of the same Snowdon investigation:

- `three/cameraLimits.js` — one shared limit derivation for the end-of-load fit and the streaming camera follow. The follow had been growing only `maxDistance`/`far`, leaving `minDistance`/`near` at their absolute defaults of 1 — clipping and shrinking sub-metre models for the whole load (the Arty symptom).
- The incremental `BatchedMesh` streams with frustum culling **off**: three caches a first-frame bounding sphere and never invalidates it as instances append, so the growing model was culled invisible until load end. `finalize()` computes real bounds and re-enables culling.
- The camera follow recovers from an over-inflated frame (refits were overflow-only, so one bad frame used to be permanent) and drops preview placements &gt;100x the accumulated preview radius — defense in depth beneath conway#465.
- Only the durable builder may decide the origin-recentre frame; previews read it and render unrecentred until the first durable batch (a mis-placed preview payload must never steer where the real model renders).
- `?feature=disableLoadOverlay` keeps the canvas interactive during load, for diagnosing streaming issues by eye.

Review hardening (from /review of this branch): gzip sniffing requires SPZ&#39;s decompressed `NGSP` magic instead of classifying every gzipped upload as a splat; the stale-LFS-pointer eviction path clears `opfsFile` and corrects the report&#39;s Source line; a failed Spark dynamic import retries instead of poisoning the session; test flush helpers are timer-agnostic per STYLE.md.

Smoke test: open a small STEP part (gear) — solid at full zoom, correctly-sized cut-plane gizmo, camera and `#cp:` share-links restore exactly. Load Snowdon — visible while streaming, no mid-load disappearance, `far placements: preview=0 | durable=0` in the console.

https://claude.ai/code/session_01KJCRAAy2HyxhUbx1PjprjA

### DIFF
--- a/.github/workflows/test-flows.yml
+++ b/.github/workflows/test-flows.yml
@@ -23,8 +23,11 @@ jobs:
     # is chosen so screenshot baselines regenerated in the dev/sandbox env
     # (also noble + chromium 1194) match CI's renders pixel-for-pixel.
     container: mcr.microsoft.com/playwright:v1.56.1-noble
-    # Headroom for the reduced worker count below: the suite took 12.8m at 3
-    # workers on this runner, so 20 left no margin at 2.
+    # Headroom, not a fitted bound: a clean run at the worker count below
+    # measured 11.7m on this runner. The margin is for the degraded case --
+    # a starved run burns far longer than a healthy one because each failure
+    # drags a 30s timeout plus a retry behind it (the 3-worker run with 4
+    # failures took 12.8m, and main's 53-failure run took 14.4m).
     timeout-minutes: 35
     steps:
       - uses: actions/checkout@v4
@@ -57,8 +60,8 @@ jobs:
       # `#viewer-container` never mounting at all — not as honest assertion
       # failures — so it reads like flake. Measured on this runner:
       # 6 workers => 53 failed / 81 passed (the state main landed in at
-      # 6df8de7), 3 workers => 4 failed / 134 passed. 2 is the next step
-      # down, and matches Playwright's own default heuristic of vCPU/2.
+      # 6df8de7), 3 workers => 4 failed / 134 passed, 2 workers => green in
+      # 11.7m. 2 also matches Playwright's own default heuristic of vCPU/2.
       - name: Run Playwright tests
         run: yarn test-flows --reporter=github,list,html --workers=2
 
--- a/.github/workflows/test-flows.yml
+++ b/.github/workflows/test-flows.yml
@@ -44,8 +44,15 @@ jobs:
       - name: Playwright Info
         run: npx playwright --version
 
+      # Worker count is tied to `runs-on` above, and overrides the local
+      # default in tools/playwright.config.js. Keep it at or below the
+      # runner's vCPU count minus one: each worker drives its own Chromium
+      # alongside the shared static server, and oversubscribing shows up as
+      # timing flake in click/visibility waits rather than as honest
+      # failures. 6 workers on this 4-vCPU runner did exactly that — update
+      # this number whenever the runner size changes.
       - name: Run Playwright tests
-        run: yarn test-flows --reporter=github,list,html --workers=6
+        run: yarn test-flows --reporter=github,list,html --workers=3
 
       # Always upload the HTML report (handy to debug failures)
       - name: Upload Playwright HTML report
--- a/.github/workflows/test-flows.yml
+++ b/.github/workflows/test-flows.yml
@@ -23,7 +23,9 @@ jobs:
     # is chosen so screenshot baselines regenerated in the dev/sandbox env
     # (also noble + chromium 1194) match CI's renders pixel-for-pixel.
     container: mcr.microsoft.com/playwright:v1.56.1-noble
-    timeout-minutes: 20
+    # Headroom for the reduced worker count below: the suite took 12.8m at 3
+    # workers on this runner, so 20 left no margin at 2.
+    timeout-minutes: 35
     steps:
       - uses: actions/checkout@v4
 
@@ -45,14 +47,20 @@ jobs:
         run: npx playwright --version
 
       # Worker count is tied to `runs-on` above, and overrides the local
-      # default in tools/playwright.config.js. Keep it at or below the
-      # runner's vCPU count minus one: each worker drives its own Chromium
-      # alongside the shared static server, and oversubscribing shows up as
-      # timing flake in click/visibility waits rather than as honest
-      # failures. 6 workers on this 4-vCPU runner did exactly that — update
-      # this number whenever the runner size changes.
+      # default in tools/playwright.config.js. Update it whenever the runner
+      # changes size.
+      #
+      # RAM is the binding constraint here, not vCPU: this runner has 8GB
+      # against the previous runner's 32GB, and every worker holds a whole
+      # Chromium rendering a WebGL scene with the conway wasm heap resident.
+      # Starvation shows up as click/visibility timeouts and as
+      # `#viewer-container` never mounting at all — not as honest assertion
+      # failures — so it reads like flake. Measured on this runner:
+      # 6 workers => 53 failed / 81 passed (the state main landed in at
+      # 6df8de7), 3 workers => 4 failed / 134 passed. 2 is the next step
+      # down, and matches Playwright's own default heuristic of vCPU/2.
       - name: Run Playwright tests
-        run: yarn test-flows --reporter=github,list,html --workers=3
+        run: yarn test-flows --reporter=github,list,html --workers=2
 
       # Always upload the HTML report (handy to debug failures)
       - name: Upload Playwright HTML report
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.1722.d41cd27",
+  "version": "1.1743.6caa6b3",
   "main": "src/index.jsx",
   "license": "AGPL-3.0",
   "homepage": "https://github.com/bldrs-ai/Share",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.1741.4a1cba9",
+  "version": "1.1722.d41cd27",
   "main": "src/index.jsx",
   "license": "AGPL-3.0",
   "homepage": "https://github.com/bldrs-ai/Share",
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@babel/plugin-syntax-import-assertions": "7.18.6",
     "@babel/preset-env": "7.18.10",
     "@babel/preset-react": "7.18.6",
-    "@bldrs-ai/conway": "1.460.1363",
+    "@bldrs-ai/conway": "1.465.1375",
     "@bldrs-ai/ifclib": "5.3.3",
     "@emotion/react": "11.10.0",
     "@emotion/styled": "11.10.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "bldrs",
-  "version": "1.1722.d41cd27",
+  "version": "1.1741.4a1cba9",
   "main": "src/index.jsx",
   "license": "AGPL-3.0",
   "homepage": "https://github.com/bldrs-ai/Share",
@@ -75,7 +75,7 @@
     "@babel/plugin-syntax-import-assertions": "7.18.6",
     "@babel/preset-env": "7.18.10",
     "@babel/preset-react": "7.18.6",
-    "@bldrs-ai/conway": "1.450.1353",
+    "@bldrs-ai/conway": "1.460.1363",
     "@bldrs-ai/ifclib": "5.3.3",
     "@emotion/react": "11.10.0",
     "@emotion/styled": "11.10.0",
--- a/src/Components/Camera/CameraControl.jsx
+++ b/src/Components/Camera/CameraControl.jsx
@@ -9,7 +9,6 @@ import {
   removeHashListener,
 } from '../../utils/location'
 import {roundCoord} from '../../utils/math'
-import {floatStrTrim} from '../../utils/strings'
 import {
   HASH_PREFIX_CAMERA,
   removeCameraUrlParams,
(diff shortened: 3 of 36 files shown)


pablo-mayrgundter:
Heads up: `playwright-run` is failing on `main` itself, independently of this PR — worth knowing before reading this branch's red X.

`main`'s tip `6df8de7` ("Change Playwright jobs to use Ubuntu 24.04") swapped the runner `ubuntu-22.04-8vcpu-32gb-300gbssd` → `ubuntu-24.04-4vcpu-8gb-150gbssd` while leaving `--workers=6` in place. That run ([31283693467](https://github.com/bldrs-ai/Share/actions/runs/31283693467)) came back **53 failed / 81 passed**. Every `main` run before it was green.

The failures don't look like real regressions — the dominant mode is `#viewer-container` never mounting at all, i.e. the app not rendering, plus click/visibility timeouts spread across ~20 unrelated specs. That's resource starvation presenting as flake. RAM looks like the binding constraint rather than vCPU: 32GB → 8GB is a 4× cut, and each worker holds a whole Chromium rendering a WebGL scene with the conway wasm heap resident.

Scaling workers down on this branch tracks that read:

| workers | result |
|---|---|
| 6 (`main` @ `6df8de7`) | 53 failed / 81 passed |
| 3 (`295c6b1`) | 4 failed / 134 passed |
| 2 (`559bce3`) | pending |

So `559bce3` here takes it to `--workers=2` (Playwright's own default heuristic is vCPU/2) and raises `timeout-minutes` 20 → 35, since the suite already took 12.8m at 3 workers and 20 left no margin at 2.

Two things worth your call, since neither is really this PR's business:

- This branch carries the workers fix incidentally. If you'd rather land it as a standalone fix directly on `main` — so `main` goes green without waiting on the camera work — I can split it out.
- If 2 workers turns out to still be short, the honest fix is probably going back up in runner size rather than down to 1 worker, which would push the suite past ~25m.

The camera change itself is untouched by any of this.

---

pablo-mayrgundter:
Filling in the pending row from above — **`--workers=2` came back green**, and the whole suite is now passing:

| workers | result |
|---|---|
| 6 (`main` @ `6df8de7`) | 53 failed / 81 passed, 14.4m |
| 3 (`295c6b1`) | 4 failed / 134 passed, 12.8m |
| 2 (`559bce3`) | **0 failed / 138 passed, 11.7m** |

Note 2 workers is also *faster* in wall-clock than 3 — the starved runs were spending their time on 30s timeouts and retries rather than on tests.

That did correct one thing I'd written: I justified raising `timeout-minutes` to 35 on the grounds that 20 "left no margin at 2," and 11.7m fits inside 20 comfortably, so that reasoning was wrong. `ee48290` keeps 35 but states the actual reason — the margin is for the degraded case, not the healthy one, since a starved run costs far more wall-clock than a clean one. Happy to drop it back to 20 if you'd rather not carry the headroom.

Two leftovers:

- **The PR body's "CI workers" section is stale** — it still says "Dropped to 3." I left the body alone rather than risk mangling it on a rewrite; worth a touch-up before squashing if that text is going into the commit message.
- **`main` is still red.** Base is still `6df8de7` and the workers fix lives only on this branch, so anything else branching off `main` right now inherits the 53-failure state. Say the word and I'll split the two workflow commits out onto `main` directly.

---

pablo-mayrgundter:
All three of the "Known, not fixed here" items are now fixed in this PR, one commit each so they can be smoke tested independently.

| commit | fix | smoke test |
|---|---|---|
| `0a5b2e5` | Cut-plane arrow scaled by model radius, no absolute clamp | Open a cut plane on a small STEP part — the arrow should be ~¼ the model, not swallowing it |
| `58adbc0` | Camera permalink rounded relative to scale | Position the camera on a small part, Share, reload the link — camera should land where you left it, not at the origin |
| `4a17538` | Cut-plane offsets no longer quantized to 3 absolute decimals | Set cut planes on a small part, share the `#cp:` link, reload — planes should hold position, not snap to centre |

Each turned out to need more than the one line called out:

- **Arrow** — the `radius * 0.25` ratio was already correct; only the absolute `[2, 40]` clamp was wrong. `MIN` won on a 3.3mm gear and pinned the gizmo at ~600x the model. `MAX` is the same mistake inverted, shrinking the gizmo on large sites. Both dropped; `DEFAULT_ARROW_SCALE` now applies only when there's no usable radius.
- **Camera** — the write-side rounding was only a third of it. `parseHashParams` re-rounded on *read*, undoing whatever precision was written; and `floatPattern` rejected exponential notation, which `String()` starts emitting below `1e-6` — reachable on a sub-millimetre part once rounding is scale-relative, and such a link matched nothing and restored no camera at all. The exponent group is a superset, so existing permalinks still parse.
- **Cut planes** — quantized at three separate hops. Beyond write and read, `getPlaneSceneInfo` rounded a *third* time at scene-placement under the name `finiteOffset`; without fixing that one the planes still land on centre however precise the URL is. It now only coerces to finite, which is all the name claimed.

New `roundCoordComponent` (`src/utils/math.js`) is the shared rule: compute both a fixed-decimal and a significant-digit rounding, keep whichever landed closer to the input. Fixed decimals still win at building scale so URLs stay short and large site coordinates keep their precision; significant digits take over exactly where fixed decimals would have collapsed the value.

**One site beyond your three:** `MarkerControl.jsx:249-250` serializes placemarks through `roundCoord`, so it's fixed by the same change. Worth knowing it was worse than imprecision — on a millimetre part *every* placemark serialized to `0,0,0,0,0,0`, and since that string is the marker's map key, they all collided as one ID. No read-side fix needed; the placemark hash is only ever used as an opaque ID.

Two notes for you:

- **The PR body is now stale in two places** — the "CI workers" section still says "Dropped to 3" (it's 2), and "Known, not fixed here" lists all three of these as unfixed. Both matter if that text becomes the squash message.
- `hashState#getPlanesFromHash` is a near-duplicate of `CutPlaneMenu#getPlanes` that only tests call. I fixed it in step so the two don't drift, but deliberately left the deduplication out of this PR.

---

PR #1745, opened 2026-08-10
Tag open_cid value so GA4 stores it as text, not a number

pablo-mayrgundter:
Fixes `open_cid` landing in GA4's numeric slot, observed in the console as values rendered `1.87152e+09`.

Follow-up to #1743.

## The bug

A GA client id is numeric-looking by construction (`1871520000.1754700000`). We send it as a JS string, but gtag classifies each event param as text or numeric when it builds the collect beacon, and numeric-looking values go out as `epn.` (number) rather than `ep.` (text). Two consequences, both of which the scientific-notation rendering is a symptom of:

1. **Distinct clients collide.** float64 carries ~15–17 significant digits against the id's 20, so `1871520000.1754700000` truncates to `1871520000.17547`. Two clients differing only in the tail collapse to one value — fatal for a metric whose entire purpose is counting events per distinct client.
2. **The dimension stays empty.** An event-scoped custom *dimension* is text and doesn't populate from a numeric param at all.

The same coercion bites downstream of GA4: any CSV/Sheets export of a bare id re-parses it to a float on open.

## The fix

`getOpenCid()` prefixes the id with a non-digit tag before it goes on the event, so the value is `cid.1871520000.1754700000`. A leading non-digit makes it un-parseable as a number, so it takes the `ep.` text path, keeps all 20 digits, and lands in the dimension.

The value becomes opaque rather than joinable against a raw client id, which this metric doesn't need — it only ever counts events per distinct value.

## Changes

- **`src/privacy/analytics.js`**: new `getOpenCid()` returning the tagged value (or null, so callers omit the param rather than send a blank). `getGaClientId()` keeps returning the raw id. The `OPEN_CID_PREFIX` doc records the coercion mechanism and both consequences.
- **`src/Containers/CadView.jsx`**: sends `getOpenCid()` instead of the raw id.
- **`src/privacy/analytics.test.js`**: asserts `Number(value)` is NaN, that the full untruncated id round-trips, that `Number(rawId)` provably differs from the raw id (pinning the collision this guards against), and that the cookie-derived path is tagged too.
- **`src/Containers/realModelOpen.spec.ts`**: E2E now pins the wire format through the real call site — seeds a `_ga` cookie (the one client-id path that works with GA unloaded), then asserts the emitted `open_cid` is the prefixed form and unparseable as a number. Without this, reverting CadView to the raw id kept the whole suite green; verified the new test fails against that revert.
- **`design/roadmap.md`**: `grow-120` records the value format and why, and notes the GA4 definition must be a *dimension*, not a metric.

## Operational notes

- **Values collected before this ships are lossy** — they were truncated at ~16 digits, so that window can't be merged with post-fix data for distinct counts. The dimension is only days old, so the loss is small.
- **Worth confirming the GA4 definition is a dimension, not a metric.** A custom metric would coerce to number regardless of what we send, showing the same scientific notation after this fix. Decisive check either way: load a prod page, filter network for `collect`, and look for `ep.open_cid` (correct) vs `epn.open_cid` (still numeric).

https://claude.ai/code/session_01EULm3yasgtCeQr8cDZaKri

### DIFF
--- a/design/roadmap.md
+++ b/design/roadmap.md
@@ -718,9 +718,15 @@ which channel *reaches* that audience is the open GTM question owned bizdev-side
   originally specced: GA4 reserves the `ga_`/`google_`/`firebase_`/`gtag.`
   prefixes and silently disables matching params, so that name would have
   ingested as nothing and left the dimension permanently empty.
+  The value is sent prefix-tagged (`cid.<client id>`) because gtag types
+  numeric-looking params as numbers (`epn.` in the beacon): a raw id landed
+  in GA4's numeric slot, rendering as `1.87152e+09` — float64 truncates the
+  id's 20 digits, colliding distinct clients, and a text dimension won't
+  populate from a numeric param.
   **Remaining GA4-side config:** Admin → Custom definitions → new
-  event-scoped dimension `open_cid` from event parameter `open_cid`; then
-  query `customEvent:open_cid × eventCount` and bucket client-side
+  event-scoped dimension `open_cid` from event parameter `open_cid` (a
+  *dimension*, not a metric); then query `customEvent:open_cid × eventCount`
+  and bucket client-side
   (1 / 2 / 3–5 / 6+ opens, fixtures excluded). No backfill — the dimension
   accrues from ship time, so land it early in a campaign window for a clean
   baseline. At larger scale GA4 rolls high-cardinality values into "(other)";
--- a/src/Containers/CadView.jsx
+++ b/src/Containers/CadView.jsx
@@ -7,7 +7,7 @@ import {captureException} from '@sentry/react'
 import {fileSuffixBoundaryRegex} from '../Filetype'
 import {useAuth0} from '../Auth0/Auth0Proxy'
 import {onHash} from '../Components/Camera/CameraControl'
-import {OPEN_CID_PARAM, getGaClientId, gtagEvent, isRealModelOpen} from '../privacy/analytics'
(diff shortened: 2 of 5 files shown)


pablo-mayrgundter:
`playwright-run` is red here, but the failure is on the base branch too and predates this PR — not fixing it in this branch.

**What's failing:** 58 tests across ~15 unrelated specs (Notes, Profile, Search, NavTree, Open, ProjectsDrawer, TopBar, Apps, BotChat, GoogleDrive, CutPlane, …), all at the same place — `waitForModel` in `src/tests/e2e/utils.ts:282`, with `#viewer-container` never appearing, or `data-model-ready` never flipping true. Models don't load at all in that environment.

**Why it isn't this PR:** [main's own run](https://github.com/bldrs-ai/Share/actions/runs/31283693467) on `6df8de7` ("Change Playwright jobs to use Ubuntu 24.04") fails with the same signature — same specs, same `waitForModel` assertion, same two flaky `Share.spec.ts` reload tests, same 44 skipped. That commit is the current head of `main` and this branch was cut from it, so it inherits the breakage. The last green `playwright-run` on main was `315dc27` (#1741), before the runner-OS change landed.

This PR's own subject area is unaffected: `build` ✅ and `playwright-webifc-run` ✅ on `afdc203`, and `realModelOpen.spec.ts` passes locally on both form factors (6/6), including the new assertion that `open_cid` is emitted in its prefixed, non-numeric form.

Happy to look at the Ubuntu 24.04 regression as its own change if that's useful — it's blocking every PR, not just this one. I'll re-run here once main is green again.

---

PR #1749, opened 2026-08-13
Add index.step — STEP twin of the logo IFC

pablo-mayrgundter:
Adds `public/index.step`, the STEP twin of `public/index.ifc` (the Bldrs logo), served at `/share/v/p/index.step`. It's the "see it in the viewer" target for the `/viewer/step` SEO landing page (roadmap epic `grow-100`) — landing a STEP search on an IFC model is a bounce.

## Key changes

- **`public/index.step`** — the logo as an AP214 (`AUTOMOTIVE_DESIGN`) assembly, the dialect Conway's STEP path parses. 20KB, next to `index.ifc`'s 18KB. Five levels deep, mirroring the IFC's spatial chain so the NavTree reads as a sentence: Bldrs → Build → Every → Thing → Together ×7.

- **`tools/models/makeIndexStep.mjs`** — the generator. The IFC is the source of truth, so its block extents are transcribed once into a `BLOCKS` constant and everything around them (BREP topology from declarative `CORNERS` / `EDGES` / `FACES` tables, assembly wiring, styling) is emitted rather than typed.

- **`tools/models/README.md`** — how to regenerate and what to check it against.

- **`src/Containers/indexStepLogo.spec.ts`** — E2E on desktop + mobile (`describeMobileAndDesktop`): the file loads through the STEP path, the NavTree reads the sentence above, clicking one block selects exactly that block, and the two logo models occupy the same world space.

- **`design/roadmap.md`** — records that `grow-100`'s `/viewer/step` page now has its demo target.

- **`AGENTS.md`** — router row for `tools/models/README.md`.

## Three things worth knowing

**Millimetre units with ×1000 coordinates.** STEP is millimetre by convention; the IFC logo is in metres. Scaling the numbers keeps both files in the same world-space box, so a `#c:` camera permalink transfers between `/share/v/p/index.ifc` and `/share/v/p/index.step` — which is how the landing page will aim at it.

**Seven blocks, two parts.** The blocks are occurrences of two reusable part shapes (30m and 15m tall) wired with `NEXT_ASSEMBLY_USAGE_OCCURRENCE` + `CONTEXT_DEPENDENT_SHAPE_REPRESENTATION`. That's STEP's native idiom, and it makes this a 20KB regression asset for occurrence-keyed selection (design/new/step-occurrence-selection.md) — today the only assembly fixture is the ~90s NIST `as1`.

**One lime colour per block.** This started as grey-with-a-lime-accent to match what `index.ifc` *says* (`IfcIndexedColourMap`, lime on the -X face). Rendering both side by side showed the IFC comes out entirely lime — nothing displays that per-face grey — so the STEP matches the render. It's also the only option available: Conway resolves AP214 `styled_item`s per representation item, so a per-face colour has nowhere to land.

## The block order in `BLOCKS` is load-bearing (second commit)

Caught on the deploy preview: the two models look identical when each auto-frames, but open `index.step` with the homepage's `#c:` camera and it sits **76m to the right**.

The cause is upstream of both files. Conway's `COORDINATE_TO_ORIGIN` open puts a model's world origin at the **first geometry the file emits** — one coordination matrix derived from that placement × the geometry's first vertex, reused model-wide (`compat/web-ifc/coordination_f64.deriveCoordinationF64`; Share passes the setting in `conwayDirectIfcLoader`). So a model's world position depends on which element its file happens to declare first. `index.ifc` declares the x=76 block first; the generator emitted blocks in ascending x, so its origin landed on the x=0 block — exactly 76m apart. Auto-framing hid it, because it centres whatever it is given.

The fix is to emit the blocks in the IFC's declaration order, which is why `BLOCKS` starts at x=76 rather than the x=0 the eye expects. Sorting it "tidily" reintroduces the bug, so the spec now loads both models and compares world bounds — verified to fail (`min.x drifted: STEP -0.5 vs IFC -76`) when the order is disturbed.

Worth flagging for a follow-up: the ordering coupling is a symptom, not the disease. Under this rule *any* two exports of one object misalign unless their element order matches, and re-exporting a model with a different element order breaks its saved cameras. A deterministic, order-independent coordination frame belongs upstream in Conway. Not attempted here — every candidate fix moves existing models and invalidates saved permalinks.

## Verification

Both models measured in the built app: world bounds agree to ~1e-6m (they were 86 units apart in x before the fix), auto-framed cameras agree, and both render pixel-identically under the reported permalink. Specs pass desktop + mobile; pre-commit gate (eslint + typecheck + 2333 Jest tests) green.

https://claude.ai/code/session_015Vh2vC2y6Q5jgzfTzyRvmA


tools/models/makeIndexStep.mjs
```diff
@@ -0,0 +1,440 @@
...
+ * shows: `AP214ProductStructureExtraction.resolveLabel` prefers the
+ * product name over the occurrence's.
+ *
+ * @param {object} step Emitter
+ * @param {object} contexts Shared `{product, definition}` context refs
+ * @param {string} name Product name
+ * @return {{definition: string, shape: string}} refs
+ */
+function emitProduct(step, contexts, name) {
+  const product = step.add(`PRODUCT('${name}','${name}','',(${contexts.product}))`)
```
pablo-mayrgundter: **Both part products end up with the same `PRODUCT.id`.**

`emitProduct` uses `name` for both attribute 1 (`id`) and attribute 2 (`name`), and both part shapes are named `'Together'`, so the file carries `PRODUCT('Together','Together','',(#3))` twice for two genuinely different parts (30m and 15m).

Concrete failure: a user follows the `/viewer/step` landing page, downloads `index.step`, and opens it in a CAD system or Part 21 validator that keys parts by `product.id` (the AP214 uniqueness convention). The two definitions collide — either a validation error, or the parts merge and the four short blocks come in at the tall part's height.

The fix is free here because Conway's `resolveLabel` reads `formation.of_product.name` (attribute 2), not the id — so a unique id keeps the NavTree label `'Together'` and the spec's `toHaveCount(7)` assertion untouched:

```suggestion
function emitProduct(step, contexts, name, id = name) {
  const product = step.add(`PRODUCT('${id}','${name}','',(${contexts.product}))`)
```
(then pass `` `Together-${height}` `` as the id from the `partsByHeight` construction).


---

tools/models/makeIndexStep.mjs
```diff
@@ -0,0 +1,440 @@
...
+  })
+
+  step.add(
+    `MECHANICAL_DESIGN_GEOMETRIC_PRESENTATION_REPRESENTATION('',` +
+      `(${styledItems.join(',')}),${geometricContext})`)
+
+  return [
+    'ISO-10303-21;',
+    'HEADER;',
+    `FILE_DESCRIPTION(('Bldrs logo — the STEP twin of public/index.ifc'),'2;1');`,
```
pablo-mayrgundter: **Non-ASCII character in a Part 21 string.**

The em dash is U+2014. ISO 10303-21 restricts string contents to the ISO 8859-1 basic set and requires anything outside it to be escaped as `\X2\2014\X0\`. `grep -P '[^\x00-\x7F]' public/index.step` confirms line 3 is the only non-ASCII line in the emitted file, and it is emitted as raw UTF-8.

Concrete failure: an importer that decodes the file as ISO 8859-1 (the Part 21 default) renders the header as `Bldrs logo â€" the STEP twin…`, and a strict Part 21 syntax checker rejects the string. Since the whole point of this asset is that STEP users download it and open it elsewhere, the header is the first thing they see.

```suggestion
    `FILE_DESCRIPTION(('Bldrs logo - the STEP twin of public/index.ifc'),'2;1');`,
```


---

tools/models/makeIndexStep.mjs
```diff
@@ -0,0 +1,440 @@
...
+
+/** Lime, from `public/index.ifc`'s `IfcColourRgb` / `theme.palette.logo.leftFace`. */
+const LIME = [0, 1, 0]
+
+/**
+ * Box corners in the part's local frame, indexed 0-7: 0-3 walk the
+ * z=0 face counter-clockwise from the origin, 4-7 the z=depth face
+ * directly above them. Edge and face tables below index into this.
+ */
+const CORNERS = [
```
pablo-mayrgundter, line 117: **`CORNERS[0]` (and the `FACES` order) is as load-bearing as `BLOCKS`, but nothing says so.**

The alignment invariant documented on `BLOCKS` and in the README is only half of it. `deriveCoordinationF64(placement, point, …)` anchors the model at `placement × point`, and the caller (`ifc_api_proxy_ap214.js`) passes `nativePt = geometry.geometry.getPoint(0)` — the *first tessellated vertex* of the first placed geometry. Today that is local `(0,0,0)` only because `CORNERS[0]` is the min corner and `FACES[0]`'s loop starts there.

Concrete failure: a maintainer tidies `CORNERS` to start at `[1,1,1]`, or reorders `FACES` so the +Z face is tessellated first. The anchor moves to the max corner, `index.step` slides (10, 11.45, height) m relative to `index.ifc`, and every cross-format `#c:` permalink is wrong. `indexStepLogo.spec.ts` fails — but both comments point the reader at `BLOCKS`, which is unchanged and innocent, so the debugging starts in the wrong place.

Worth one sentence here: "the tessellator's first vertex is the coordination anchor, so `CORNERS[0]` must stay the minimum corner and `FACES[0]` must keep starting there — see the `BLOCKS` note."


---

tools/models/makeIndexStep.mjs
```diff
@@ -0,0 +1,440 @@
...
+
+
+/**
+ * Formats a JS number as a STEP real, which — unlike JSON — must carry
+ * a decimal point even when integral.
+ *
+ * @param {number} value
+ * @return {string} STEP real literal
+ */
+function real(value) {
```
pablo-mayrgundter, line 166: **`real()` silently emits illegal Part 21 for anything JS formats exponentially.**

Two holes:
- `Number.isInteger(1e21)` is `true`, so the integer branch produces `1e+21.`
- a magnitude below `1e-6` stringifies as `1e-7` — no decimal point, lowercase exponent

Neither is a valid Part 21 REAL (decimal point mandatory, exponent must be uppercase `E`).

Nothing in the current `BLOCKS` triggers it, but `BLOCKS` is explicitly documented as a hand-transcription that gets re-done when the logo changes. Transcribe a coordinate that lands outside JS's decimal window and the generator writes an unparseable file with no error — the only symptom is the viewer failing to load `index.step`. Cheap guard:

```js
const text = normalized.toString()
if (text.includes('e')) {
  const [mantissa, exponent] = text.split('e')
  return `${mantissa.includes('.') ? mantissa : `${mantissa}.`}E${exponent.replace('+', '')}`
}
```


---

tools/models/makeIndexStep.mjs
```diff
@@ -0,0 +1,440 @@
...
+    'DATA;',
+    step.toString(),
+    'ENDSEC;',
+    'END-ISO-10303-21;',
+    '',
+  ].join('\n')
+}
+
+
+const outPath = process.argv[2] ?? path.join(process.cwd(), 'public', 'index.step')
```
pablo-mayrgundter: **Default output path is cwd-relative, so the documented command only works from the repo root.**

`cd tools/models && node makeIndexStep.mjs` — the natural thing to do while editing the generator — throws `ENOENT` because `tools/models/public/` doesn't exist. From any other directory that happens to have a `public/`, it writes the wrong file and the committed model is never updated, with a cheerful `Wrote …` on stdout either way.

Resolving against the script rather than the caller makes the README's command work from anywhere:

```suggestion
const defaultOut = path.resolve(path.dirname(url.fileURLToPath(import.meta.url)), '..', '..', 'public', 'index.step')
const outPath = process.argv[2] ?? defaultOut
```
(with `import url from 'node:url'` at the top).


---

src/Containers/indexStepLogo.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * World-space bounds of everything rendered in the scene, in metres.
+ * BatchedMesh/InstancedMesh need `computeBoundingBox()` on the mesh —
+ * `geometry.boundingBox` is the un-instanced prototype shape, which for
+ * the STEP model is one block rather than the whole logo.
+ *
+ * @param page the Playwright page, on a loaded model
+ * @return `{min, max}` as `[x, y, z]` metre triples
+ */
+function sceneBounds(page: import('@playwright/test').Page) {
```
pablo-mayrgundter: **This measures the whole scene; the sibling spec documents why it should measure the loaded models.**

`src/Components/Camera/smallPartNearPlane.spec.ts` hits exactly this and says so in a comment: *"Measure the loaded models, not the scene: the `?feature=look` ground plane is a real mesh sized to a multiple of the model, so a whole-scene walk reads several times too large."* It reads `context.getLoadedModels()` instead.

Concrete failure: run this spec against a build or URL with `?feature=look` on (or once the look ships default-on — viewer-replacement.md §6e). `ShareViewer.groundModel` then adds a `contactShadowGround` mesh scaled to `GROUND_SIZE_FACTOR × maxXZ` of `robustBoundsFor(model)`. Both scenes get one, but any difference in robust bounds between the IFC's and the STEP's mesh partitioning is multiplied by `GROUND_SIZE_FACTOR` and blows past `ALIGNMENT_TOLERANCE_M` — the test fails for a reason with nothing to do with alignment, in the flag configuration where it is hardest to debug.

Switching to `context.getLoadedModels()` also drops the `isBatchedMesh`/`isInstancedMesh` special-casing, since `Box3.expandByObject` handles both.


---

src/Containers/indexStepLogo.spec.ts
```diff
@@ -0,0 +1,147 @@
...
+        min[i] = Math.min(min[i], box.min[axis])
+        max[i] = Math.max(max[i], box.max[axis])
+      }
+    })
+    return {min, max}
+  })
+}
+
+describeMobileAndDesktop('index.step logo', () => {
+  test('loads and shows the logo assembly, one selectable node per block', async ({page}) => {
```
pablo-mayrgundter, line 146: **Nothing here asserts the colour the docstring says this spec pins.**

The header above claims the spec pins *"a styling chain Conway doesn't walk still 'loads' — it renders as nothing, or as default grey"*, but neither test reads a material or a colour. The entire `emitSolidColour` chain (`STYLED_ITEM → PRESENTATION_STYLE_ASSIGNMENT → SURFACE_STYLE_USAGE → SURFACE_SIDE_STYLE → SURFACE_STYLE_FILL_AREA → FILL_AREA_STYLE → FILL_AREA_STYLE_COLOUR → COLOUR_RGB`) is untested.

Concrete failure: drop `SURFACE_STYLE_FILL_AREA` — or any other link — from `emitSolidColour` and regenerate. `AP214GeometryExtraction.extractSurfaceStyle` falls through to its `0.8` grey default, the logo renders grey instead of lime, and both tests still pass. The PR spends a whole section on getting this colour right, and AGENTS.md's standing rule is "if it's not tested, it doesn't work".

One `page.evaluate` reading a mesh material's colour off `window.store` (the same handle `sceneBounds` already uses) closes it.


---

pablo-mayrgundter (commented):
Careful pass over the whole diff. No showstoppers — the BREP topology checks out (I verified all six face loops wind CCW about their plane normals in each plane's own parameter space, and the plane `ref` directions are perpendicular to their axes), the styling chain matches what `AP214GeometryExtraction.extractSurfaceStyle` actually walks, `resolveLabel` does prefer the product name so the NavTree sentence holds, and the `BLOCKS` extents match `index.ifc`'s point lists block-for-block including declaration order. `node tools/models/makeIndexStep.mjs` reproduces the committed `public/index.step` byte-identically, and eslint + `yarn typecheck` are clean on the new files.

7 findings below, none blocking: two interop issues in the emitted STEP (duplicate `PRODUCT.id`, raw UTF-8 in a Part 21 string), an under-documented half of the coordination invariant, a latent hole in `real()`, a cwd-relative output path, and two spec notes (whole-scene bounds vs. loaded models; the colour the docstring claims to pin isn't asserted).


---

src/Containers/indexStepLogo.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+    // three hands back.
+    const colours = await modelColours(page)
+    expect(colours.length).toBeGreaterThan(0)
+
+    const isLime = ([r, g, b]: number[]) =>
+      g > CHANNEL_ON_MIN && r < CHANNEL_OFF_MAX && b < CHANNEL_OFF_MAX
+    expect(
+      colours.some(isLime),
+      `no lime material; got ${JSON.stringify(colours)}`,
+    ).toBe(true)
```
pablo-mayrgundter: **`.some()` is weaker than the contract this test states.** A single lime instance satisfies it, so it can't distinguish an all-lime model from a partly-unstyled one.

`generate()` emits one `STYLED_ITEM` per part shape (15m and 30m). Drop the second — move `styledItems.push(...)` out of the `heights.map` loop, or list only `styledItems[0]` in the `MECHANICAL_DESIGN_GEOMETRIC_PRESENTATION_REPRESENTATION` — and regenerate: the three 30m blocks fall back to Conway's 0.8 grey. `productPalette.repaintByProduct` won't rescue them either, since it no-ops the moment any real colour is present and the four 15m blocks are still lime. Half the logo renders grey and this test stays green, against the docstring's own "It renders lime".

`.every()` is the obvious fix but gets brittle if Share ever parents a helper mesh under the model. Counting is tighter and names the expected shape:

```suggestion
    const isLime = ([r, g, b]: number[]) =>
      g > CHANNEL_ON_MIN && r < CHANNEL_OFF_MAX && b < CHANNEL_OFF_MAX
    expect(
      colours.filter(isLime).length,
      `expected ${BLOCK_COUNT} lime blocks; got ${JSON.stringify(colours)}`,
    ).toBe(BLOCK_COUNT)
```


---

tools/models/makeIndexStep.mjs
```diff
@@ -0,0 +1,486 @@
...
+
+// Resolved against the script, not the caller: a cwd-relative default
+// makes `cd tools/models && node makeIndexStep.mjs` throw ENOENT, and
+// from any other directory that happens to have a `public/` it writes
+// the wrong file while still reporting success.
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..')
+const outPath = process.argv[2] ?? path.join(repoRoot, 'public', 'index.step')
+fs.writeFileSync(outPath, generate())
+// eslint-disable-next-line no-console
+console.log(`Wrote ${outPath}`)
```
pablo-mayrgundter: **Nothing pins the committed `public/index.step` to this generator.** The README calls the generator the source you edit and the `.step` a build product, but no test regenerates and compares, so the two can silently diverge.

Concrete path: someone sorts `BLOCKS` ascending by x — the exact tidy-up the L80 and L110 comments warn about — and doesn't rerun the script. eslint, Jest and `indexStepLogo.spec.ts` are all green, because the spec loads the stale committed bytes rather than the generator's output; the alignment test both comments point at as the safety net never sees the changed code. The regression lands, and surfaces at the next unrelated regeneration with the causing commit long out of view.

Output is fully deterministic (fixed `FILE_NAME` timestamp), so a byte-compare is a five-line Jest test — except the module has no exports and runs `generate()` + `writeFileSync` at import, so a test can't import it without writing into `public/`. Exporting `generate` and guarding the write behind a main check makes `expect(generate()).toBe(fs.readFileSync('public/index.step', 'utf8'))` possible:

```suggestion
// Resolved against the script, not the caller: a cwd-relative default
// makes `cd tools/models && node makeIndexStep.mjs` throw ENOENT, and
// from any other directory that happens to have a `public/` it writes
// the wrong file while still reporting success.
const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '..', '..')

export {generate}

// Guarded so a drift test (`generate()` vs the committed bytes) can import
// this module without writing into `public/`.
if (process.argv[1] && path.resolve(process.argv[1]) === fileURLToPath(import.meta.url)) {
  const outPath = process.argv[2] ?? path.join(repoRoot, 'public', 'index.step')
  fs.writeFileSync(outPath, generate())
  // eslint-disable-next-line no-console
  console.log(`Wrote ${outPath}`)
}
```


---

pablo-mayrgundter (commented):
Reviewed the full diff (generator, generated STEP, E2E spec, docs). No correctness bugs found in the emitted model — I verified it independently:

- **Reference integrity:** 450 entities, no duplicate ids, no dangling `#N`, and the 24 unreferenced entities are all legitimate Part 21 roots (`APPLICATION_PROTOCOL_DEFINITION`, `PRODUCT_RELATED_PRODUCT_CATEGORY`, `SHAPE_DEFINITION_REPRESENTATION`, `CONTEXT_DEPENDENT_SHAPE_REPRESENTATION`, the MDGPR).
- **BREP orientation:** all 6 `FACES` loops produce the declared outward normal under the right-hand rule, each plane's `ref` is perpendicular to its axis, and every `origin` lies on its own plane. `ORIENTED_EDGE` sense flags check out against the `EDGES` table.
- **`BLOCKS` vs the IFC:** all seven entries match `public/index.ifc`'s `IFCCARTESIANPOINTLIST3D`s in both value and declaration order (worst disagreement 1e-10 m, from `BLOCK_DEPTH` rounding). Bounds agree: x 0..86, y -24.0980..12.4238, z 0..30.
- **Reproducibility:** re-running the generator reproduces `public/index.step` byte-for-byte.
- **The `resolveLabel` assumption holds** — `ap214_product_structure_extraction.resolveLabel` does prefer the product name, so the seven NAUOs named `'1'`..`'7'` will read as `Together` in the NavTree.
- `productPalette`'s palette contains no colour passing the spec's `isLime` bounds, so that assertion isn't vacuous, and the ground plane is added to the scene rather than to a model, so `getLoadedModels()` correctly excludes it.
- eslint clean on both new files.

Two test-quality findings inline; neither blocks correctness of the model itself.


---

PR #1752, opened 2026-08-15
Bump GLB schema to 0.16.0 so cached models pick up the new coordination frame

pablo-mayrgundter:
Follow-up to #1749. Reported from a real homepage load: **the logo rendered offset, and clearing the cache fixed it.**

## Why

#1749 shipped conway `1.501.1426`, which changed where models land — inside the precision budget a model now keeps the coordinates its file authored (conway#501 / conway#87). `index.ifc` moved by `(+76, 0, +11.4504)`, and Share's cameras moved with it, including the homepage default in `utils/navigate.js`.

But **geometry is baked into the cached GLB in world coordinates, and nothing in the cache key identifies the engine.** So anyone holding a pre-`1.501` artifact keeps rendering the model at its old position while the new cameras aim at the new one. The model looks offset and stays that way until the cache is cleared — which is precisely the report.

## The fix

Bump `BLDRS_GLB_SCHEMA_VERSION` to `0.16.0`. The version is part of the artifact filename, so `index.0.15.0.glb` → `index.0.16.0.glb`: every pre-fix artifact reads as a miss and the next load re-bakes it in the new frame.

This is the same failure mode as the `0.15.0` entry directly below it in the version log — an engine change that moves baked world coordinates — and that entry's own note about pairing an engine bump with a schema bump is the rule I should have followed when landing the conway bump in #1749. The new log entry records the mechanism, the measured offset, and why it only shows up for cache-hit users.

## Verification, and its limit

- Full Jest suite green (2347). The schema-version tests read the constant rather than hardcoding it, so they follow the bump.
- Mechanism checked directly: `glbArtifactPath('index.ifc')` returns `index.0.16.0.glb` against `index.0.15.0.glb` for the old version, so the invalidation is real.

**No E2E covers this**, and that's worth flagging rather than hiding: `SHARE_CONFIG=playwright` sets `OPFS_IS_ENABLED=false`, so no spec in the suite exercises a cache hit at all (`Properties.cacheHit.spec.ts` is already `test.fixme` for related reasons). I tried to reproduce the round-trip under Playwright and got two identical passes with no `[glb] cache HIT/MISS` lines, because the cache was never engaged. That gap is why this bug reached a user: it only exists for cache-hit users, and local testing is all cache-miss.

https://claude.ai/code/session_015Vh2vC2y6Q5jgzfTzyRvmA

---
_Generated by [Claude Code](https://claude.ai/code/session_015Vh2vC2y6Q5jgzfTzyRvmA)_

### DIFF
--- a/src/loader/glbCacheKey.js
+++ b/src/loader/glbCacheKey.js
@@ -15,6 +15,25 @@
 /**
  * Current Bldrs GLB artifact schema version. Bumped on any backwards-
  * incompatible change to the BLDRS_* extension contract or cache-key shape.
+ * 0.16.0 — retires artifacts baked in the old coordination frame. conway
+ *         1.501.1426 (bldrs-ai/conway#501, issue conway#87) stopped anchoring
+ *         `COORDINATE_TO_ORIGIN` on whichever element a file happens to
+ *         declare first: a model inside the precision budget now keeps the
+ *         coordinates its file authored. index.ifc moved by (+76, 0,
+ *         +11.4504) — x ∈ [-76, 10] became x ∈ [0, 86] — and every other
+ *         near-origin model moved by its own old recentre.
+ *         Geometry is baked into the GLB in world coordinates and nothing in
+ *         the cache key identifies the engine, so without this bump a user
+ *         holding a pre-1.501 artifact keeps rendering the model at its old
+ *         position while the app's cameras — the homepage default in
+ *         `utils/navigate.js`, and every `#c:` permalink captured after the
+ *         bump — point at the new one. Reported as an offset homepage logo
+ *         that a cache clear fixed, which is exactly the shape of this bug:
+ *         cache-hit users see it, cache-miss users don't, so it survives
+ *         local testing.
+ *         Same failure mode and same remedy as 0.15.0 below; that entry's
+ *         note about pairing an engine bump with a schema bump is the rule
+ *         this follows.
  * 0.15.0 — retires STEP artifacts baked at the wrong world scale. conway
  *         1.460.1363 (bldrs-ai/conway#458, PR conway#460) stopped applying a
  *         STEP file's length-unit factor as its RECIPROCAL, so a millimetre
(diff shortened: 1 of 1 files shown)


pablo-mayrgundter:
The three failing checks (`Pages changed`, `Header rules`, `Redirect rules` — all facets of one deploy) are the **`bldrs-share-dev` Netlify deploy**. I don't think this is the diff, and here's what I checked rather than just asserting it:

- **The same commit deployed fine to `bldrs-share-prod`.** `96bd951` produced a working preview there while dev failed. Both projects build from the same `netlify.toml` (`yarn build`), so an identical build of an identical tree succeeded on one and failed on the other.
- **`yarn build-prod` passes locally** on this branch, exit 0.
- **The diff is one string constant plus comments** — `BLDRS_GLB_SCHEMA_VERSION` `'0.15.0'` → `'0.16.0'`. There's no plausible mechanism by which that breaks a site build, let alone only one of two.

`netlify.toml` also documents this exact shape of failure: *"previews and production are two separate Netlify projects, and only one of them carried a NODE_VERSION in its UI environment"* — which is why that got pinned in-repo. Per-project environment drift is the standing precedent for dev-green/prod-red (or the reverse) on one commit.

What I could not do: read the deploy log. `app.netlify.com` is blocked from this sandbox (proxy returns 403 on CONNECT), and the check run itself carries no output beyond a link to it. So the actual error message is unread, and I'm reasoning from the asymmetry rather than from the log.

Someone with Netlify access should open [the deploy log](https://app.netlify.com/projects/bldrs-share-dev/deploys/6a8056d5e56275000820af92?deploy_diagnostics=true) to confirm. If it turns out to be the diff after all, I'll fix it — but I'd rather flag an unread log than quietly retry the deploy until it goes green.

---

pablo-mayrgundter:
Resolved on its own: `bldrs-share-dev` re-ran and went green on the **identical commit** (`96bd951`, deploy `6a80585d`), no code change in between. Transient, as the prod-succeeded/dev-failed asymmetry suggested.

The log is still unread, so the root cause is unconfirmed — worth a look only if dev starts failing repeatedly.

---

Issue #1640, opened 2026-07-24
Conway: demand pump delivers aggregate part geometry empty/truncated (ILNA facades missing, −20–34% triangles vs classic)


pablo-mayrgundter:
Follow-up to the #1632 retrospective, split out of the #1627 discussion. Work lands in `bldrs-ai/conway`; tracked here from Share's side.

## Symptom

`test-models-private` vyzn/`ILNA 3D_SIA2040_optimized.ifc` (98 MB) renders with large facade sections missing on prod and on the #1627 preview — visible as exposed structural framing where the old (pre-demand) render showed complete clad walls. Old deploy-preview-1572 (classic pipeline era) looks correct.

## Measured (per-product placement + triangle sums, node)

| conway | path | placements | triangles |
|---|---|---|---|
| 1.385.1233 | classic | 37,667 | 4,096,285 |
| 1.432.1305 (prod) | classic | 37,667 | 4,096,277 |
| 1.432.1305 (prod) | **demand** | **28,637** | **2,704,552 (−34%)** |
| 1.436.1311 (#1627) | classic | 37,667 | 4,096,277 |
| 1.436.1311 (#1627) | **demand** | 37,667 | **3,287,759 (−20%)** |

- **The classic path is correct and stable at every version** — this is not a version regression; it is a demand-pump extraction bug.
- On 1.432, 8,305 products are affected — **8,226 `IfcBuildingElementPart` + 78 `IfcColumn` + the `IfcSite`** — each losing a placement (2→1) *and* its geometry content (e.g. `#471607`: 21,238 → 98 tris). The parts are exactly the facade cladding (vyzn decomposes walls into parts via rel-aggregates).
- On 1.436, the dropped placements are restored (2→2 — the visible improvement in the #1627 preview) but **2,047 products** (1,969 `IfcBuildingElementPart` + 78 `IfcColumn`) still lose their content (21,238 → 196 tris).

## Mechanism probe (1.436, same build, same model)

Reading each affected product's placements *at delivery time*:

```
CLASSIC #471607: gid=1324977 tris=10619 | gid=1324977 tris=10619
DEMAND  #471607: gid=1324977 tris=0     | gid=1324977 tris=0      ← content empty at delivery
```

[...]

Issue #1602, opened 2026-07-17
Streaming loader: consume openStream / delete post-parse OPFS spill (↔ conway M1)


pablo-mayrgundter:
Part of #1601. Pairs with [conway M1](https://github.com/bldrs-ai/conway/issues/392). Design: [`streaming-federated-loader.md`](https://github.com/bldrs-ai/conway/blob/main/design/new/streaming-federated-loader.md).

## Goal

Move Share's IFC open path onto conway's new `openStream` engine API, where OPFS is the byte store from t=0 — and delete the post-parse `spillModelSource` dance in `Loader.js` (it becomes redundant: the source is never fully resident to spill).

## Work

- Switch the loader to `openStream(source)` when the conway build exposes it (feature-detected, like the roots sweep); keep the current `OpenModel` path as fallback.
- Remove the `spillModelSource(...)` call from the GLB writer's `.finally` once the stream path is active (source is disk-backed from the start).
- Verify the props capture + names tree + GLB writer still run against the streamed model (they consume the same passthrough/index surface).

## Exit criteria

- [ ] PSB opens via `openStream`; no full-source-resident moment (task-manager / heap-snapshot check)
- [ ] `opfsSourceByteStore` spill path retired or reduced to the fallback branch
- [ ] Props panel + spatial tree + GLB cache write unchanged in output
- [ ] Bump `@bldrs-ai/conway` to the M1 release; tarball-verify `openStream` before bumping

Blocked by: conway M1 (#392) release.

Issue #593, opened 2023-02-11
Viewer: Change the color of the cut planes to match the theme.


OlegMoshkovich:
<img width="879" alt="image" src="https://user-images.githubusercontent.com/3433606/218262732-ad5d3070-6ce8-4df1-9c5a-3441f4e0e309.png">

<img width="594" alt="image" src="https://user-images.githubusercontent.com/3433606/218262606-06bc953c-f225-4524-b9ed-eeaa1fa45028.png">
set the color of the cutplanes to grey light
<img width="247" alt="image" src="https://user-images.githubusercontent.com/3433606/218262656-3a1f8c7f-eb6a-44cd-be26-4411ef3a3d16.png">

PR #1753, opened 2026-08-15
M3: fixed-memory streaming load with parse-time preview geometry (conway 1.501 → 1.588.1550)

pablo-mayrgundter:
The Share half of conway's M-milestone loading work: **fast, fixed-memory model loading with streaming preview geometry, for IFC and STEP**. Bumps `@bldrs-ai/conway` **1.501.1426 → 1.588.1550** and carries every Share-side half of the engine contract changes that range introduced.

14 files, all on the conway-direct load path. There is no new feature flag — this rides the existing deferred/store path.

> **Two rows in the ledger below have a minor number that goes backwards. That is correct — do not "fix" them.**
> conway's `auto-publish` derives the minor from the first `#N` in the squash
> title, so #581 (which closes issue #564) published as 1.**564**.1548, and
> #531 published as 1.**394**.1504. The **build number is the monotonic
> part**. Root cause: [conway#533](https://github.com/bldrs-ai/conway/issues/533).
>
> **The current pin is not one of them.** 1.588.1550 is monotonic in both
> parts against 1.580.1549, and `npm dist-tags.latest` resolves to it
> without auto-publish's out-of-order fallback.

---

### Opening a model without copying it

| | before | now |
|---|---|---|
| OPFS cache-hit IFC | full `ArrayBuffer` materialised (848 MB on PSB) | opened from the `File` in place via `OpenModelStream` |
| fingerprint | needed the whole buffer resident | native per-slice SHA-1, one reused scratch buffer |
| LFS probe | only when `opfsEntryKey` was set | always, before any `File` passthrough |

STEP falls back to `OpenModelStreamed`; an engine that refuses store-open (returns `-1`) buffers and retries rather than failing the load. The IFC is no longer `arrayBuffer()`'d twice — the parse buffer is hashed directly.

### Holding memory flat

- **Windowed source.** The demand pump pages the source through a ~64 MB window; the Geometry line reports `window=N MB`.
- **Resident-geometry budget** (`accc55a`). `GEOMETRY_BUDGET_MB = 64` on the deferred path, turning on conway#535's LRU eviction. Engine-measured on PSB at our batch size of 8: **wasm high-water 1284 → 298 MB**, delivered meshes identical, Geometry slightly faster (23.6 s vs 25.7 s). 256 MB was also measured and buys much less (803 MB) — the live set rarely reaches that ceiling.

  Safe for one specific reason worth stating: eviction frees an asset from `GetGeometry` until something re-extracts it, and this loader copies every payload at delivery (the #1640 invariant). That invariant used to be merely true; it is now load-bearing, and the test asserts both halves — the classic path stays unbudgeted, because it does not pump and would have geometry evicted from under a later consumer.
- **Preview teardown disposes pooled resources exactly once** (`c1812b3`, `09b943d`). The dispose pass had been dropped wholesale when the shared imposter cube arrived, rather than made shared-aware. Teardown now walks stray imposters plus the preview group and disposes each unique geometry/material once.
- **The rel-aggregates pass no longer retains the whole model** (conway#550). Reading `relAggregate.RelatedObjects` memoised the materialised entity array onto the relationship, so every related product — and the tessellation subtree each one memoises — stayed reachable for the whole loop, and the per-product B-rep graphs summed instead of overlapping. The pass also had no suspension point, so on a model whose products are *all* aggregate targets the cooperative driver yielded nothing during 100% of the real work. This is what made SKYLARK250 hang the browser at an earlier pin.
- **And the prefetch that fed it no longer pages a whole relationship ahead of the stepper** (conway#563 + conway#566). #550 made the *pass* step per product and left the *prefetch* per relationship, so on SKYLARK250 — where two `IfcRelAggregates` cover all 1,992 products — windowed residency was still defeated for exactly the model it exists for. Together the two take parse→first-mesh from **315,903 ms → 266 ms** and whole-load peak resident source from **381.7 → 61.7 MiB**.
- **AP214 demand-unit slicing** (conway#582). Arty_Z7's worst single main-thread block **17,731 ms → 600 ms**. This fixes the *freeze*, not the *cost* — the work still happens, it is just sliced into units the event loop can interleave with, so the tab stays responsive through it.

### First pixels during the parse

Real placed prefix meshes stream while the file is still parsing, through the same 64 MB window, extraction time-budgeted at ~20 ms/tick (conway#519) so the preview fills in batches rather than one girder at a time.

The engine-side reason this works at all is [conway#542](https://github.com/bldrs-ai/conway/issues/542), whose fix landed in [conway#543](https://github.com/bldrs-ai/conway/pull/543). Its finding reframes what "slow first load" even means: whether a model shows anything during parse is a property of the **writer**, not the authoring tool or the file size.

| writer | forward refs | products deferring | first preview mesh |
|---|---|---|---|
| ODA (PSB, 861 MB) | 0.1% | 0.5% | 269 ms |
| Tekla (D3D, 224 MB) | 0.0% | 0.0% | streams cleanly |
| Archicad/DDS_IFC (DOWA, 398 MB) | 19.8% | **100%** | 8891 ms → **5849 ms** |
| Archicad/DDS_IFC (MB-Khaya, 31 MB) | 74.1% | **100%** | 979 ms → 758 ms |
| ST-DEVELOPER (BLSN, 282 MB STEP) | 25.7% | **100%** | no prefix trick helps |

PSB — the model we have been smoking all along — is the *best* case in the corpus, which is why the blank-screen problem kept not reproducing. Numbers are from conway's Node harness on the dev corpus, not from a browser. conway#543's review found two defects in the script that produced them, both flattering, so treat the exact percentages as approximate pending a re-run; the ordering and the conclusion are unaffected.

**Time-to-first-pixel is measured but not yet visible here.** conway#543 adds a `Preview:` line. Share cannot display it: `applyEngineLogLevel()` mutes conway to `WARNING` by default, `loadProgress.js`'s console tee captures only `warn`/`error`, and `setPreviewStats()` has no caller. Tracked as [conway#544](https://github.com/bldrs-ai/conway/issues/544). So first pixels on this smoke are still judged by eye. Known gap, not an oversight.

### Spatial AABB plates are OFF

`SHOW_AABB_IMPOSTERS = false` (`e9896c4`). The plates cost three conway rounds — scale/solid ([#516](https://github.com/bldrs-ai/conway/pull/516)), rotation ([#517](https://github.com/bldrs-ai/conway/issues/517)), banding and early skeleton ([#518](https://github.com/bldrs-ai/conway/issues/518)) — and they are cosmetic: real prefix meshes carry first pixels. Turned off at the Share end rather than by asking conway to stop emitting, so the engine contract is unchanged and re-enabling is one constant.

The consumer contract stays in place and tested: an `aabb` payload re-emitted with an `expressID` already drawn **replaces** that plate rather than adding a second one — no extra cap slot, cannot evict a visible plate, nothing disposed on replace, and a plate rejected by the outlier guard leaves the standing plate on screen, because a refused update must not delete scenery it failed to refine.

### Observability

Geometry is its own load-log stage during the demand pump, with heap delta and `window=`.

---

## Conway version ledger

Each bump is a contract change with a Share-side half in this PR.

| pin | conway | what it brought here |
|---|---|---|
| 1.510.1439 | #510 | `OpenModelStream` store-backed open |
| 1.513.1447 | #513 | rematerialise extract buffers; parallel `#ref` prefetch |
| 1.514.1452 | #514 | `Faces[]` as one address span; storey plates; store-path prefix extract during parse |
| 1.516.1463 | #516 | imposter coordination-frame fix → Share drops the first-box re-anchor, threads `coordination` through the aabb path |
| 1.519.1471 | #519 | rotation-aware plates (#517); time-budgeted preview + early skeleton (#518) → Share keys plates by `expressID` |
| 1.529.1492 | #529 | concurrent-prefetch residency fix (#526) — the `StepBufferNotResidentError` that aborted a large-model load outright |
| 1.530.1503 | #530 | M2 prefix index: type index from columns; spatial-names skeleton during parse; `LongName` fix |
| 1.394.1504 | #531 | scripts/design only — engine-identical to 1.530.1503; the number moved *backwards* (see the note at the top) |
| 1.534.1505 | #534 | incremental delta capture — no more O(batches × scene) re-walk; PSB Geometry 37.0 → 25.3 s at batch 8 |
| 1.535.1506 | #535 | resident-geometry LRU budget + native-identity refcounting |
| 1.536.1507 | #536 | dispatch-time placement + `SetGeometryShard` — **inert on this PR's path**; gated behind a claimed shard, which Share never claims. Consumed by #1756 instead. |
| 1.543.1513 | #538, #543 | #538: residency-independent dispatch keys + `SetCoordinationFrame`. **#543**: deferred preview products are retried against the longer index instead of abandoned, on **both** the store and resident paths, plus time-to-first-pixel instrumentation. |
| 1.548.1514 | #548 | harness only — RFC 4180 quoting for the benchmark CSVs and a delta emitted when blessing a release. |
| 1.549.1515 | #550 | The rel-aggregates pass stops retaining the model and gains a suspension point. SKYLARK250: peak live JS heap **3784 → 982 MB**, peak RSS **5314 → 3152 MB**, event-loop heartbeat from **1 firing in 104 s (max stall 100,877 ms)** to 597 firings at 11.98/s. Digests byte-identical. |
| 1.560.1539 | #553, #556, #557/#559, #558/#560 | **Measurement and CI only — no engine change reaches Share.** |
| 1.578.1546 | #563, #566, #567, #568, #569, #570, #578 | **Four of seven are runtime for Share.** **#563** fixes the style-seed scan on the windowed path: SKYLARK250 parse→first-mesh **315,903 → 41,658 ms**. **#566** finishes it by paging an aggregate's products per stepper step: **35,758 → 266 ms**, whole-load peak resident source **381.7 → 61.7 MiB**, longest event-loop starvation **34,866 → 45 ms**. **#568 §4** stops a STEP complex instance being extracted by both passes and drawn twice. **#569** counts the Geometry denominator in products rather than relationships. **#567** and **#578** move `conway-geom`. |
| 1.564.1548 | #581, #582 (+ conway-geom#183) | **The STEP/AP214 round.** **#582** slices AP214 demand units: Arty_Z7's worst main-thread block **17,731 → 600 ms**. **#581 + conway-geom#183** scope the deflection floor to the **defining representation**: Arty geometry **33,981 → 20,265 ms**, payload **403.878 → 138.366 MB**, peak wasm heap **707.0 → 357.5 MB**. Composed on one box, Arty geometry **34,169–37,235 → 22,484–24,985 ms**, payload **−66%**, heap **−49%**. |
| **1.588.1550** | #586, #588 | **The current pin — and the first head in this PR's life whose engine has blessed baselines behind it.** Neither changes geometry. **#586** gates the per-item error log on prefixes and stops `DanglingReferenceError` reporting recoverable prefix misses as errors — console noise only. **#588** fixes the **STEP NavTree dying with `Error: Function not implemented.` on any product with no name**: `resolveLabel()` ended on `productDef.name ?? ''`, but `name` is not an attribute of `product_definition` in AP214 — it is DERIVED onto a getter that throws, so the `?? ''` was unreachable and `getSpatialStructure` failed for the whole model. See the smoke item. |

### conway-geom churn — **now blessed**

**This is the part of the bump that can change what a model looks like.** Three rounds moved `conway-geom` during this PR's life, and all three are blessed together by the `rc-1.588.1550` pass:

| round | via | driver | outcome |
|---|---|---|---|
| **R1** | conway#567 | conway-geom#178 — NURBS inverse-solve convergence target made relative | 4 models, 1,179 of 178,273 rows |
| **R2** | conway#578 | conway-geom#180 — stale tessellation candidates + NURBS trim-bounds cap | 4 models, 15 of 179,788 rows |
| **R3** | conway#581 | conway-geom#183 — deflection floor scoped to the defining representation | **blessed** — see below |

**The combined bless is [test-models#58](https://github.com/bldrs-ai/test-models/pull/58)**, from [run 32884504977](https://github.com/bldrs-ai/conway/actions/runs/32884504977) on tag `rc-1.588.1550` @ [`d75c2f03`](https://github.com/bldrs-ai/conway/commit/d75c2f0337f2588df2282ca760dff4bdde5aff14). 33 baselines changed, dominated by `Arty_Z7.csv` at **1171 rows** — which matches conway#581's own prediction of "1,169 of 94,510 rows across 3 models". Everything else moves 1–34 rows. No NEW parse/extract failures, no unexpected zero-geometry models. `benchmarks/conway1.588.1550-ci_test-models/` carries the perf snapshot and its delta against the previous blessed one.

R3 is still the round most likely to be *visible*, because a deflection floor scoped differently changes tessellation density directly, and the payload dropping 66% on Arty is exactly that showing up as bytes. **A digest bless records that the change happened; it does not judge whether it looks right.** That is what the Arty_Z7 "look at it" smoke item is for.

Two things in R2 are improvements you could see rather than just digest movement: `Orbiter_v1.1_Gear_7.5.step` stops trapping (`memory access out of bounds` on two `ADVANCED_FACE`s, gone), and conway#578 adds an extractor fix so an UNSPECIFIED-trimmed `IFCLINE` carrying only parameter values no longer collapses to a single point.

### Why 1.549 → 1.560 needed no re-smoke

**Scoped to that bump only.** Share reaches conway through exactly three import paths — `@bldrs-ai/conway/compiled/src/compat/web-ifc/`, `@bldrs-ai/conway/src/core/progress_log`, and `@bldrs-ai/conway/web-ifc`. **Neither `compat/` nor `core/progress_log` is touched anywhere in 1.549 → 1.560.** Outside the CLI harnesses and tests, five files changed, all either comment-only, Node-only, additive perf columns, or not referenced by Share. And rc-1.560.1539's full-corpus run changed **zero** digest files on **both** corpora against 1.549.1515.

### Why everything after 1.560 *does* need one

**Six PRs across 1.560 → 1.578 → 1.564.1548 change the code Share's store-backed deferred open runs**, and three of them change geometry: #563 and #566 rewrite the windowed prefetch; #568 §4 changes which related objects the aggregates pass defers; #569 changes the numbers on the `geometry` progress events Share renders; #582 changes AP214 unit scheduling; #567, #578 and #581 move `conway-geom` three times.

**And 1.564.1548 → 1.588.1550 adds one more that is runtime for Share:** #588 changes `resolveLabel` in `AP214ProductStructureExtraction`, which is what Share's STEP NavTree reaches through `getSpatialStructure`. It is a pure repair — the old path threw — but it is not inert, so it gets its own smoke item rather than being waved through.

---

## Smoke history

| round | conway | outcome |
|---|---|---|
| 1 | 1.514.1452 | PSB 62.7 s; imposters offset and solid-faced → conway#515 |
| 2 | 1.516.1463 | 61.4 s — frame fix cost nothing; plates still rotated; blank first ~10 s → conway#517 / #518 |
| 3 | 1.519.1471 | PSB clean at 53.4 s, but a **larger model failed outright** with `StepBufferNotResidentError` → conway#526 |
| 4 | 1.394.1504 | PSB 52.1 s — Parse 15.5 s, Geometry 30.1 s, 60.7 → 2299 MB, `window=60.66` |
| 5 | 1.543.1513 | First round where the preview read as intended. **SKYLARK250 hung the browser** → conway#549, fixed by #550. |
| 6 | 1.578.1546 | Not smoked — superseded before a human got to it. |
| 7 | 1.564.1548 | Not smoked — superseded by the rc pin below. |
| **8** | **1.588.1550** | **Smoked and approved.** |

Two runs at 84 s and 127 s early on were machine-load noise: Geometry varied 1.9× between identical runs.

---

## Smoke checklist for this head (`00de01a6`, conway 1.588.1550)

Deploy preview; load PSB from OPFS cache, the large model that died on round 3, **SKYLARK250**, **Arty_Z7**, and **`nist_ctc_04_asme1_rd.stp`**.

- **Arty_Z7 is the headline.** It is the model both #581 and #582 target, and a STEP/AP214 file, so it exercises the *resident* path rather than the windowed one. Three things to judge:
  - **No long freeze.** #582 cuts the worst single main-thread block **17,731 → 600 ms**. A multi-second lockup is a finding.
  - **Geometry roughly a third faster.** Engine-measured **34,169–37,235 → 22,484–24,985 ms** composed. A Geometry time back near 34 s is a regression.
  - **Look at it, not just the clock.** Payload **−66%**, peak wasm heap **−49%**, because the deflection floor is now scoped to the defining representation — a tessellation-density change. The baselines are blessed now, but a bless records *that* geometry moved, not that it looks right. **Anything coarser, faceted, or dropped is exactly what this smoke is for.**
- **`nist_ctc_04_asme1_rd.stp` NavTree** (#588). The tree should read **`NIST PMI CTC 04 ASME1`**, not `Object`, with **no red `Error: Function not implemented.`** in the console. Four of the six affected corpus models get a real name; `nist_stc_09_asme1_ap242-e3` and `nist_ftc_09_asme1_ap242-e1` get `''` because those files carry no name, id or description at all — that is correct, not a miss. Note a reload will look fine either way, because the GLB cache hits and the STEP path never re-runs — **judge it on a cold load.**
- **SKYLARK250 must load without hanging the tab.** Primary pass condition is **UI responsiveness**. conway#566 measured parse→first-mesh at **266 ms** (from 315,903 ms before #563) and longest event-loop starvation at **45 ms** (from 34,866 ms). Those are Node-harness figures, so a browser number in the same *order* is confirmation — **but a stall of even one or two seconds is a 20–40× regression and a finding, not a pass.** Nothing in this pin touches that path, so a *change* here is the finding.
- **The Geometry progress bar should advance in products, not sit at half.** conway#569 fixed a denominator that counted `IfcRelAggregates` records rather than the products under them. **Two caveats before calling a stall a regression:** (1) the **Share half of [conway#565](https://github.com/bldrs-ai/conway/issues/565) is not in this PR** — `conwayDirectIfcLoader.js` still latches `geometryTotal` from the first pump return and accumulates `extracted` itself, competing with conway's own `geometry` events; and (2) `geometryDone` only counts products that actually produced geometry, so on a model with geometry-less products the bar can legitimately stop short of full. A bar that *moves* is the pass condition; a bar frozen at a fixed fraction for the whole stage is the regression.
- **PSB Geometry should hold at round 4's ~⅓ drop** from 30.1 s. Parse and Assemble unchanged. A Geometry *regression* is the finding.
- **Memory should stay visibly lower** than round 4's 2299 MB peak, from the 64 MB budget. The report's heap column is `performance.memory` and excludes `WebAssembly.Memory`, so wasm-side improvements are only partly visible.
- **The round-3 model must complete.** A product that genuinely cannot be paged should degrade to one missing product plus a console log, never an aborted load.
- **Preview meshes appear during Parsing, filling in batches**, not one girder at a time.
- **No cubes or plates at any point** — imposters are off. If one appears, that is a finding.
- **Do not look for a `Preview:` line**; Share cannot show it yet (conway#544).
- Nothing left over after Total; `window=` still ≈64 MB; no extract `TypeError`s.
- Load log will read **`Conway v1.588.1550`**.

---

## Verified before smoking

**Stated precisely, because the two heads were verified to different depths.**

The full local gate — the checks a draft PR never gets, since `build` and both `test-flows` jobs are skipped on drafts — was run at **conway 1.564.1548 on head `0d0ed0f5`**:

| check | result at 1.564.1548 (`0d0ed0f5`) | baseline at 1.578.1546 (`9b1fe767`) |
|---|---|---|
| `yarn build-prod` | **exit 0** | exit 0 |
| `yarn test-ci` | **240 suites / 240, 2382 passed + 5 skipped / 2387, 13 snapshots; coverage gate held** | identical |
| **Playwright E2E** | **153 passed / 0 failed / 44 skipped**, 197 tests, **0 flaky**, **12.7 min** at CI's own 2 workers | 153 / 0 / 44, 197, 0 flaky, 11.8 min |

Zero delta in either direction. The 11.8 → 12.7 min wall clock is machine-load noise: no individual spec changed status.

**At the current head `00de01a6` (conway 1.588.1550), only the husky pre-commit gate has been re-run** — eslint + typecheck + full Jest, green: **240 suites / 240, 2382 passed + 5 skipped, 13 snapshots**. `yarn build-prod`, `yarn test-ci` coverage thresholds and Playwright have **not** been re-run at this pin. The delta over the fully-verified head is two conway PRs, and only one of them (#588) is runtime for Share. Flipping to ready fires the gated CI jobs, which covers it.

Note what the gate does and does not cover. It exercises Share's own code against the new engine; it does **not** load PSB, SKYLARK250, Arty_Z7 or any large model, so **none of the runtime improvements this pin is for are verified by it** — that is what the smoke checklist is for. In particular **the E2E suite cannot see the conway-geom changes at all**, because no fixture in it is NURBS-heavy. The suite guards against the bump *breaking* something; it is not evidence that it *helps*.

The suite was also run at **1.536.1507, 1.543.1513, 1.549.1515 and 1.578.1546** earlier in this PR's life, with the failing-test sets diffed programmatically: no test newly failed at the new pin and none newly passed, so nothing has implicated a conway bump at any point in this arc.

**The two E2E failures this PR previously carried are gone.** `src/Components/Open/Github.spec.ts` `:152` and `:163` were filed as #1759 and fixed on main by #1764, which this branch merged in (`4ac7836`).

Worth recording, because it changes how to read the earlier rounds: **#1759 was never a product bug.** The dropdowns open on an ordinary click. The tests were clicking with `force: true` straight through a MUI backdrop that stays mounted for the ~400 ms Grow exit, so the click was eaten as a click-away; and the always-on render loop makes every Playwright action ~7× more expensive on a GPU-less runner (one click: 978 ms with the viewer live, 139 ms with rendering paused). The fix is test-only.

One caveat on that merge: **#1764's final diff never received an automated review pass.** Codex hit its usage limit. What backs it instead is green CI on every job, a full local suite at 2 workers, and independent mutation checks on both fixes.

---

## Corpus evidence

**rc-1.588.1550** ([run 32884504977](https://github.com/bldrs-ai/conway/actions/runs/32884504977)) — **the pass this PR was waiting on**, and the one that blesses R1, R2 and R3 together:

- Public bless [test-models#58](https://github.com/bldrs-ai/test-models/pull/58): 33 baselines, **`Arty_Z7.csv` 1171 rows** (R3's deflection floor, matching #581's prediction), everything else 1–34 rows.
- **No NEW parse/extract failures.** No unexpected zero-geometry models.
- `benchmarks/conway1.588.1550-ci_test-models/` carries this release's steady per-model timings plus a delta against the previous blessed snapshot — the first rc in this arc to produce one, since the perf snapshot is `rc-*`-tag-gated by design and an earlier `workflow_dispatch` run skipped it silently.

**rc-1.549.1515** ([run 32550854090](https://github.com/bldrs-ai/conway/actions/runs/32550854090)) — private corpus:

- **Zero digest churn.** **Zero `OK → FAIL`.** 97 `OK → OK`, 2 `FAIL → OK`, 8 `N/A → OK`.
- Of the 28 models taking ≥5 s, **26 got faster** — PSB 43.8 → 24.7 s, Autodesk Hospital 30.2 → 19.2 s, DOWA 16.4 → 13.0 s. (That delta spans 1.451 → 1.549, so it credits the whole streaming arc.)

Small models pay a flat ~+85–108 ms in the geometry stage regardless of size — a fixed setup cost, not a scaling regression. Worth a decision at some point; not a blocker here.

**rc-1.560.1539** ([run 32609023219](https://github.com/bldrs-ai/conway/actions/runs/32609023219)) — **both** corpora: zero digest churn on both, `geometryMemoryMbDelta` 0.00 on all 97 public rows. The public corpus became whole again here — two formerly LFS-broken models produce digests for the first time ([test-models#52](https://github.com/bldrs-ai/test-models/issues/52) → [#53](https://github.com/bldrs-ai/test-models/pull/53)).

That run gives SKYLARK250 its numbers from the shipped bench. **These are the pre-#563/#566 figures:**

| | |
|---|---|
| parse / geometry / total | 4,964 / 21,780 / **26,744 ms** |
| `geometryMemoryMb` (payload) | 185.2 |
| `peakWasmHeapMb` | 706.9 |
| `peakRssMb` | **2,618.3** |
| `retainedRssMb` after teardown + settle | 2,263.7 |

The bench opens a **resident, fully-extracted** model where Share runs a **windowed deferred pump**, so it cannot exercise #563 or #566 and has no column approximating time-to-first-mesh — [conway#562](https://github.com/bldrs-ai/conway/issues/562) §2, still open. It does exercise the resident path that #581/#582 change.

The JS heap releases completely; RSS does not. Retained external (572 MB) is close to the wasm heap, which is grow-only and so structurally un-returnable. Corpus-wide the retention ratio climbs with model size (31% → 87%, median 56% over the 29 models above 300 MB), which reads as allocator arenas rather than a SKYLARK-specific leak. Tracked under [conway#554](https://github.com/bldrs-ai/conway/issues/554); does not gate this PR.

Ignore the `perf-three-*` delta comments on conway's rc runs. That is a separate headless-three render benchmark whose baseline is "whatever artifact ran last" and whose engine labels both read `conway1.1.0`. Its 1.558 → 1.560 table shows every one of 47 models 19–32% slower with `Δ geomMem MB` = 0 and RSS flat — a uniform scale factor across a 4000× dynamic range, i.e. a slower runner, not conway.

---

## Open upstream, and relevant to what you will see

- **[conway#565](https://github.com/bldrs-ai/conway/issues/565)** — **open for the Share half.** conway#569 fixed the engine's Geometry denominator; `conwayDirectIfcLoader.js` still derives a second, competing set of `geometry` progress events. Likely fix: delete Share's `reportGeometry` and let conway's pump own the phase. Directly shapes what the progress-bar smoke item can conclude.
- **[conway#574](https://github.com/bldrs-ai/conway/issues/574)** — a related product whose extraction throws is abandoned by the relationship-wide catch while the progress prefix still counted it, so the bar can drop a tail in one step.
- **[conway#546](https://github.com/bldrs-ai/conway/issues/546)** — grid-placed products lose preview retry, a regression from #543's own classification narrowing.
- **[conway#547](https://github.com/bldrs-ai/conway/issues/547)** — retry-first scheduling may starve fresh-product discovery on a large unresolvable backlog. Unverified.
- **[conway#544](https://github.com/bldrs-ai/conway/issues/544)** — the `Preview:` line still is not visible in Share.
- **[conway#533](https://github.com/bldrs-ai/conway/issues/533)** — auto-publish derives the minor from the merge title's first `#N`. Bitten twice in this ledger (1.394.1504, 1.564.1548); the current pin is unaffected.
- **[conway#545](https://github.com/bldrs-ai/conway/issues/545)** — conway's CI clones an `eigen` submodule from gitlab.com on every run and never compiles it. Has failed with both 403 and 502, and once silently blocked a publish.
- **[conway#551](https://github.com/bldrs-ai/conway/issues/551)** — the SKYLARK probe should be committed so its numbers stay falsifiable.

### Landing behind this PR, not in it

The `rc-1.588.1550` engine still carries two known silent-correctness defects in STEP curve handling, both found while investigating [test-models-private#93](https://github.com/bldrs-ai/test-models-private/issues/93) (Orbiter). Neither is in this pin; both are queued for the next rc, and both are listed here so a later finding on a STEP model can be matched against them rather than filed twice:

- **[conway#589](https://github.com/bldrs-ai/conway/issues/589)** (fix in [#593](https://github.com/bldrs-ai/conway/pull/593), green, held) — `readReal` returns `NaN` for any real whose total decimal shift reaches 64, silently, in **both** the STEP and IFC front ends. On Orbiter that is 48,593 NaN vertices across five solids. 6 models move.
- **[conway#599](https://github.com/bldrs-ai/conway/issues/599)** (fix in [conway-geom#185](https://github.com/bldrs-ai/conway-geom/pull/185) + [conway#600](https://github.com/bldrs-ai/conway/pull/600), drafts, held) — a cartesian B-spline trim is searched over raw knot values instead of the normalised parameter, so **86% of the corpus's cartesian-trimmed 3D extractions** search a wrong sub-interval and many collapse to a straight chord where an arc belongs. 51% of Orbiter's trimmed edges and 72% of `Right_Hand.step`'s were chords. 24 models move.

## Not closed by this PR

Share #1602, conway #446 and conway #392 stay open until write-through of a network body lands.

## Related

[#1756](https://github.com/bldrs-ai/Share/pull/1756) (geometry worker pool, `?feature=workers`) is stacked on this branch and consumes the #536/#538 sharding machinery that is inert here.

**Smoked and approved on this head.** Merging main in before submit.

### DIFF
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -362,8 +362,9 @@ export async function load(
       // Non-GitHub sources have no upstream sha. Hash the parse buffer
       // itself so we do not materialise the file twice (once for SHA-1,
       // once for Conway) — ~860 MB on PSB. sha1Hex only reads; it does
-      // not copy. A GLB cache hit still replaces modelData with the
-      // artifact (different bytes).
+      // not copy. Store-backed opens fingerprint the File in native
+      // per-slice SHA-1s (not a JS SHA-1 of the whole object). A GLB
+      // cache hit still replaces modelData with the artifact.
       //
       // Self-contained try/catch: if the hash or cache lookup fails (e.g.
       // `window.crypto.subtle` absent in some jsdom test envs, OPFS sync-
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -62,7 +62,7 @@ import stlToThree from './stl'
 import xyzToThree from './xyz'
 import {isFeatureEnabled} from '../FeatureFlags'
 import useStore from '../store/useStore'
-import {sha1Hex} from '../utils/contentHash'
+import {sha1Hex, sha1HexFromBlob} from '../utils/contentHash'
 import {markIfOutOfMemory} from '../utils/oom'
 
 
@@ -80,6 +80,7 @@ import {markIfOutOfMemory} from '../utils/oom'
 // the camera nonstop). 5s is past the point where a returning visitor
 // would notice "cache hasn't warmed yet."
 const SCHEDULE_IDLE_TIMEOUT_MS = 5_000
+const LFS_POINTER_PROBE_BYTES = 256
 
 
 /**
@@ -91,6 +92,42 @@ function scheduleIdleWork(fn) {
     return
   }
   setTimeout(fn, 0)
+}
+
+
+/**
+ * True when Conway can parse an OPFS File through `OpenModelStream`
+ * (M1b) so we never allocate a full-source ArrayBuffer.
+ *
+ * @param {object} viewer
+ * @param {boolean} isIfc
+ * @return {boolean}
+ */
+function canOpenFromStore(viewer, isIfc) {
+  if (!isIfc || isFeatureEnabled('disableStreamOpen')) {
+    return false
+  }
+  const ifcAPI = viewer?.IFC?.loader?.ifcManager?.ifcAPI
+  return typeof ifcAPI?.OpenModelStream === 'function'
+}
+
+
+/**
+ * Bytes for the Git LFS pointer sniff. A live File/Blob was already
+ * probed via its first 256 bytes before being passed through as
+ * `modelData`; don't treat the handle itself as a pointer.
+ *
+ * @param {ArrayBuffer|Uint8Array|Blob|string|null|undefined} data
+ * @return {ArrayBuffer|Uint8Array|string|null|undefined}
+ */
+function probeLfsBytes(data) {
+  if (data !== null && data !== undefined &&
+      typeof data.size === 'number' && typeof data.slice === 'function' &&
+      !(data instanceof ArrayBuffer) && !ArrayBuffer.isView(data) &&
+      typeof data !== 'string') {
+    return null
+  }
+  return data
 }
 
 
@@ -338,9 +375,18 @@ export async function load(
       // do anyway if the lookup returned null.
       if (wantGlb && !cacheKeyArgs && file) {
         try {
-          onProgress('Buffering model bytes...')
-          modelData = await file.arrayBuffer()
-          const contentSha = await sha1Hex(modelData)
+          // M1b: when Conway can parse from the OPFS File store, hash
+          // the File in slices and skip the full-source ArrayBuffer —
+          // ~860 MB on PSB. Older engines fall through to buffering.
+          if (canOpenFromStore(viewer, isIfc)) {
+            onProgress('Hashing model...')
+          } else {
+            onProgress('Buffering model bytes...')
+            modelData = await file.arrayBuffer()
+          }
+          const contentSha = modelData !== undefined ?
+            await sha1Hex(modelData) :
+            await sha1HexFromBlob(file)
           cacheKeyArgs = buildNonGitHubCacheArgs(kindLabel, path, contentSha)
           if (cacheKeyArgs) {
             glbVerbose(
@@ -374,11 +420,23 @@ export async function load(
         glbExportContext = {kindLabel, cacheKeyArgs, sourceFile: opfsSourceFile}
       }
 
+      if (modelData === undefined &&
+          canOpenFromStore(viewer, isIfc) &&
+          !cameFromGlbCache) {
+        const head = await file.slice(0, LFS_POINTER_PROBE_BYTES).arrayBuffer()
+        if (opfsEntryKey !== null && looksLikeLfsPointer(head)) {
+          // Fall through to the existing LFS-evict path below with a
+          // tiny probe so we don't have to buffer the whole file.
+          modelData = head
+        } else {
+          modelData = file
+        }
+      }
       if (modelData === undefined) {
         onProgress('Buffering model bytes...')
         modelData = await file.arrayBuffer()
       }
-      if (opfsEntryKey !== null && looksLikeLfsPointer(modelData)) {
+      if (opfsEntryKey !== null && looksLikeLfsPointer(probeLfsBytes(modelData))) {
         // A cache entry written before the Git LFS redirect landed
         // holds the ~130-byte pointer instead of the model, and its key
         // (pointer-blob sha + owner/repo/branch/path) is exactly the
@@ -480,7 +538,7 @@ export async function load(
   // us ~130 bytes of text naming the object instead of the model, and
   // the format loader then fails somewhere deep in its parser with an
   // error that never mentions LFS. Say what actually went wrong.
-  if (looksLikeLfsPointer(modelData)) {
+  if (looksLikeLfsPointer(probeLfsBytes(modelData))) {
     throw new Error(
       'This file is stored with Git LFS, so the URL returned a pointer file instead of the model. ' +
       'Open it via its github.com/<org>/<repo>/blob/<ref>/<path> URL, which resolves LFS content.')
@@ -505,7 +563,7 @@ export async function load(
   reportModelInfo({
     fileName: path.substring(path.lastIndexOf('/') + 1) || path,
     schema: formatTag,
-    byteLength: modelData?.byteLength ?? modelData?.length,
+    byteLength: modelData?.byteLength ?? modelData?.length ?? modelData?.size,
   })
 
   // readModel throws on any falsy model (surfacing viewer.IFC.ifcLastError
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -317,9 +317,16 @@ export async function load(
           'Source: network download (GitHub), cached to OPFS')
       }
 
-      // For non-GitHub sources we don't have an upstream sha, so we hash the
-      // bytes ourselves to build the cache key. This is the same File we'd
-      // read for parse below; reading it twice is cheap (OPFS).
+      // Disk handle of the *source* (IFC/STEP), captured before any GLB
+      // swap. Spill pages this File; the hash/parse ArrayBuffer is a
+      // separate JS copy. An OPFS File pins nothing.
+      const opfsSourceFile = file
+
+      // Non-GitHub sources have no upstream sha. Hash the parse buffer
+      // itself so we do not materialise the file twice (once for SHA-1,
+      // once for Conway) — ~860 MB on PSB. sha1Hex only reads; it does
+      // not copy. A GLB cache hit still replaces modelData with the
+      // artifact (different bytes).
       //
       // Self-contained try/catch: if the hash or cache lookup fails (e.g.
       // `window.crypto.subtle` absent in some jsdom test envs, OPFS sync-
@@ -331,8 +338,9 @@ export async function load(
       // do anyway if the lookup returned null.
       if (wantGlb && !cacheKeyArgs && file) {
         try {
-          const sourceBytes = await file.arrayBuffer()
-          const contentSha = await sha1Hex(sourceBytes)
+          onProgress('Buffering model bytes...')
+          modelData = await file.arrayBuffer()
+          const contentSha = await sha1Hex(modelData)
           cacheKeyArgs = buildNonGitHubCacheArgs(kindLabel, path, contentSha)
           if (cacheKeyArgs) {
             glbVerbose(
@@ -346,6 +354,7 @@ export async function load(
               ;[loader, isLoaderAsync, isFormatText, isIfc, fixupCb] = swapToGlbLoader(viewer)
               file = glbFile
               cameFromGlbCache = true
+              modelData = await glbFile.arrayBuffer()
             } else {
               glbInfo(`reader: ${kindLabel} cache MISS, will export after parse`)
             }
@@ -359,17 +368,16 @@ export async function load(
       }
 
       if (wantGlb && isIfc && cacheKeyArgs) {
-        // `sourceFile` is the OPFS-backed File whose exact bytes are
-        // parsed below (`modelData = await file.arrayBuffer()`; IFC/STEP
-        // load with isFormatText=false, so no decode in between). The
-        // post-writer source spill backs Conway's window reads with it —
-        // identity by construction, and a disk-backed handle, so
-        // capturing it here pins nothing.
-        glbExportContext = {kindLabel, cacheKeyArgs, sourceFile: file}
+        // Spill backs Conway's window reads with this OPFS File — the
+        // source, never a swapped-in GLB. Identity by construction
+        // (same handle we arrayBuffer'd for parse on a miss).
+        glbExportContext = {kindLabel, cacheKeyArgs, sourceFile: opfsSourceFile}
       }
 
-      onProgress('Buffering model bytes...')
-      modelData = await file.arrayBuffer()
+      if (modelData === undefined) {
+        onProgress('Buffering model bytes...')
+        modelData = await file.arrayBuffer()
+      }
       if (opfsEntryKey !== null && looksLikeLfsPointer(modelData)) {
         // A cache entry written before the Git LFS redirect landed
         // holds the ~130-byte pointer instead of the model, and its key
--- a/src/loader/Loader.js
+++ b/src/loader/Loader.js
@@ -414,9 +414,8 @@ export async function load(
       }
 
       if (wantGlb && isIfc && cacheKeyArgs) {
-        // Spill backs Conway's window reads with this OPFS File — the
-        // source, never a swapped-in GLB. Identity by construction
-        // (same handle we arrayBuffer'd for parse on a miss).
+        // Spill pages from the OPFS source handle captured before any
+        // GLB swap — never a cache-hit artifact.
         glbExportContext = {kindLabel, cacheKeyArgs, sourceFile: opfsSourceFile}
       }
 
@@ -424,9 +423,7 @@ export async function load(
           canOpenFromStore(viewer, isIfc) &&
           !cameFromGlbCache) {
         const head = await file.slice(0, LFS_POINTER_PROBE_BYTES).arrayBuffer()
-        if (opfsEntryKey !== null && looksLikeLfsPointer(head)) {
-          // Fall through to the existing LFS-evict path below with a
-          // tiny probe so we don't have to buffer the whole file.
+        if (looksLikeLfsPointer(head)) {
           modelData = head
         } else {
           modelData = file
--- a/src/viewer/ifc/conwayDirectIfcLoader.js
+++ b/src/viewer/ifc/conwayDirectIfcLoader.js
@@ -158,12 +158,39 @@ export async function parseIfcWithConway(
     // onto the screen and one that shows nothing until the end-of-load
     // build, and until now the log said nothing either way — a blank
     // 9-second parse and a healthy streaming parse looked identical.
+    //
+    // Conway's deferred open only emits headerParse/dataParse; the
+    // pump never starts a geometry phase. Report one here so the load
+    // log splits Parsing from Geometry (same labels as a classic
+    // extractIFCGeometryData open). elapsedMs omitted: the reporter
+    // stamps wall-clock so we don't fight Conway's parse-relative clock.
     let pumpedBatches = 0
+    let geometryTotal
+    let geometryDone = 0
+    const reportGeometry = (completed, total) => {
+      if (typeof onProgress !== 'function') {
+        return
+      }
+      onProgress({
+        phase: 'geometry',
+        completed,
+        total,
+        unit: 'products',
+      })
+    }
     for (;;) {
       const batch = []
       // eslint-disable-next-line new-cap
       const {extracted, remaining} = ifcAPI.ExtractGeometryBatch(
         modelID, DEMAND_EXTRACT_BATCH_SIZE, (flatMesh) => batch.push(flatMesh))
+      if (geometryTotal === undefined && (extracted > 0 || remaining > 0)) {
+        geometryTotal = extracted + remaining
+        reportGeometry(0, geometryTotal)
+      }
+      geometryDone += extracted
+      if (geometryTotal !== undefined) {
+        reportGeometry(geometryDone, geometryTotal)
+      }
       if (batch.length > 0) {
         captured.push(...batch)
         pumpedBatches++
--- a/src/viewer/ifc/conwayDirectIfcLoader.js
+++ b/src/viewer/ifc/conwayDirectIfcLoader.js
@@ -207,9 +207,14 @@ export async function parseIfcWithConway(
       let extracted
       let remaining
       if (typeof ifcAPI.ExtractGeometryBatchAsync === 'function') {
+        // Store-backed extract pages each product's #ref closure from
+        // OPFS before extracting it. A 64-product batch serialises
+        // that I/O and holds first pixels until ~halfway through
+        // Geometry; 8 keeps file-order streaming and still amortises
+        // the per-batch scene update.
         // eslint-disable-next-line new-cap
         const pumped = await ifcAPI.ExtractGeometryBatchAsync(
-          modelID, DEMAND_EXTRACT_BATCH_SIZE, (flatMesh) => batch.push(flatMesh))
+          modelID, ASYNC_DEMAND_EXTRACT_BATCH_SIZE, (flatMesh) => batch.push(flatMesh))
         extracted = pumped.extracted
         remaining = pumped.remaining
       } else {
@@ -324,8 +329,11 @@ export async function parseIfcWithConway(
 
 // Products extracted per demand batch: large enough that per-batch
 // capture/render overhead amortizes, small enough that first pixels
-// arrive within a couple of seconds of parse completing.
+// arrive within a couple of seconds of parse completing. The async
+// (store-backed) path is smaller because each product pays OPFS
+// prefetch before extract — see ExtractGeometryBatchAsync above.
 const DEMAND_EXTRACT_BATCH_SIZE = 64
+const ASYNC_DEMAND_EXTRACT_BATCH_SIZE = 8
 
 
 /**
--- a/src/viewer/ifc/conwayDirectIfcLoader.js
+++ b/src/viewer/ifc/conwayDirectIfcLoader.js
@@ -139,7 +139,11 @@ export async function parseIfcWithConway(
       !isFeatureEnabled('disableStreamOpen') &&
       typeof ifcAPI.OpenModelStreamed === 'function' &&
       typeof ifcAPI.ExtractGeometryBatch === 'function') {
-    const deferSettings = {...openSettings, DEFER_GEOMETRY: true}
+    const deferSettings = {
+      ...openSettings,
+      DEFER_GEOMETRY: true,
+      GEOMETRY_BUDGET_MB: GEOMETRY_BUDGET_MB,
+    }
     if (onPreviewMesh) {
       // Slice A2 (parse-time preview channel): conway emits preview
       // payloads between the parse's cooperative yields. Passing the
@@ -254,6 +258,28 @@ export async function parseIfcWithConway(
 // Products extracted per demand batch: large enough that per-batch
 // capture/render overhead amortizes, small enough that first pixels
 // arrive within a couple of seconds of parse completing.
+/**
+ * Cap on the native geometry conway keeps resident, in MB, for the deferred
+ * path only. Least-recently-used assets are evicted at each pump batch.
+ *
+ * Measured engine-side on PSB (860 MB) at our batch size: the wasm high-water
+ * falls from 1284 MB to 298 MB, with the geometry stage slightly faster
+ * (23.6s against 25.7s) and identical delivered meshes. 256 MB was also
+ * measured and buys much less (803 MB), because the live set rarely reaches
+ * the ceiling. The heap runs 3-4x the budget — allocator overhead and
+ * fragmentation — so this targets roughly a 300 MB residency, not 64 MB.
+ *
+ * Safe here for one specific reason: eviction frees an asset from
+ * GetGeometry until something re-extracts it, and this loader copies every
+ * payload at delivery (the bldrs-ai/Share#1640 invariant). A change that
+ * made it hold geometry IDs and fetch them later would break quietly, so
+ * that invariant is now load-bearing rather than merely true.
+ *
+ * Ignored by engines predating conway#535 — unknown settings are dropped —
+ * so this is inert until the conway bump in #1753 lands.
+ */
+const GEOMETRY_BUDGET_MB = 64
+
 const DEMAND_EXTRACT_BATCH_SIZE = 64
 
 
--- a/src/viewer/ifc/conwayDirectIfcLoader.js
+++ b/src/viewer/ifc/conwayDirectIfcLoader.js
@@ -122,7 +122,7 @@ export async function parseIfcWithConway(
       typeof ifcAPI.OpenModelStream === 'function' ?
     makeBlobByteStore(buffer) :
     null
-  const data = store === null ?
+  let data = store === null ?
     (buffer instanceof Uint8Array ? buffer : new Uint8Array(buffer)) :
     null
   let openSettings = settings ?? {COORDINATE_TO_ORIGIN: true, USE_FAST_BOOLS: true}
@@ -158,12 +158,20 @@ export async function parseIfcWithConway(
       deferSettings.ON_PREVIEW_MESH = onPreviewMesh
     }
     let modelID
+    let openData = data
     if (store !== null) {
       // eslint-disable-next-line new-cap
       modelID = await ifcAPI.OpenModelStream(store, deferSettings)
+      if (typeof modelID !== 'number' || modelID < 0) {
+        // IFC-only store path: STEP / failed sniff falls back to a
+        // buffered streamed open (conway#510 contract).
+        openData = await bytesFromSource(buffer)
+        // eslint-disable-next-line new-cap
+        modelID = await ifcAPI.OpenModelStreamed(openData, deferSettings)
+      }
     } else {
       // eslint-disable-next-line new-cap
-      modelID = await ifcAPI.OpenModelStreamed(data, deferSettings)
+      modelID = await ifcAPI.OpenModelStreamed(openData, deferSettings)
     }
     if (typeof modelID !== 'number' || modelID < 0) {
       throw new Error(`parseIfcWithConway: OpenModel returned ${modelID}`)
@@ -287,6 +295,11 @@ export async function parseIfcWithConway(
   if (!isFeatureEnabled('disableStreamOpen') && store !== null) {
     // eslint-disable-next-line new-cap
     modelID = await ifcAPI.OpenModelStream(store, openSettings)
+    if (typeof modelID !== 'number' || modelID < 0) {
+      data = await bytesFromSource(buffer)
+      // eslint-disable-next-line new-cap
+      modelID = await ifcAPI.OpenModelStreamed(data, openSettings)
+    }
   } else if (!isFeatureEnabled('disableStreamOpen') && typeof ifcAPI.OpenModelStreamed === 'function') {
     // eslint-disable-next-line new-cap
     modelID = await ifcAPI.OpenModelStreamed(data, openSettings)
@@ -325,6 +338,26 @@ function isBlobSource(value) {
   return value !== null && value !== undefined &&
     typeof value.size === 'number' && typeof value.slice === 'function' &&
     !(value instanceof ArrayBuffer) && !ArrayBuffer.isView(value)
+}
+
+
+/**
+ * Materialise a parse source as a Uint8Array (store-open fallback).
+ *
+ * @param {ArrayBuffer|Uint8Array|Blob} source
+ * @return {Promise<Uint8Array>}
+ */
+async function bytesFromSource(source) {
+  if (source instanceof Uint8Array) {
+    return source
+  }
+  if (source instanceof ArrayBuffer) {
+    return new Uint8Array(source)
+  }
+  if (typeof source?.arrayBuffer === 'function') {
+    return new Uint8Array(await source.arrayBuffer())
+  }
+  throw new Error('parseIfcWithConway: cannot buffer source for fallback open')
 }
 
 
--- a/src/viewer/ifc/conwayDirectIfcLoader.js
+++ b/src/viewer/ifc/conwayDirectIfcLoader.js
@@ -48,6 +48,7 @@
 
 import {isFeatureEnabled} from '../../FeatureFlags'
 import {reportEngineVersion} from '../../loader/loadProgress'
+import {makeBlobByteStore} from '../../loader/opfsSourceByteStore'
 import debug, {WARN} from '../../utils/debug'
 import {attachInstanceMapSubsets} from '../three/elementSubsets'
 import {instanceMapFromGeometry} from './IfcInstanceMap'
@@ -116,7 +117,14 @@ export async function parseIfcWithConway(
   if (typeof ifcAPI.getConwayVersion === 'function') {
     reportEngineVersion(ifcAPI.getConwayVersion())
   }
-  const data = buffer instanceof Uint8Array ? buffer : new Uint8Array(buffer)
+  const store = isBlobSource(buffer) &&
+      !isFeatureEnabled('disableStreamOpen') &&
+      typeof ifcAPI.OpenModelStream === 'function' ?
+    makeBlobByteStore(buffer) :
+    null
+  const data = store === null ?
+    (buffer instanceof Uint8Array ? buffer : new Uint8Array(buffer)) :
+    null
   let openSettings = settings ?? {COORDINATE_TO_ORIGIN: true, USE_FAST_BOOLS: true}
   if (onProgress) {
     // ON_MODEL_INFO (conway extension) arrives once, right after the header
@@ -137,8 +145,10 @@ export async function parseIfcWithConway(
   // through to the classic selection below.
   if (isFeatureEnabled('demandGeometry') &&
       !isFeatureEnabled('disableStreamOpen') &&
-      typeof ifcAPI.OpenModelStreamed === 'function' &&
-      typeof ifcAPI.ExtractGeometryBatch === 'function') {
+      (typeof ifcAPI.OpenModelStreamed === 'function' ||
+        typeof ifcAPI.OpenModelStream === 'function') &&
+      (typeof ifcAPI.ExtractGeometryBatch === 'function' ||
+        typeof ifcAPI.ExtractGeometryBatchAsync === 'function')) {
     const deferSettings = {...openSettings, DEFER_GEOMETRY: true}
     if (onPreviewMesh) {
       // Slice A2 (parse-time preview channel): conway emits preview
@@ -147,8 +157,14 @@ export async function parseIfcWithConway(
       // settings are ignored), so no feature detection is needed here.
       deferSettings.ON_PREVIEW_MESH = onPreviewMesh
     }
-    // eslint-disable-next-line new-cap
-    const modelID = await ifcAPI.OpenModelStreamed(data, deferSettings)
+    let modelID
+    if (store !== null) {
+      // eslint-disable-next-line new-cap
+      modelID = await ifcAPI.OpenModelStream(store, deferSettings)
+    } else {
+      // eslint-disable-next-line new-cap
+      modelID = await ifcAPI.OpenModelStreamed(data, deferSettings)
+    }
     if (typeof modelID !== 'number' || modelID < 0) {
       throw new Error(`parseIfcWithConway: OpenModel returned ${modelID}`)
     }
@@ -180,9 +196,21 @@ export async function parseIfcWithConway(
     }
     for (;;) {
       const batch = []
-      // eslint-disable-next-line new-cap
-      const {extracted, remaining} = ifcAPI.ExtractGeometryBatch(
-        modelID, DEMAND_EXTRACT_BATCH_SIZE, (flatMesh) => batch.push(flatMesh))
+      let extracted
+      let remaining
+      if (typeof ifcAPI.ExtractGeometryBatchAsync === 'function') {
+        // eslint-disable-next-line new-cap
+        const pumped = await ifcAPI.ExtractGeometryBatchAsync(
+          modelID, DEMAND_EXTRACT_BATCH_SIZE, (flatMesh) => batch.push(flatMesh))
+        extracted = pumped.extracted
+        remaining = pumped.remaining
+      } else {
+        // eslint-disable-next-line new-cap
+        const pumped = ifcAPI.ExtractGeometryBatch(
+          modelID, DEMAND_EXTRACT_BATCH_SIZE, (flatMesh) => batch.push(flatMesh))
+        extracted = pumped.extracted
+        remaining = pumped.remaining
+      }
       if (geometryTotal === undefined && (extracted > 0 || remaining > 0)) {
         geometryTotal = extracted + remaining
         reportGeometry(0, geometryTotal)
@@ -256,7 +284,10 @@ export async function parseIfcWithConway(
   //   3. OpenModel: classic synchronous open (real web-ifc, old pins).
   // All feature-detected, so any engine pin keeps loading.
   let modelID
-  if (!isFeatureEnabled('disableStreamOpen') && typeof ifcAPI.OpenModelStreamed === 'function') {
+  if (!isFeatureEnabled('disableStreamOpen') && store !== null) {
+    // eslint-disable-next-line new-cap
+    modelID = await ifcAPI.OpenModelStream(store, openSettings)
+  } else if (!isFeatureEnabled('disableStreamOpen') && typeof ifcAPI.OpenModelStreamed === 'function') {
     // eslint-disable-next-line new-cap
     modelID = await ifcAPI.OpenModelStreamed(data, openSettings)
   } else if (typeof ifcAPI.OpenModelAsync === 'function') {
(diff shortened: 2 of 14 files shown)


src/viewer/ifc/conwayDirectIfcLoader.js
```diff
@@ -147,8 +157,14 @@ export async function parseIfcWithConway(
...
-    const modelID = await ifcAPI.OpenModelStreamed(data, deferSettings)
+    let modelID
+    if (store !== null) {
+      // eslint-disable-next-line new-cap
+      modelID = await ifcAPI.OpenModelStream(store, deferSettings)
+    } else {
+      // eslint-disable-next-line new-cap
+      modelID = await ifcAPI.OpenModelStreamed(data, deferSettings)
+    }
     if (typeof modelID !== 'number' || modelID < 0) {
```
pablo-mayrgundter: **[bug]** After Conway #510, OPFS `.step`/`.stp` loads will fail. Share sets `isIfc` for `ifc`/`step`/`stp`, so `canOpenFromStore` passes the File through and `parseIfcWithConway` calls `OpenModelStream`. Conway's contract is IFC-only: `OpenModelStream` returns `-1` with no fallback to `arrayBuffer()` + `OpenModelStreamed`.

**Suggestion:** If `OpenModelStream` returns `-1`, materialise the Blob and retry the existing streamed/classic selection.

src/loader/Loader.js
```diff
@@ -472,7 +538,7 @@ export async function load(
   // us ~130 bytes of text naming the object instead of the model, and
   // the format loader then fails somewhere deep in its parser with an
   // error that never mentions LFS. Say what actually went wrong.
-  if (looksLikeLfsPointer(modelData)) {
+  if (looksLikeLfsPointer(probeLfsBytes(modelData))) {
```
pablo-mayrgundter: **[bug]** `probeLfsBytes` returns `null` for a live File/Blob, so `looksLikeLfsPointer` is a no-op on the store-open path. The 256-byte head sniff only runs when `opfsEntryKey !== null` (GitHub). External/uploaded pointer files will be handed to Conway as a File after #510.

**Suggestion:** Always sniff `file.slice(0, LFS_POINTER_PROBE_BYTES)` before assigning `modelData = file`.

src/loader/Loader.js
```diff
@@ -359,18 +414,29 @@ export async function load(
...
-        // `sourceFile` is the OPFS-backed File whose exact bytes are
-        // parsed below (`modelData = await file.arrayBuffer()`; IFC/STEP
-        // load with isFormatText=false, so no decode in between). The
-        // post-writer source spill backs Conway's window reads with it —
-        // identity by construction, and a disk-backed handle, so
-        // capturing it here pins nothing.
-        glbExportContext = {kindLabel, cacheKeyArgs, sourceFile: file}
+        // Spill backs Conway's window reads with this OPFS File — the
+        // source, never a swapped-in GLB. Identity by construction
+        // (same handle we arrayBuffer'd for parse on a miss).
```
pablo-mayrgundter: **[suggestion]** The spill comment still says identity is "the same handle we arrayBuffer'd for parse on a miss." On the M1b path a miss never `arrayBuffer()`s the source.

**Suggestion:** State that `sourceFile` is the OPFS handle captured before any GLB swap. Drop the milestone/PSB recap.

pablo-mayrgundter:
Review findings from the pending pass are addressed in 0eb159be:

- **STEP fallback:** `OpenModelStream` returning `-1` (IFC-only) now buffers the File and retries `OpenModelStreamed`.
- **LFS sniff:** the 256-byte probe always runs before a File is passed through, not only when `opfsEntryKey` is set.
- **Spill comment:** now says the OPFS source handle is the one captured before any GLB swap.

Store-open stays a no-op on 1.509.1435. After conway#510 publishes we bump this pin for the PSB smoke.

pablo-mayrgundter:
Pinned **@bldrs-ai/conway@1.510.1439** (conway#510 merged + published). Draft deploy preview is the PSB smoke: expect no 848 MB Buffering copy on an OPFS cache-hit IFC. Leave this draft until that smoke lands.

pablo-mayrgundter:
Conway handoff for the 1.514.1452 smoke (`v1.1753.a9820d6`, 62.7 s, 69 → 2321 MB, `window=60.66`): [conway#515](https://github.com/bldrs-ai/conway/issues/515). Still draft; not merging from this comment.

pablo-mayrgundter:
Review of the imposter offset is on [conway#515](https://github.com/bldrs-ai/conway/issues/515#issuecomment-5310270661): conway emits the plates in raw model space (no scale, no coordination), and this PR's `applyAabbImposter` first-box re-anchor plus the aabb branch of `payloadToPreviewMesh` ignoring `coordination` compound it. Share-side changes queued for after conway publishes the frame fix:

- drop the ring-origin translation in `applyAabbImposter` (keep the cap + teardown),
- pass `coordination` into the aabb branch so imposters get the same durable offset as real preview meshes.

Wireframe plates come from conway dropping `solid: true` — no Share change needed there. Staying draft until that conway release + a human smoke.

---

pablo-mayrgundter:
Merged the imposter coordination-frame consumer fix (`474ee8c`, `258b90a`, `6219fc2` — fast-forward) and bumped the pin to **@bldrs-ai/conway 1.516.1463** for [conway#516](https://github.com/bldrs-ai/conway/pull/516).

Both halves of the contract change now move together:

- aabb-imposter payloads carry `flatTransformation` in the durable coordination frame (metres, Y-up, recentred), like every other preview payload — so the Share side drops the first-box ring-origin re-anchor and threads `coordination` through the aabb branch of `payloadToPreviewMesh` via a shared `recenterOntoCoordination_` helper.
- `payload.solid` is no longer emitted, so plates render wireframe.

Pre-commit gate is green (240 suites / 2371 tests), including the co-located `parsePreviewMesh.test.js` and `ShareIfcLoader.aabb.test.js` updates. The lockfile change is the conway entry only — no unrelated churn, and `package.json`'s `version` field is untouched.

The deploy preview is the smoke vehicle for this one: imposter placement and wireframe plates need eyes on a real model. Staying **draft** pending that human smoke test.

---

pablo-mayrgundter:
PSB smoke on `v1.1753.717bd01` / conway **1.516.1463** (clean machine): **Total 61.4 s** (Parse 16.7, Geometry 37.6, Assemble 3.5), 90 → 2305 MB, `window=60.66`, 8.80 M verts / 11.57 M tris — parity with the 62.7 s 1.514 baseline, so #516 cost nothing on wall-clock. (Two earlier runs at 84 s / 127 s were machine-load noise — Geometry varied 1.9× between identical runs; per-stage heap deltas swing with GC timing and aren't allocation truth.)

Smoke findings, filed upstream:

- Plates are now wireframe and correctly scaled ✅, but still offset/rotated vs the durable geometry → [conway#517](https://github.com/bldrs-ai/conway/issues/517) (placement-chain rotation ignored in the spatial walk).
- Blank first ~10 s + girder-by-girder preview deltas → [conway#518](https://github.com/bldrs-ai/conway/issues/518) (one-product ticks, late-placement deferral, spatial walk waiting for the full index; plan is time-budgeted ticks + skeleton from the first prefix generation).

Next conway round addresses both; this PR then takes one more pin bump + the imposter replace-by-expressID consumer change. Still draft pending that.

---

pablo-mayrgundter:
Pushed `aa531a0` — conway bumped to **1.519.1471**, plus the Share-side consumer half of the contract change that bump brings.

**What the bump carries (conway#519)**

- **Rotation-aware spatial plates (conway#517)** — imposter placement now composes the full `IfcLocalPlacement` chain, so plates no longer sit as a rotated ghost beside the durable geometry on Revit true-north models.
- **Time-budgeted preview extraction (conway#518)** — the store preview channel extracts under a ~20ms budget per tick instead of one product per tick, so the preview fills in batches rather than girder-by-girder.
- **Early skeleton + re-emission (conway#518)** — spatial plates are emitted early, from a prefix generation seconds into the parse, and re-emitted after the parse with full samples and the latched coordination frame.

**Share side: replace-by-expressID**

The third item changes the consumer contract: an `aabb` payload re-emitted with an `expressID` we've already drawn a plate for **replaces** that plate rather than adding a second one. Double emission is designed behaviour, so `applyAabbImposter` in `src/viewer/ifc/ShareIfcLoader.js` now keys accepted plates by `payload.expressID` alongside the arrival-order ring:

- a re-emission detaches the old mesh from the preview group and takes its ring slot — no extra cap slot consumed, and it can't evict a visible plate;
- nothing is disposed on replace (the unit-cube geometry and wire materials are pooled per load in `payloadToPreviewMesh`);
- cap eviction still applies to genuinely new expressIDs, and drops the key along with the mesh;
- teardown clears both the list and the keying map;
- the parented-only rule is preserved, which also settles the update semantics: if the outlier guard rejects the *new* mesh, the standing plate stays on screen — a refused update must not delete scenery it failed to refine.

Vertex-carrying payloads are unaffected (keyed by `geometryExpressID`, never re-sent). One known-and-accepted wrinkle: `ProgressiveLoadSession`'s preview bounds union is grow-only, so a replaced coarse plate's old bounds can linger in the camera-fit union until the next refit — noted in a comment, harmless for preview scenery that's torn down at load end. Upstream also documents a gap needing no Share action: a wrapper node emitted early but collapsed out of the final walk's emit set lingers as one small plate until load-end teardown.

Tests in `src/viewer/ifc/ShareIfcLoader.aabb.test.js` cover same-expressID replacement, no-cap-slot-spent on replace, distinct-expressID cap+evict (including the key going with the evicted mesh), rejected-replacement-keeps-standing-plate, and teardown clearing the map. Full pre-commit gate green locally (eslint + typecheck + 240 suites / 2374 tests).

**Smoke vehicle is the deploy preview.** What to look for on a big Revit model: plates appearing within the first seconds of the parse, correctly rotated and scaled against the durable geometry, preview filling in batches rather than one product at a time, and plates refining once at parse end without doubling up.

Staying in **draft** pending a human smoke.

---

pablo-mayrgundter:
@codex review

---

pablo-mayrgundter:
Codex's P1 ([preview resources never disposed](https://github.com/bldrs-ai/Share/pull/1753#discussion_r3799202358)) is confirmed and fixed in `c1812b3` + `09b943d` — replying here because a stale pending review blocks threaded replies. The dispose pass had been dropped wholesale when the shared aabb cube arrived rather than made shared-aware. `teardownPreview_` now collects UNIQUE geometries and materials from the stray imposters and the preview group's children and disposes each exactly once — correct for the pooled resources (rgba-keyed materials, `geometryExpressID`-keyed payload geometry, the imposter unit cube), whose caches are per-load Maps in `ShareIfcLoader.parse`, so nothing re-uses the disposed instances. Regression test: two preview meshes plus a stray imposter sharing one geometry/material dispose each exactly once at `finish()`. Touched suites green (32 tests), eslint clean.

The three older review threads (STEP `-1` fallback, unconditional LFS sniff, spill comment) were addressed back in `0eb159b` per the earlier thread; they predate this round.

---

pablo-mayrgundter:
Smoke round 4 is up: bumped to conway **1.529.1492** (`b46bbef`).

That release carries [conway#529](https://github.com/bldrs-ai/conway/pull/529), which closes [conway#526](https://github.com/bldrs-ai/conway/issues/526) — the `StepBufferNotResidentError` that killed the whole load on a large model during the 1.519.1471 round. Conway CI on that merge: `build` green, smoke regression clean (no failures, no errors, one resolved `No basis found for brep!` on `nist_ctc_02_asme1_rc.stp`), visual-diff below the 0.05% threshold on all 11 digest-changed models.

Share-side gate green on this commit: eslint + typecheck + 240 suites / 2375 tests.

Description updated with the round-4 smoke list. The new line to check is the model that failed on 1.519 — it should now complete; a product that genuinely cannot be paged should degrade to one missing product plus a console log rather than aborting the load. Staying draft for the human smoke.

---

pablo-mayrgundter:
## Bumped 1.529.1492 → 1.530.1503 (`3e42fe9`)

Picks up **conway#530 (M2)** on top of everything already on this branch:

- Streaming type index derived from the **columns**, not the events
- Prefix index made constructible, restart-safe, and honestly timed; a record is indexed *before* it's announced
- **Spatial names skeleton built while the file is still parsing** — relevant to this PR's early-plates path, since it's the same "usable structure before parse end" seam
- `LongName` read from each branch's own attribute position; rebuild minimum no longer suppresses rebuilds; pooled cursors returned after a prefix query

### Why 1.530.1503 and not main's head

Conway's head publish is **1.394.1504**, which is a *lower* semver than 1.530.1503 despite being newer. `auto-publish` derives the minor from the first `#N` in the merge commit message, and the M3 squash title led with the issue (`M3 (#394): … (#531)`), so it took 394. The `sort -V` guard caught the inversion and then took its `--tag latest` fallback, which pointed npm's `latest` **backwards**. Filed as [conway#533](https://github.com/bldrs-ai/conway/issues/533); it needs a manual `npm dist-tag` fix, which can't be done from a session (OIDC publishing, no token).

The two versions are **the same engine** — the M3 merge changed only `scripts/` and `design/` — so pinning 1.530.1503 loses nothing and avoids baking the backwards number into Share.

### Smoke additions for this round

Everything in the description still applies. Additionally, because M2 touches the parse-time index:

- Spatial tree/NavTree names appear during Parsing and are **correct** (the `LongName` attribute-position fix landed in this range — watch for storeys or spaces showing the wrong name)
- Names that appear early still match after parse end — the skeleton refines, it shouldn't rewrite
- No regression in the plates path, which shares the prefix-generation seam

Local gate passed (eslint + typecheck + full Jest, 13 snapshots). Staying **draft** per the description — still wants a human on the deploy preview.

---

pablo-mayrgundter:
Correction to the comment above: repinned to **1.394.1504** in `0410fae` — conway main's head, not 1.530.1503.

I diffed the two published tarballs to make the choice on evidence rather than assumption. The only differences are the version string itself:

```
package/package.json                       "version"
package/compiled/src/version/version.js    'Conway v1.530.1503' vs 'Conway v1.394.1504'
package/compiled/examples/*-bundled.cjs    the same string, baked in
package/compiled/tsconfig.tsbuildinfo      build stamp
```

Engine-identical, so the only question is which number Share carries — and main's head is the right answer. It's what npm resolves as `latest`, it keeps "which conway commit is Share on" answerable, and because Share pins exact versions the backwards semver from [conway#533](https://github.com/bldrs-ai/conway/issues/533) never reaches dependency resolution. The 1.530.1503 pin was protecting the version string from looking like a downgrade in the load log, which isn't worth the traceability.

Note for smoke: the load log will read **`Conway v1.394.1504`**. That is main's head despite the lower middle segment — see conway#533.

Local gate passed again (eslint + typecheck + Jest). Still draft.

---

pablo-mayrgundter:
## Bumped again: 1.394.1504 → 1.534.1505 (`07a9845`)

Picks up **conway#534 (M3 delta capture)**, merged just now. Worth re-smoking rather than smoking the previous pin, because this one changes the stage your load log actually reports.

### What it does

The demand pump's delta capture re-walked the **entire scene** on every `ExtractGeometryBatch` call, using per-entity watermarks to skip what it had already emitted — O(batches × scene). It now walks only what the scene gained since the last call.

Share pumps at **batch 8** (`DEMAND_EXTRACT_BATCH_SIZE` / `ASYNC_DEMAND_EXTRACT_BATCH_SIZE`), which is the worst point of the old curve, so this is aimed squarely at us. Engine-only measurements, minimal driver, no harness work in the timings:

| model | batch | before | after |
|---|---|---|---|
| **PSB.ifc** | **8** | 37.0 s | **25.3 s (1.47×)** |
| PSB.ifc | 64 | 27.3 s | 25.1 s |
| D3D.ifc (562 k instances) | 64 | 198.6 s | **66.8 s (2.97×)** |
| D3D.ifc | 256 | 102.2 s | 69.9 s |

### What to watch in the smoke

Your last run on 1.394.1504 logged:

```
Parsing:  15.543s, +524.7 MB heap
Geometry: 30.080s, +1246.7 MB heap, window=60.66 MB
Total:    52.067s, 60.7 → 2299.2 MB heap
```

- **The Geometry line is the one that should move** — the engine-side measurement predicts roughly a third off that stage on PSB. Parsing and Assembling should be unchanged.
- **Heap should not move.** The `+1246 MB` is `performance.memory.usedJSHeapSize`, which excludes `WebAssembly.Memory` — this change is about walking, not residency. If heap moves noticeably, something unexpected happened and I'd want to know.
- Everything already in the description still applies: plates early, correctly rotated, filling in batches, no doubling, the large model completing.

### Version note

`latest` moved forward properly this time (1.534.1505 > 1.530.1503). The previous pin, 1.394.1504, was numerically *below* the release it superseded because conway's version derivation read the issue number out of that merge title — [conway#533](https://github.com/bldrs-ai/conway/issues/533), still open, since this publish landing correctly is a property of my commit title rather than a fix.

Local gate passed (eslint + typecheck + Jest). Still **draft** pending your smoke.

---

pablo-mayrgundter:
## Bumped 1.534.1505 → 1.535.1506 (`039c4af`)

Picks up **conway#535**, the M3 resident-geometry budget. Local gate passed; still draft for your smoke.

### Behaviourally identical here, deliberately

The budget is **unlimited by default and Share does not set one**, so nothing about this load changes. Turning it on is a separate change, because eviction alters a contract: an evicted asset is gone from `GetGeometry` until something re-extracts it. Share is already compliant — `conwayDirectIfcLoader` copies payloads at delivery — which is precisely what makes the budget available to us when we want it.

What it buys when enabled, measured engine-side at Share's batch size of 8:

| budget | wasm peak | geometry |
|---|---|---|
| none | 1284 MB | 25.7 s |
| **64 MB** | **298 MB** | 23.6 s |
| 256 MB | 803 MB | 23.4 s |

### One fix that *does* apply today

conway#535 also refcounts geometry residency on **native identity**. A native can be cached under several local IDs (6 % of D3D's), and any mass delete previously freed it per key, leaving siblings pointing at freed memory. That affects `ReleaseModelGeometry`, which Share calls — so this release fixes a latent hazard on a path we already use, independent of the budget.

### Smoke

Same list as the previous round. The `Geometry:` stage should look like the 1.534 run (~27.5 s on PSB); this bump is not expected to move it. If it moves in either direction, that is worth knowing.

Next step for a follow-up, not this PR: wire `GEOMETRY_BUDGET_MB` or `SetGeometryBudget` into the loader so the tab holds a bounded residency, which is where the 1284 → 298 MB actually lands for users.

---

pablo-mayrgundter:
## Bumped to conway 1.536.1507 (`550c662`)

**1.506 → 1.507 is [conway#536](https://github.com/bldrs-ai/conway/pull/536) only, and on Share's path it is a no-op.** Worth stating up front so this doesn't get smoked looking for a difference that isn't there.

Everything #536 adds is gated behind a claimed shard:

| added | reachable from Share? |
|---|---|
| `geometryDispatchKey` + `SetGeometryShard` | no — Share never calls it |
| five shard preconditions / refusals | no — all return immediately when unsharded |
| `recentreEnabled_()` | behaviourally identical when unsharded |

The only thing running unconditionally on the normal path is recording one boolean at open. Conway's own regression and visual diff were clean across all twelve review rounds of #536, consistent with that.

### So what should this smoke show?

**The same as the 1.535.1506 round.** The M3 work that actually moves Share — incremental delta capture and the 64 MB residency budget — was already in 1506. This bump keeps the release candidate current with conway main; it is not expected to change load times, memory, or pixels.

If it *does* differ from the 1506 round, that's a finding worth reporting, because nothing in the diff should be able to cause it.

### What #536 buys, and why Share can't have it yet

Measured on the worker pool: **PSB geometry 40.5 s → 7.4 s at N=4** (2.34× wall-clock on D3D, 1.76× on MB-Khaya). The bottleneck moves to parse — 15.2 s of a 29.1 s load — which makes construct-from-columns the next lever rather than an optimisation.

Two things gate Share from using it, both now explicit errors in conway rather than silent misbehaviour:

1. **`SetGeometryShard` refuses `COORDINATE_TO_ORIGIN`**, which is exactly how Share opens. Each shard derives its own recentre anchor from whichever product it extracts first, so a model spanning more than one recentre cell merges with subsets shifted by whole grid cells. Needs an anchor agreed *before* the split.
2. **Workers need the OPFS store, not a resident buffer each** — four workers × 860 MB is fine on a 12 GB box and fatal in a tab.

Neither is in #536, and both are named in its description as the preconditions for the next step.

---

pablo-mayrgundter:
## Head re-pinned to conway 1.588.1550 — read this alongside the smoke checklist

`00de01a6`. The checklist above is still correct on what to look at, but two of its framing statements are now out of date. Rather than rewrite it mid-smoke, here is the delta.

### The conway-geom baselines are blessed now

The checklist says "the baselines are unblessed (R3)" and calls three unblessed geometry rounds "the largest unverified surface this PR has carried at any point". **That is no longer true.** `rc-1.588.1550` was tagged at [`d75c2f03`](https://github.com/bldrs-ai/conway/commit/d75c2f0337f2588df2282ca760dff4bdde5aff14) and the full-corpus pass ran on both corpora ([run 32884504977](https://github.com/bldrs-ai/conway/actions/runs/32884504977)). R1, R2 and R3 are blessed together in that pass, as planned.

Public bless: [test-models#58](https://github.com/bldrs-ai/test-models/pull/58) — 33 baselines, no new parse/extract failures, no unexpected zero-geometry models, and `benchmarks/conway1.588.1550-ci_test-models/` carries the perf snapshot. The churn is dominated by `Arty_Z7.csv` at **1171 rows**, which is #581's deflection floor and matches its own prediction of "1,169 of 94,510 rows across 3 models". Everything else is 1–34 rows.

So the Arty_Z7 "look at it, not just the clock" item is **still the right thing to do** — a tessellation-density change is exactly the sort of thing a digest bless records without judging — but it is no longer being judged against nothing.

### Two additions since 1.564.1548

Neither touches geometry.

- **[#586](https://github.com/bldrs-ai/conway/pull/586)** — the per-item error log on prefixes is gated, and `DanglingReferenceError` stops reporting recoverable prefix misses as errors. Console noise only.
- **[#588](https://github.com/bldrs-ai/conway/pull/588)** — **STEP NavTree no longer dies on a product with no name.** `resolveLabel()` ended on `productDef.name ?? ''`, but `name` is not an attribute of `product_definition` in AP214 — it is DERIVED onto a getter that throws, so the `?? ''` was unreachable and `getSpatialStructure` failed for the whole model. It fired only on a single-part file with no occurrence name, i.e. the NIST PMI set. Geometry was unaffected, which is why it showed as a red console error plus a NavTree collapsed to `AuxScene > Object` while the model rendered fine — and why a reload looked normal, since the GLB cache hit and the STEP path never re-ran.

  **Worth adding to the smoke:** load `nist_ctc_04_asme1_rd.stp` and check the NavTree shows `NIST PMI CTC 04 ASME1` rather than `Object`, with no console error. Four of the six affected corpus models get a real name; two get `''` because the files carry nothing else at all.

### Unchanged

`Conway v1.588.1550` in the load log, not `1.564.1548`. Every other checklist item stands as written — SKYLARK250 responsiveness, the Geometry progress denominator and its two caveats, PSB holding its ⅓ drop, no cubes or plates, no `Preview:` line.

### Still open before this can merge

`mergeable_state` is `behind` — main has moved past this branch's base, so it needs main merged in, which will also want a fresh CI run at that point.

---

pablo-mayrgundter:
## Status of the four open review threads — all addressed, verified against the current head

Posting this as a comment rather than inline replies: a **pending review from 2026-08-15** (against `e9c352b8`, never submitted) is still open on this PR, and GitHub allows only one pending review per user, so inline replies are refused until it is submitted or discarded. Worth knowing before merge — it is also why the thread list still looks busier than it is.

Two of the four are already flagged outdated by GitHub. All four are fixed in the code as it stands.

### 1. `conwayDirectIfcLoader.js` — OPFS `.step`/`.stp` would fail after conway#510 ([r3790470156](https://github.com/bldrs-ai/Share/pull/1753#discussion_r3790470156))

**Fixed**, in the shape the comment suggested. `conwayDirectIfcLoader.js:168–178`:

```js
modelID = await ifcAPI.OpenModelStream(store, deferSettings)
if (typeof modelID !== 'number' || modelID < 0) {
  ...
  modelID = await ifcAPI.OpenModelStreamed(openData, deferSettings)
}
```

Same guard on the non-deferred path at `306–310`. A refusal materialises the Blob and retries the streamed/classic selection rather than failing the load. STEP reaches conway through this path today and loads — `Arty_Z7.stp` is the headline smoke model for this very pin.

### 2. `Loader.js` — LFS sniff was a no-op on the store-open path ([r3790470157](https://github.com/bldrs-ai/Share/pull/1753#discussion_r3790470157))

**Fixed.** `Loader.js:426–427` now sniffs unconditionally, before any `File` passthrough:

```js
const head = await file.slice(0, LFS_POINTER_PROBE_BYTES).arrayBuffer()
if (looksLikeLfsPointer(head)) {
```

The surviving `opfsEntryKey !== null` guard at `:437` is a **different branch and is deliberate**: it gates the *re-fetch recovery*, which needs the GitHub coordinates (`filePath`, `shaHash`, `owner`, `repo`, `branch`) to resolve the pointer. So detection is now universal and only recovery is GitHub-scoped — which is right, since an uploaded pointer file can be detected and reported but has nothing to be re-fetched from.

### 3. `Loader.js` — stale spill comment ([r3790470159](https://github.com/bldrs-ai/Share/pull/1753#discussion_r3790470159)) — *outdated*

**Fixed.** The comment no longer claims identity comes from "the same handle we `arrayBuffer()`'d for parse on a miss", which was false once a miss stopped materialising the source.

### 4. Codex P1 — preview resources detached without disposal ([r3799202358](https://github.com/bldrs-ai/Share/pull/1753#discussion_r3799202358)) — *outdated*

**Fixed**, and the diagnosis was correct: the dispose pass had been dropped *wholesale* when the shared imposter cube arrived, rather than made shared-aware — so it was a real GPU-memory leak on every load, not a false positive. Teardown now walks stray imposters plus the preview group and disposes each unique geometry/material exactly once (`c1812b3`, `09b943d`). The "exactly once" matters because the imposter geometry and material are pooled and shared across plates; a naive per-child dispose would double-free them.

---

Nothing in these four is outstanding. If the current codex round raises anything new I will work it under the same bar — push small confident fixes, escalate anything genuinely major.

---

pablo-mayrgundter:
## Codex round on `a177142c`: one P1, verified real, filed as #1768 rather than fixed here

Replying as a comment because the Aug-15 pending review still blocks inline replies.

### The finding is correct, and I checked rather than took it on trust

> **P1 — Preserve evicted geometry for the fallback build** ([r3856798979](https://github.com/bldrs-ai/Share/pull/1753#discussion_r3856798979))

The chain holds exactly as described:

- `captured` accumulates **raw `FlatMesh` references** (`conwayDirectIfcLoader.js:240`) — no copy happens in that file.
- `ShareIfcLoader.js:455` → `buildConwayIfcModel(captured, …)` → `flatMeshToBufferGeometry` → `api.GetGeometry(modelID, geomExpressID)` per placed geometry.
- `flatMeshToBufferGeometry:140` treats a null/empty return as `skippedPlacedGeometries++; continue` — and an evicted asset is precisely that.
- `GEOMETRY_BUDGET_MB = 64` evicts as the pump advances.

So the end-of-load fallback re-reads geometry the budget has already discarded, and emits a model missing every evicted shape — silently, since the skip only lands in `buildStats`. The `?feature=batchedMesh` fallback at `:445` reads through the same call and shares the exposure. And the models where eviction bites are the same ones most likely to stress the incremental builder into failing, so the two conditions correlate rather than being independent.

**Codex has now been right twice on this PR** — the earlier P1 about preview resources being detached without disposal was also a genuine leak, not a false positive.

### Why it is filed rather than fixed in this PR

Scope, not doubt:

- **The happy path never reaches it.** Nothing here runs unless `builder.finalize()` or an append has already thrown.
- Before this PR the fallback was sound because nothing evicted, so this is a **regression in recovery quality on an error path**, not a new failure on a working load.
- The right fix is codex's own first option — build the fallback from the copies the incremental builder already holds (#1640), instead of discarding them and re-reading from the engine. That is a real change to the recovery path and wants its own diff and its own test, not a late patch on a release PR.

Tracked as **#1768** with the full chain, three fix options ranked, and a repro sketch (force a throw from `finalize()` behind a test hook on a model large enough to evict, then diff `skippedPlacedGeometries` and triangle count against the same load with the budget unset). It also notes that whatever the fix, a fallback build with `skippedPlacedGeometries > 0` should stop being silent.

### Earlier threads

The four pre-existing threads are all addressed; details and line numbers in [the previous comment](https://github.com/bldrs-ai/Share/pull/1753#issuecomment-5415976110). Two are resolved, two were already flagged outdated by GitHub.

### CI on this head

`build` ✅, `playwright-run` ✅, `playwright-webifc-run` ✅ on `a177142c`. The three Netlify checks are `neutral`, which is their no-rules-changed result. That also closes the gap the description flags: `build-prod`, `test-ci` and Playwright had only run at the earlier head locally, and have now run here in CI.

---

Issue #1747, opened 2026-08-11
`yarn build-prod` rewrites tracked package.json, so the version stamp keeps leaking into commits


pablo-mayrgundter:
Surfaced during review of #1746, where it leaked twice — once in that PR (`231b5aa`, reverted by `efa81fe`) and previously in `32362cc`.

## What happens

`tools/updateVersion.mjs` runs as part of `yarn build-prod` and rewrites the `version` field of the tracked `package.json` in place:

```diff
-  "version": "1.1743.6caa6b3",
+  "version": "1.1745.198f7d4",
```

The stamp is derived from the local checkout, so the numbers are wrong on any branch: `1745` is the last *merged* PR rather than the current one, and the sha is a sandbox commit. Anyone who runs `build-prod` locally — to check the esbuild path, to reproduce a CI failure, to look at a bundle — leaves a modified tracked file behind, and `git add -A` then carries it into the commit.

The pre-commit hook can't catch it: the file is legitimately modified from git's point of view, and there's nothing to lint.

## Why it's about to get worse

#1746 gates Share's CI `build` job on PR draft state, so a draft gets no `build-prod` signal from CI at all. `AGENTS.md` therefore now tells people to run `yarn build-prod` once before flipping a PR to ready — which makes an occasional footgun into a step in the documented workflow. The doc has a `git checkout -- package.json` chaser, but that's a manual cleanup step guarding a tracked file, which is exactly the shape of thing that gets skipped.

## Options

1. **Have `updateVersion.mjs` restore `package.json` when it's done**, writing the stamped value only into the build output. Smallest change; keeps every existing consumer working.
2. **Write the version to an untracked generated file** (e.g. `src/version.generated.js`, gitignored) and import it, leaving `package.json` alone entirely. Cleanest, but touches every reader of the version.
3. **Only stamp when `CI=true`**, leaving local builds unstamped. Simple, but local bundles then carry a placeholder version, which is misleading in the console banner.

Worth checking against the Sentry release tag before picking — the stamp is what ties a deployed bundle to its release, so whatever replaces it has to reach the Sentry config the same way.

PR #566, opened 2023-01-27
Icons cleanup

OlegMoshkovich:
This PR deletes unused icons.

Issue #563, opened 2023-01-26
Bug:  Notes:  Navigation is broken


OlegMoshkovich:
Step to reproduce:
- Select a note
- Try to navigate to the next note using navbar arrows
- Get NaN

This bug was introduced in one of the recent PRs.
Investigate the test cases for the notes.

OlegMoshkovich:
fixed by #566

PR #1756, opened 2026-08-19
Geometry worker pool: IFC extraction in N workers, sharded by product

pablo-mayrgundter:
Share half of conway's M3 — items 4 and 5 of the [#394 handoff plan](https://github.com/bldrs-ai/conway/issues/394#issuecomment-5343844967).

**No longer stacked.** #1753 has landed on `main`, so this PR is now based on **`main`** directly and `origin/main` is merged in at `5e4f9b97` (`main` is a full ancestor of this head). Two earlier revisions of this description were stale in the other direction — one said the branch was *at* 1.538.1511, another said #1753's 1.578.1546 was not merged here yet. Both are superseded by the line below.

**Conway pin: this branch carries `1.588.1550`**, which is simply what `main` carries — `package.json` here is now byte-identical to `main`. The pool's own requirement is only **≥ 1.538.1511**, the version that makes it legal on Share's open settings, per "What conway had to land first" below; the pin is a separate number and has moved well past it.

**The gate has not been re-run at 1.588.1550.** Everything under "Verification" was recorded at 1.560.1539. The forward merge crossed conway #563/#566/#567/#568/#569/#570/#578 and beyond — **#566 in particular rewrites the aggregate prefetch on the windowed path, which is the path this pool's shards page through**. Re-run `yarn build-prod`, `yarn test-ci` and the Playwright specs before flipping to ready.

Behind `?feature=workers`, off by default.

## What it does

Cache-miss IFC geometry extraction moves into N Web Workers, each with its own conway instance and wasm heap, each pumping a **disjoint** shard of the model's products. The main thread keeps its own instance for everything that is not geometry — the parse, properties, the spatial tree, the preview channel, and the coordination frame the workers are handed.

Measured in node at **40.5 s → 7.4 s on PSB at N=4**. This PR does not claim a browser number: `index.ifc` has seven products, so the E2E proves the machinery, not the speedup.

## What conway had to land first

Both blockers were real and both are now lifted by [conway#538](https://github.com/bldrs-ai/conway/pull/538) (merged, published as 1.538.1511):

- **B1 — residency-independent dispatch keys.** The key deciding shard ownership walks attribute records; on a windowed source — the only kind Share opens — whether a hop resolves depended on which chunks *that worker* held, so workers disagreed and a product was extracted twice or dropped.
- **B2 — `SetCoordinationFrame`.** `COORDINATE_TO_ORIGIN` anchors on the first geometry an instance captures, so N workers derived N frames. The main thread derives it once and hands the same matrix to every worker.

**B4 dissolved rather than being solved.** An OPFS `createSyncAccessHandle` is exclusive per file, which read as a blocker for a pool. It isn't one: `makeBlobByteStore` reads through `blob.slice()`, and a `File` is structured-cloneable, so the same handle goes to every worker. Off the OPFS path the resident bytes are wrapped in a `Blob` once — cloned by reference, so N workers share **one** copy — which is what lets the pool serve an ordinary first load rather than only a cache hit.

## Design notes worth reading in review

**The frame has to be derived before it is handed out.** A deferred model's frame is anchored on the *first geometry it captures*, so until something is pumped `GetAppliedCoordinationMatrix` returns **identity** — and a supplied frame *suppresses* the one each worker would otherwise derive. The frame carries the Z-up → Y-up normalize, so supplying identity renders the whole model 90° out. The pool therefore pumps products on the main thread until the frame actually anchors, purely to derive it; that lands on the same product a single-threaded load anchors on, so the pooled frame doesn't merely agree across workers, it *matches*. (This shipped broken in the first revision and smoke caught it — see "How this was found" below. It then shipped broken a *second* way, fixed in `37039dc` for #1763: `extracted` counts products, and a product with no representation anchors nothing, so a model that leads with containers reported identity.)

**Two engines, deliberately.** `geometryApi` resolves vertices; `ifcAPI` is the model's own engine and is what property and spatial closures bind to. They are the same object on every path but this one. Passing the adapter where decoration was expected is exactly how an earlier revision broke (`Cannot read properties of undefined (reading 'getSpatialStructure')`), so the split is now explicit at each call site.

**Determinism is the subtle part.** `incrementalBatchedBuilder` fixes the model-wide origin-recenter offset from the *first placement it is handed*, rounded to whole metres. Across N workers that would be whichever shard won the race — so a georeferenced model would land on a different offset run to run, and every saved camera against it would be wrong by the difference. The pool queues other shards until shard 0 has spoken, delivers shard 0's batch first, and drains lowest-shard-first if shard 0 turns out to own nothing.

**Failure is not symmetric.** Failing *before* delivering anything falls back to the main-thread pump, which is a complete recovery — nothing was extracted here yet. Failing *part-way* cannot silently fall back: the partial scene came from payloads the pump cannot resolve, and appending pump batches to the same builder would draw those products twice from two engines. So the loader drops the partial group via `onGeometryReset` first, and fails loudly if no reset hook was supplied. A clean run that delivers **nothing** is treated as a miss for the same reason.

## The load report had to be fixed too

Two things went stale under a pool. Both are fixed rather than documented around:

- **Progress.** Reporting once on completion collapsed the Geometry line to `0.007s, +0.000000 MB heap` — the phase began and ended in one event — and left the reporter's 30-second stall watchdog with nothing to hear during the longest phase of a large load. Now reported per batch, in **products**, summing each shard's `extracted`/`remaining`.
- **Memory.** The report's heap comes from `performance.memory` on the **main thread**, so moving extraction into workers moves those allocations somewhere it cannot see — the Geometry line would show a large improvement while real process memory went *up* by N wasm heaps. Each worker reports its own and the summary line carries the sum (`wasmHeapMb=`). Kept separate rather than folded into the report's heap column: summing a main-thread sample with N worker samples would invent a number neither engine measured.

Also: an unrecognised `?feature=` value was silently ignored, so `?feature=worker4` (singular) reads exactly like `workers4` and quietly produces a baseline run — a benchmark you then believe. Unknown values now warn once, naming them. Deliberately **not** auto-corrected to the nearest flag: that turns a typo into a different silent behaviour.

## How this was found, and what it says about the tests

The rotation shipped in the first revision and survived the whole suite, because vertices, triangles, instances and placement counts are all **rotation-invariant** — the E2E's "identical to a single-threaded load" comparison passed on a model lying on its side. A human loading the page caught it.

The spec now asserts the **frame itself** is not identity, and the summary line prints it, precisely because supplying the wrong frame leaves every other number on that line unchanged. Verified against a mutated build: both form factors fail without the fix.

Same pattern earlier in this PR's history: the first version of the E2E spec *failed*, which is how the store-path-only limitation was found rather than shipped.

## Verification

Recorded at conway **1.560.1539**. **Not re-run at this branch's current 1.588.1550** — see the note at the top.

- **2414 Jest tests / 244 suites** (5 skipped), lint + typecheck clean via the pre-commit gate. `test-ci` coverage gate and `build-prod` passed at the head these were first recorded on.
- **E2E on desktop *and* mobile** (`geometryWorkers.spec.ts`, via `describeMobileAndDesktop`): the pool engages at N=2 in a real browser, supplies a non-identity frame, and assembles identical build stats to a single-threaded load.
- `?feature=workers4` renders **pixel-identically** to a baseline load (screenshot-compared locally at a fixed camera).
- Unit tests pin the merge ordering, the progress aggregation, the memory sum and the adapter seam; the ordering tests fail with the shard-0 gate removed.

**`package.json` carries no build stamp.** `617e3f6f` had committed `version: 1.1755.8b1f229a` — the [#1747](https://github.com/bldrs-ai/Share/issues/1747) trap, written by `tools/updateVersion.mjs` on every build. Reverted in `016f7654`; `package.json` here is now identical to `main`.

## Flags

| flag | effect |
|---|---|
| `?feature=workers` | pool sized from `hardwareConcurrency − 2`, capped at 8 |
| `?feature=workers1` … `workers8` | pin the count — how N is compared against N=1 on one machine |

`workers1` is the plain move-to-worker baseline (item 4): same products, same order, no sharding. The bare `?feature=workers` **declines** below `MIN_POOL_SOURCE_MB` (16 MB) — standing the pool up costs 0.7–1.6 s, which a small model cannot pay back. A pinned `workersN` ignores that gate, so small-model benchmarks still exercise the machinery.

## Smoke

**Deploy previews are live** — now that this is based on `main`, Netlify builds them:

- **prod:** https://deploy-preview-1756--bldrs-share-prod.netlify.app
- **dev:** https://deploy-preview-1756--bldrs-share-dev.netlify.app

(An earlier revision of this section said there was no preview and offered to rebase onto `main`. That is done, and the rebase no longer costs anything: `main` now carries conway 1.588.1550, so nothing is dropped by smoking against it.)

Load PSB and compare:

1. `?feature=workers1` against no flag — does moving extraction off the main thread cost or save?
2. `?feature=workers4` against `?feature=workers1` — does it scale?
3. Visually identical in all three, and `[conwayDirect] parsed` stats identical.
4. The console line to read:

```
[conwayDirect] geometry workers: n=4 placements=… geometries=… wasmHeapMb=… frame=[…]
```

If that line is absent, the pool declined and the load fell back — report that (on a model under 16 MB with the bare flag, declining is the intended behaviour). If `frame=` comes back as identity, the recentre was suppressed and the model will be rotated. And read `wasmHeapMb=` *alongside* the report's heap column, not instead of it — the report still only sees the main thread.

**Known caveat on the browser numbers.** The first real browser run of the pool on PSB was **3× slower**, not faster (`8b1f229a`): every worker parses the whole model before it can extract anything, so parse contention dominates. The node speedup ran over resident in-memory buffers where redundant parses are cheap. The unblock is [conway#541](https://github.com/bldrs-ai/conway/issues/541) — an `IfcAPI` open that consumes a prebuilt index. The flag stays off until then.

## Carries one workaround

conway's `isWebPlatform()` has no Worker branch, so it loads the Node wasm and `Init()` dies. Filed as [conway#540](https://github.com/bldrs-ai/conway/issues/540); the worker declares the platform itself until that lands. No build define fixes it — the guard is `typeof process !== 'undefined'`, which no bundler `define` supplies.

Design doc: [design/new/geometry-workers.md](design/new/geometry-workers.md), routed from AGENTS.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -101,6 +101,7 @@ This file is the router for AI assistants working in this repo. Keep it small. T
 | AdSense / ads strategy, route policy, test-hermeticity rules | [design/new/ads.md](design/new/ads.md) |
 | Usage quotas: tiers, 30-day rolling window, `record-load` server gate, GitHub privacy detection, OPFS local fallback, `quotas` feature flag | [design/new/quotas.md](design/new/quotas.md) |
 | Conway-direct IFC pipeline, IfcInstanceMap, per-instance picking, `?feature=conwayDirectIfc` | [design/new/viewer-replacement.md](design/new/viewer-replacement.md) §3b |
+| Geometry worker pool — extraction in N workers with sharded products, `?feature=workers`, why the origin-recenter offset has to be gated on shard 0, reading one OPFS file from N workers | [design/new/geometry-workers.md](design/new/geometry-workers.md) |
 | STEP occurrence-keyed selection (NavTree↔scene per-occurrence), `PlacedGeometry.occurrencePath`, why one nut highlights all | [design/new/step-occurrence-selection.md](design/new/step-occurrence-selection.md) |
 | The `?feature=look` render look — PBR materials + gradient IBL + tone-mapping, Neutral/Flat toggle, `looks.js`/`lookMaterial.js`, `LightingGui`, why it's all behind one flag (default off) | [design/new/viewer-replacement.md](design/new/viewer-replacement.md) §6e |
 | Removing the `conway-web-ifc-adapter` shim, the `web-ifc` engine seam (`webIfcShimAlias`/`USE_WEBIFC_SHIM`), Conway version-lag, runtime engine swap | [design/new/adapter-removal.md](design/new/adapter-removal.md) |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,11 +27,16 @@ This file is the router for AI assistants working in this repo. Keep it small. T
   in flight and CI runners are capped at 4 concurrent jobs, so a PR that
   runs the full suite on every rework push starves the PRs that are
   ready. (1) Open it with `draft: true` — not opened-then-marked.
-  (2) Keep it in draft until `/review` has run and every finding is
-  fixed or answered in the thread. (3) Flip to ready
+  (2) Keep it in draft until review has run and every finding is
+  fixed or answered in the thread — who reviews, the timeout fallback,
+  the round cap, and which PRs skip review entirely are all specified
+  in [design/new/agent-workflow.md](design/new/agent-workflow.md)
+  §"Review", which is the single source for review policy. Read it
+  before requesting or skipping a round; don't restate its rules here.
+  (3) Flip to ready
   (`update_pull_request`, `draft: false`); that fires
   `ready_for_review`, which is what starts the gated jobs. (4) Drive CI
-  green, and `/review` again if the fixes changed the diff beyond a
+  green, and re-request review if the fixes changed the diff beyond a
   trivial revert — otherwise the reviewed diff isn't the merged diff.
   (5) Update the PR description to match what the change became, merge,
   then close or narrow its issues (partly-addressed → a comment saying
@@ -39,10 +44,12 @@ This file is the router for AI assistants working in this repo. Keep it small. T
   **What the gate covers:** `build` (main.yml) and both `test-flows`
   jobs are skipped on drafts. **Netlify deploy previews still run** —
   they come from Netlify's GitHub integration, not Actions — so a draft
-  normally has a preview URL to click through. (Not for a docs-only PR:
-  `tools/netlify/ignore-build.sh` skips the deploy when every changed
-  file matches `.md` / `design/` / `notes/`. Marketing posts are `.mdx`,
-  so they still build.) Before editing those
+  normally has a preview URL to click through. (`tools/netlify/ignore-build.sh`
+  skips the deploy when every changed file matches `.md` / `design/` /
+  `notes/` — but only from a branch's *second* push onward, since it
+  bails to build when `CACHED_COMMIT_REF` is unset and a new branch has
+  no prior deploy, so the push that opens a docs-only PR still builds
+  (#1766). Marketing posts are `.mdx`, so they always build.) Before editing those
   workflows: the gate is a job-level `if:` (a skipped-by-if job
   satisfies a required check; a workflow that never triggers leaves the
   PR waiting forever); `ready_for_review` and `converted_to_draft` must
@@ -55,14 +62,14 @@ This file is the router for AI assistants working in this repo. Keep it small. T
   **Consequence to be aware of:** because every job is gated here, a
   draft gets no CI signal at all, and the husky pre-commit hook covers
   only part of the gap. Three things it does not run, each of which will
-  otherwise first fail at step 3 — after `/review` has signed off, which
+  otherwise first fail at step 3 — after review has signed off, which
   is exactly the step-4 re-review this lifecycle is trying to avoid:
   (a) `yarn build-prod`, so an esbuild-only breakage (missing asset
   loader, plugin misconfig) survives every draft commit; (b) coverage —
   the hook's `yarn test` runs `test-src`, while CI runs `test-ci` →
   `test-coverage` with thresholds enforced (`tools/jest/jest.config.js`);
   (c) **Playwright**, the big one — E2E specs never execute before step
-  3, so `/review` is reading specs that have never run, against this
+  3, so the reviewer is reading specs that have never run, against this
   file's own mandatory desktop+mobile E2E rule. Before flipping to ready,
   run `yarn build-prod`, `yarn test-ci`, and
   `yarn test-flows-build-and-serve` + `yarn test-flows [spec]` — then undo
@@ -110,7 +117,7 @@ This file is the router for AI assistants working in this repo. Keep it small. T
 | Epic/Story/Track catalogue, milestone tier rubric (§2.1), MVP bar + phase plan (§6, ex-"Pro-MVP"), growth-funnel Phase G, AI-workspace pivot (§7), post-MVP loveables | [design/roadmap.md](design/roadmap.md) |
 | Conversational-CAD epic plan (workspace shell / ProjectsDrawer + TopBar, fluid Nav+search, convo panel, multi-user channels), wireframe→issue mapping, epic/sub-issue process | [design/new/conversational-cad.md](design/new/conversational-cad.md) |
 | Persistence direction: OPFS as a git-versioned workspace (repo-as-workspace, LFS-pointer blobs, record-vs-stream convo split, wasm-git vs isomorphic-git investigation, exit plan) | [design/new/workspace-store.md](design/new/workspace-store.md) |
-| Working an issue queue with AI agents — coordinator + per-issue sub-agents, model tiers (Fable coordinates, Opus executes), the dispatch-brief requirements and the 10-point rubric | [design/new/agent-workflow.md](design/new/agent-workflow.md) |
+| Working an issue queue with AI agents — coordinator + per-issue sub-agents, model tiers (Fable coordinates, Opus executes), the dispatch-brief requirements, the 10-point rubric, and the review lifecycle — the single source for who reviews a PR, when review is skipped, and how rounds are capped | [design/new/agent-workflow.md](design/new/agent-workflow.md) |
 
 Anything not in this table is invisible to the router. When you create a doc that future assistants should consult, add a row above with a one-line "when to read" hint. Don't rely on filesystem discovery.
 
--- a/design/new/agent-workflow.md
+++ b/design/new/agent-workflow.md
@@ -77,3 +77,44 @@ Reviewers hold the same bar, in this order: verify claims against
 the diff (not the description), against repo history, then against
 the actual failure path — and grade findings by evidence, not
 plausibility.
+
+## Review: codex first, sub-agent on timeout
+
+Sub-agents do not review their own code, and the coordinator does not
+review theirs. Review comes from **codex** — usually automatic;
+otherwise request it with an `@codex review` comment on the PR.
+
+**Docs-only changes skip review by default.** There is no code to
+attack, and a review round on a paragraph costs more than it finds.
+Make an exception where the doc *is* the policy — a change to how we
+review, dispatch or release is worth having codex read, precisely
+because it is the thing codex will be held to.
+
+Three rules keep review from becoming either a bottleneck or a
+rubber stamp.
+
+**Timeout.** If codex has not responded ~10 minutes after the
+request, dispatch a **sub-agent review** and treat that as the round
+rather than waiting. A late codex finding still counts — fold it in
+when it arrives, even if a sub-agent round has already run.
+
+**A substituting reviewer needs to be pointed at the hazards.** A
+generic "review this diff" comes back clean on exactly the changes
+that most need scrutiny, because the risky part of a good fix is
+usually an invariant the diff does not mention. Hand the reviewer the
+issue as well as the diff, and name the specific claims to attack —
+the order-preservation argument behind a zero-digest-churn claim, the
+state that has to stay stable across a demand pump, whatever the
+change is actually betting on. A clean review that never engaged with
+the load-bearing claim has not reviewed it.
+
+**Cap the rounds.** A few rounds, not an open-ended dialogue. If
+findings are still arriving after ~3, or the review turns into a long
+back-and-forth, pause it and escalate to the coordinator — that
+pattern usually means the change needs a design decision, not more
+review turns.
+
+One caution from experience: codex has reversed itself on an
+identical commit more than once, clean on one pass and not on the
+next. A single clean pass is not by itself a merge signal. Read what
+it said.
--- a/design/new/geometry-workers.md
+++ b/design/new/geometry-workers.md
@@ -209,6 +209,30 @@ The node spike ran 4 processes on an idle box over **resident in-memory
 buffers**. Redundant parses were cheap there. In a browser, over OPFS, with
 six wasm heaps competing for bandwidth, they are not.
 
+## Worker startup is cheap — it is not what you are waiting for
+
+Measured on `index.ifc` (18 KB, 7 products), so parse and extract are ~0 and
+what remains IS the cost of standing workers up:
+
+| N | last worker ready | conway `Init()` | model open | phase total |
+|---|---|---|---|---|
+| 1 | 780 ms | 82 ms | 82 ms | 0.75 s |
+| 2 | 894 ms | ~100 ms | ~57 ms | 0.93 s |
+| 4 | 1 329 ms | 77–122 ms | ~38 ms | 1.35 s |
+| 6 | 1 751 ms | 131–221 ms | ~44 ms | 1.77 s |
+
+Roughly **0.7 s for the first worker and ~0.2 s for each after** — about 2 s at
+N=6, fixed, regardless of model size.
+
+The decomposition matters more than the total. wasm `Init()` is **80–220 ms**
+and the model open is **tens of ms**; nearly all of the rest is fetching and
+compiling the worker bundle, which is 5.5 MB because it carries all of conway.
+Trimming that bundle to the extraction path is the lever if startup ever needs
+to be faster — not the wasm.
+
+So the ~20 s gap a user sees between parse and first geometry on PSB is **not**
+worker startup. It is the redundant parse, and startup is ~10 % of it at N=6.
+
 ## The ceiling, stated honestly
 
 Even with a free shared index, PSB is bounded by Amdahl — parse is not
@@ -221,6 +245,41 @@ prep 6 + parse 15.8 + geometry 9.5 + assemble 5.3  ≈  37 s   vs   52.9 s
 So **~30 %**, not the 5× the geometry-phase headline suggests. That is the
 same observation conway#536 made when it noted the bottleneck had moved to
 parse. Construct-from-columns is the lever that matters after this one.
+
+## Parallelising the parse
+
+The geometry pool alone caps at ~30 % because parse is serial. Sharding the
+parse too is what makes the pipeline worth rebuilding, and it looks tractable:
+a STEP parse is a scan over `#N=…;` records, so N workers can each tokenise a
+disjoint byte range into partial index columns. conway already emits SoA
+columns in 64 K-row segments (`ColumnarIndexSink`), which is a friendly shape
+to concatenate.
+
+The parts that need care, none of them blocking:
+
+- **Local IDs are parse-order**, and parse order is byte order — so shard
+  columns concatenate in address order provided each shard reports its record
+  count so bases can be assigned at merge.
+- **Inline entities** unfold into a tail range, and the express-ID lookup
+  table is built from the merged columns. Both are post-merge steps.
+- **Shard boundaries** must land between records, and a STEP string literal
+  can contain `;` — so the boundary scan has to be tokenizer-aware rather
+  than a naive search. This is the genuinely fiddly bit.
+
+The payoff compounds with the geometry pool. PSB parse is 15.8 s for 860 MB
+(~54 MB/s), which is CPU-bound tokenising, so even 3× gives:
+
+```
+prep 6 + parse 5.3 + geometry 9.5 + assemble 5.3  ≈  26 s   vs   52.9 s
+```
+
+roughly **2×**, against the geometry pool's ~30 % alone.
+
+Note this does **not** remove the need for conway#541. A worker that parsed
+one byte range holds a partial index, and extraction needs the whole model —
+a product references records anywhere. So the shape is: parse shards → merge
+columns once → hand the merged columns to every worker → extract shards. The
+distribution step is conway#541 either way; parse sharding feeds it.
 
 **The unblock is conway#541**: an `IfcAPI` entry that opens from a prebuilt
 index. The serialisation half already exists (M4a's `index_sidecar.ts`, whose
--- a/design/new/geometry-workers.md
+++ b/design/new/geometry-workers.md
@@ -88,6 +88,31 @@ the resident bytes are wrapped in a `Blob` once. A Blob clones by reference,
 so N workers share **one** copy rather than N. Without this the pool would
 only ever engage on a cache hit, which is not the load a user waits for.
 
+## The frame has to be derived before it is handed out
+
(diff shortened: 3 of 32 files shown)


pablo-mayrgundter:
## Frame derivation fix + the test-gap closure for #1761

Full investigation and evidence in [#1761](https://github.com/bldrs-ai/Share/issues/1761#issuecomment-5374532212). Summary of what changes here.

### The bug

The one-product seed that derives the recentre frame is not sufficient:

```js
await ifcAPI.ExtractGeometryBatchAsync(modelID, 1, (flatMesh) => seedBatch.push(flatMesh))
const coordination = ifcAPI.GetAppliedCoordinationMatrix(modelID)
```

`extracted` counts **products**, not geometry. A product with no representation anchors nothing, so on an assembly-first export the seed reads conway's not-yet-anchored **identity** and supplies it — which `SetCoordinationFrame` then *applies*, suppressing the recentre and the Z-up → Y-up normalize with it. The model renders 90° out.

On `171210AISC_Sculpture_param.ifc` (already in this repo) the frame does not anchor until the **50th** product; the first 49 are `IfcElementAssembly` containers. This is the same defect class as the rotation this PR already shipped once, reached through the case the one-product seed does not cover.

It is invisible to every existing assertion for the reason the PR description itself gives: rotation leaves vertices, triangles, instances and placements unchanged. Verified directly — identity vs the derived frame changes geometry totals by exactly 0.

### The fix

New exported `deriveCoordinationFrame(ifcAPI, modelID, seedBatch)` in `conwayDirectIfcLoader.js`: pump one product at a time until `GetAppliedCoordinationMatrix` stops reporting identity, then stop. That costs the geometry-less prefix plus one product — and it lands on exactly the product a single-threaded load anchors on, so the pooled frame doesn't merely agree across workers, it matches the flag-off frame.

One product at a time is deliberate: a larger seed batch would pump geometry-bearing products *past* the anchor, and every one of those is main-thread work the pool exists to move off the main thread.

If a model never anchors, it returns `null` and **the pool declines** rather than supplying identity — falling back to the main-thread pump, which derives its own frame as it goes. Supplying identity is the one thing that must not happen. Bounded by iterations, not by products extracted, so an engine reporting work it never extracts cannot spin the loop.

Verified against the real engine at conway 1.543.1513 — the pooled frame now equals the single-threaded frame on `171210AISC_Sculpture_param.ifc`, `Momentum.ifc`, `SEESTRASSE.ifc` and `index.ifc` (it was wrong on the first, right on the other three, which is why nothing caught it).

### The test gap

The E2E ran on `index.ifc` — 7 products, and its product #1 has geometry, so the one-product seed anchors correctly there. It could not fail for this class.

- **`conwayDirectIfcLoader.coordinationFrame.test.js`** (new, 5 tests) drives `deriveCoordinationFrame` with fake engines that anchor at different depths. **Mutation-checked**: reverted to the one-product seed, 3 of the 5 fail, including `keeps pumping past products that emit no geometry`.
- **`geometryWorkers.spec.ts`** gains two cases on the AISC fixture — 296 products, so four shards own ~74 each instead of ~2 — via `describeMobileAndDesktop`. One asserts the frame is non-identity on a model that leads with geometry-less products. The other compares the **emitted `expressID` set**, read off the assembled model's `instanceParents` rather than off a log line, because identity is the point: a rotated model, and a dropped product cancelling a doubled one, both leave the totals intact.

### Not landed yet — working-tree collision

Two agents are live in the **same working tree** (`/home/user/Share`), not just the same branch. `conwayDirectIfcLoader.js` currently holds both my `deriveCoordinationFrame` work and the in-flight `geometryWorkerCount(sourceBytes)` change from the #1760 side, and `conwayGeometryPool.js` / `FeatureFlags.js` / `geometry-workers.md` are being edited concurrently. Committing that file would carry someone else's half-finished cross-file change under my authorship, so I have stopped short of committing and held the change as a patch instead. Flagging rather than racing — say how you want it sequenced.

All 19 `src/viewer/ifc` Jest suites (252 tests) pass on the combined tree; lint clean on the touched files.

---

pablo-mayrgundter:
Filed the defect this fixes as its own issue: **#1763** — the pool supplies an identity coordination frame on assembly-first models, rendering them 90° out. Splitting it out of #1761 because they are different bugs: #1763 is understood and fixed here; #1761's 741-vertex deficit is **still unreproduced** and stays open.

Landing the fix from a separate git worktree rather than `/home/user/Share`, since the #1760 agent is live in that tree (it is currently running `yarn test-flows-build-and-serve` there, and `test-flows-build` starts with `yarn clean`). The `geometryWorkerCount(sourceBytes)` hunk from that work has been stripped from my patch — this commit carries only `deriveCoordinationFrame` and its tests.

---

pablo-mayrgundter:
Two things from working Share#1760 on this branch (`5c18fd4`), one of which clears a question hanging over the PR and one of which changes a default.

## The flag-off regression reported against this PR does not exist

Quantified on this branch's head against its base #1753 (`3b22a3a`), cold `index.ifc`, fresh browser context per run, 10 runs each, headless Chromium:

| build | Total (median) | engine stages (median) |
|---|---|---|
| #1753 `3b22a3a` | 0.772 s | 0.352 s |
| **#1756 `f0af733`, pool OFF** | **0.692 s** | 0.345 s |

Inside the spread of either build. Checked against the built artifact rather than the imports, as the issue asked: the main bundle grows by **6 154 bytes** (15 375 362 → 15 381 516), which is `conwayGeometryPool.js` + `workerGeometryApi.js` and no engine. The worker's own 5.78 MB bundle is not in it and neither is a second copy of the conway wasm — `new URL(…, import.meta.url)` stays a runtime URL through esbuild, which is why the worker is its own entry point in `tools/esbuild/build.js`. Full numbers and the other two hypotheses in Share#1760.

Two guards so it stays true, both mutation-checked:

- `geometryWorkerBundle.test.js` walks the module graph esbuild resolves from `conwayDirectIfcLoader.js` and fails if any `*.worker.js` is in it;
- `conwayDirectIfcLoader.test.js` runs a real flag-off parse with `global.Worker` spied and asserts it is never constructed, and that `geometryApi` is the model's own engine.

## `?feature=workers` now declines below 16 MB of source

Standing the pool up is charged before the first product is touched — measured here at 0.74 s at N=1, 1.0 s at N=2 and 1.4–1.6 s at N=4 on 4 cores — while `index.ifc` spends 0.035 s in geometry. So the bare flag now means *use the pool where it pays*, falls back to the main-thread pump below `MIN_POOL_SOURCE_MB`, and says so on the console, because `?feature=workers` with no pool summary after it is exactly what a pool that *failed* looks like, and this PR's own smoke instructions read it that way. On `index.ifc` that is 1.62–1.80 s → 0.65–0.69 s.

**A pinned count overrides the gate**, which is what keeps `geometryWorkers.spec.ts` meaningful: it pins N=2 and N=4 on small fixtures precisely because a tiny model makes the machinery cheap to exercise, and a gate that silently overrode the pin would turn every small-model benchmark into a baseline run you then believe — the same failure the unknown-`?feature=` warning in this PR exists to prevent.

The size gate and Share#1761's "decline rather than supply an identity frame" compose in the obvious order: the size gate answers first, before any seed product is pumped, so a small model never pays for the frame derivation either; the frame check then still guards every model that does get a pool, and it correctly ignores the pin, since supplying identity is a correctness bug rather than a performance trade.

## Also, for the fix list in Share#1760

The profile says the standup is **worker creation and bundle evaluation**, not the wasm: `init` (compile + instantiate) is 63–188 ms, the model open is tens of ms, and the remaining ~0.6 s per worker scales with N. Sharing one compiled `WebAssembly.Module` — the obvious-looking remedy — would therefore buy back at most that 150 ms. Trimming the worker bundle, and spawning the pool *during* the parse instead of after it (~15 s of cover on PSB), are the levers that match where the time goes. Neither is in this pass.

`ENVIRONMENT === unknown` is benign and does not select a slower emscripten bootstrap — verdict with the evidence in Share#1760.

## One thing noticed, deliberately not touched

`package.json`'s `version` on this branch is `1.1755.8b1f229a` against main's `1.1743.6caa6b3` — a build stamp that reached a commit, which #1747 says must not happen. Left alone rather than reverted here: another agent is pushing to this same branch, and a one-line revert of a generated field is the kind of change that turns into a conflict for no benefit. Worth a separate one-liner before this merges.

## Gate

`yarn test-ci` 244 suites / 2412 tests green with coverage thresholds; `build-prod`, `typecheck`, `eslint` clean. The geometry-worker E2E passes on desktop and mobile, including the four Share#1761 tests, plus a new `geometryWorkerThreshold.spec.ts` (4 passed) covering the gate and its pin override.

---

pablo-mayrgundter:
## Pushed: `37039dc`, on top of #1760's `5c18fd4`

Landed the #1763 fix from a separate git worktree, rebased onto the #1760 commit that arrived meanwhile. Three files, all mine — the `geometryWorkerCount(sourceBytes)` hunk from #1760 is not in this commit (it is in `5c18fd4`, where it belongs).

### Gate

| | result |
|---|---|
| `yarn build-prod` | ✅ exit 0 |
| `yarn test-ci` | ✅ 244 suites / 2412 tests (5 skipped), coverage thresholds met |
| pre-commit hook | ✅ eslint + typecheck + jest, all green |
| `git diff package.json` | ✅ clean — the `1.1755.…` build stamp was reverted before committing (#1747) |

### E2E, and the mutation check

All **8** geometry-worker cases pass on desktop and mobile — the 6 pre-existing plus my 2 new ones. Zero failures in `geometryWorkers.spec.ts`.

**Mutation check.** Reverted `deriveCoordinationFrame` to the one-product seed, rebuilt, re-ran the spec: **exactly the two new frame cases fail, desktop and mobile**, and the other six still pass — which is the point, since a rotation leaves every count untouched.

```
✘ geometryWorkers.spec.ts:204 [desktop] › anchors the frame on a model that leads with geometry-less products
✘ geometryWorkers.spec.ts:204 [mobile]  › anchors the frame on a model that leads with geometry-less products
  Error: expect(received).not.toEqual(expected)
  Expected: not [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, …]
  > 229 |     expect(frame).not.toEqual(IDENTITY_MAT4)
```

The mutated build really does log `frame=[1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1]` in a real browser. That is the bug, reproduced end-to-end, and the new test is what sees it.

Unit side: `conwayDirectIfcLoader.coordinationFrame.test.js`, 5 tests; against the one-product seed, 3 of the 5 fail including `keeps pumping past products that emit no geometry`.

### The suite baseline does not hold in this sandbox

The full run is **134 passed / 27 failed / 44 skipped** (205 tests = 197 + my 8), not the 151/2/44 baseline. That is not this change:

- The #1760 agent's independent run on the same branch gave **137 / 24 / 44**. **23 of my 27 failures are the same tests as theirs.** The 4 that differ (Apps drawer resize, Google Drive empty state, Quota Pro tier, Login-with-Google screenshot) are load-sensitive specs that also flipped between their two attempts.
- None of the failures are in `geometryWorkers.spec.ts`.
- And this diff is unreachable without the flag: `geometryWorkerCount()` returns 0 unless `?feature=workers`/`workersN` is set, and everything I touched sits inside `if (workerCount > 0)`. No spec without that flag can execute a line of it.

So the sandbox carries ~24 pre-existing E2E failures on this branch, on 4 CPUs where Playwright's 30s timeouts bite. Worth a look on real CI, but it is not from here.

One process note: the two runs overlapped early on and starved each other — the first ~15 minutes of the #1760 run were competing with my Jest suite. I stopped mine and let theirs finish, and it visibly sped up. Two full E2E suites do not fit on this box at once.

---

pablo-mayrgundter:
## Correction: the build stamp is still on the branch, and it is the whole of the `package.json` diff

Checked at head `37039dc` rather than taken from the diff stat, because it's my mistake and the thread now reads as though it were handled.

| | `version` | `@bldrs-ai/conway` |
|---|---|---|
| base `3b22a3a` | `1.1743.6caa6b3` | `1.543.1513` |
| head `37039dc` | **`1.1755.8b1f229a`** | `1.543.1513` |

Two things follow.

**It was not reverted.** The gate table above says "`git diff package.json` ✅ clean — the `1.1755.…` build stamp was reverted before committing". That's accurate about *that commit* — `37039dc` doesn't touch `package.json` — but the stamp is still on the branch. It came from **my** commit `8b1f229a` and has been sitting there since. Flagging so nobody merges on the strength of that line.

**The conway bump is already in base.** Base carries `1.543.1513` too, so the `+1/-1` this PR still shows on `package.json` is **entirely the stamp** — there is no dependency change left in it. The fix is to restore `version` to `1.1743.6caa6b3`, which makes `package.json` drop out of the diff completely.

Not pushing it. Two agents are committing to this branch and one has already flagged working-tree collisions in `/home/user/Share`; a one-line revert of a generated field is exactly the change that turns into a conflict for no benefit if it races something. Whoever is next in the tree can take it, or say the word and I'll do it from a separate worktree.

---

Separately, on the frame bug in [#1763](https://github.com/bldrs-ai/Share/issues/1763): that one is mine, and it's the **second** instance of the same defect class in this PR. The first revision supplied identity because it read the frame before pumping anything; I fixed that with a one-product seed and called it done. The one-product seed has the same hole — `extracted` counts products, not geometry — so an assembly-first export still gets identity. `deriveCoordinationFrame` closes it properly by pumping until the frame actually anchors. The lesson I should have taken the first time and didn't: the E2E's count-based comparison is rotation-invariant, so no assertion built on vertices, triangles, instances or placements can ever see this class. Asserting the frame itself is what was missing, and it's right that it's now asserted on a fixture that leads with geometry-less products rather than on `index.ifc`, whose product #1 has geometry and could never fail.

---

pablo-mayrgundter:
Closing out the doubt I raised in https://github.com/bldrs-ai/Share/pull/1756#issuecomment-5375193179: the stamp revert in `016f7654` is correct, and the caveat I attached does not apply.

I had warned that `1.1743.6caa6b3` was read off a base that had since moved twice, so it might no longer be the right restore value. It still is — `main` never moved off it. Verified at `016f7654`:

```
main                          "version": "1.1743.6caa6b3"
grok/bump-conway-1.509.1435   "version": "1.1743.6caa6b3"
PR head                       "version": "1.1743.6caa6b3"
```

And `package.json` at head now differs from `main` by exactly one line, the conway pin (`1.501.1426` → `1.560.1539`) — nothing else rode along with the revert. No further action needed on #1747 for this branch.

---

pablo-mayrgundter:
## Landing state, 2026-08-26 — burn-down checklist

Verified against the repo at the time of writing, not carried over from the description.

| | |
|---|---|
| head | `5e4f9b97434fa57aa9639b85e89b45829cd8f19c` ("Merge remote-tracking branch 'origin/main'", 2026-08-26T17:04:54Z) |
| base | `main` — **no longer stacked**; #1753 landed and the stack collapsed |
| state | open, **draft**, `mergeable_state: behind` |
| `main` tip | `f38ea4f48a3a0cf632434b1f7216660dc589d316` (#1771, 2026-08-26T19:21:11Z) |
| size | 10 commits, 18 files, +2881/−32 |
| reviews | **0** |
| conway pin | `1.588.1550` — identical to `main`; npm latest is **`1.603.1553`** |

### Done

- [x] Pool is **correct**: exact partition on PSB — `placements=23454`, `geometries=20178`, zero cross-shard duplication (measured in [conway#541](https://github.com/bldrs-ai/conway/issues/541)).
- [x] Retargeted off `grok/bump-conway-1.509.1435` onto `main`, so Netlify builds previews again:
  - prod https://deploy-preview-1756--bldrs-share-prod.netlify.app
  - dev https://deploy-preview-1756--bldrs-share-dev.netlify.app
  - Lighthouse on dev: Perf 47, A11y 88, Best Practices 92, SEO 100 — all reported "no change from production".
- [x] **#1747 stamp trap is settled — do not re-litigate.** `617e3f6f` committed `version: 1.1755.8b1f229a`; reverted in `016f7654`. At head, `package.json` `version` is `1.1743.6caa6b3`, byte-identical to `main`'s. Verified at both SHAs just now.

### Blocking the flip to ready

- [ ] **Forward-merge `main` again.** GitHub reports the PR `behind`; `main` has moved to `f38ea4f4` since the 17:04Z merge. `main` moves several times a day — re-check rather than trusting this line.
- [ ] **Re-run the gate at the current conway.** Everything under "Verification" in the description was recorded at **1.560.1539**; the branch now pins **1.588.1550**. That delta crosses [conway#566](https://github.com/bldrs-ai/conway/pull/566), which rewrites the aggregate prefetch on the **windowed** path — the exact path this pool's shards page through. That is why a re-gate is not optional here, and why the recorded numbers do not describe the code as it now stands.
  Per the repo's pre-ready gate: `yarn build-prod`, `yarn test-ci`, `yarn test-flows-build-and-serve` + `yarn test-flows <spec>` — then restore `package.json`, because `tools/updateVersion.mjs` stamps a local version into that tracked file on every build (#1747, above).
- [ ] **Get a codex review**, and answer or fix every finding, before flipping out of draft.

### Standing facts worth not rediscovering

- **Nothing in CI has ever validated this branch.** 9 check runs on the head: 6 Netlify (success/neutral), and `build`, `playwright-run`, `playwright-webifc-run` all with conclusion **`skipped`** — the draft gate. So the pre-ready run above is the *first* Actions signal this branch will get, and the husky hook does not cover `build-prod`, coverage thresholds, or Playwright.
- **The feature is a net loss today, and that is expected.** First real browser run on PSB was **3.0× slower overall** ([conway#541](https://github.com/bldrs-ai/conway/issues/541)): every worker re-parses the whole model, so parse contention dominates. [conway#541](https://github.com/bldrs-ai/conway/issues/541) — an `IfcAPI` open that consumes a prebuilt index instead of re-parsing per worker — is the **precondition** for the pool paying off. Even then the ceiling on PSB is **~30 %**, not 5×, because parse is not parallelised. `?feature=workers` stays **off by default** until [#541](https://github.com/bldrs-ai/conway/issues/541) lands.
- **Carries a workaround for [conway#540](https://github.com/bldrs-ai/conway/issues/540)**, in `src/viewer/ifc/ConwayGeometry.worker.js` (`declareWebPlatform()`, called at the top of `run()`). Recorded on that issue so the shim gets deleted when the upstream fix lands.

Nothing here is a request to change code — it is the state of the PR as of now.

---

pablo-mayrgundter:
Re-verification attempted 2026-08-31 (per the burn-down checklist above); **stopped at step 1 — the forward-merge is no longer mechanical.** Recording the state so the checklist doesn't mislead:

- `main` is 161 commits ahead of `5e4f9b97`, and merging it conflicts structurally in exactly the two files this PR lives in: `src/viewer/ifc/conwayDirectIfcLoader.js` (~706 conflict lines) and `src/viewer/ifc/ShareIfcLoader.js` (~138).
- The conflict is a **design collision, not drift**: this PR threads `{modelID, captured, geometryApi}` out of `parseIfcWithConway` — `geometryApi` being the worker-payload adapter that resolves pool-built geometry — while `main` (via `70f1c132`/`77644752`/`ecbfb973`/`08e6be85`, tied to conway#638/#660/#672) replaced that contract with conditional retention + an async `recapture()` closure, and `geometryApi` has no equivalent in it. `captured` is now empty precisely when `onMeshBatch` is passed, which the pool always does.
- Reconciling needs an explicit design call — either `recapture()` learns a second-engine source, or the pool produces `recapture()`-shaped output — before any rebase is meaningful. The merge was aborted; the branch is untouched at `5e4f9b97`, nothing pushed.
- Version gap has also crossed the engine's versioning-scheme migration: verified at conway `1.560.1539`; Share main pins `1.1585.676-g80859a4d`; npm latest is `1.1589.394-gf06115fb` (published today).
- conway#540 is still open, so the `declareWebPlatform()` workaround in `ConwayGeometry.worker.js` must stay.

Disposition: keep in draft. The pool logic itself (sharding, frame derivation, failure semantics) was sound against its old baseline; the integration seam moved out from under it. Also still upstream of shipping: conway#640 (sharded output divergence) and conway#682 (replicated demand prep).

---

Issue #1759, opened 2026-08-21
Github.spec.ts file-browser dropdown tests are deterministically broken — the MUI listbox never opens


pablo-mayrgundter:
Two E2E tests fail every attempt. Found while running Playwright on #1753's branch — **the first time the suite has ever run there**, which is the more uncomfortable half of this report (see *Why nobody noticed* below).

## The failures

`src/Components/Open/Github.spec.ts` — "Open 100: GitHub Integration › File browser Selector UI":

- `:152` — "selects org → repo → file through the dropdowns"
- `:163` — "filters repositories with the live text input"

Both fail with `Test timeout of 30000ms exceeded`, inside the shared `pickOption` open-and-pick helper.

## Deterministic, not flake — and here is the evidence

Run twice, against two different `@bldrs-ai/conway` pins, 4 attempts per test in total (Playwright config sets `retries: 1`):

| run | conway pin | passed | failed | skipped | total |
|---|---|---|---|---|---|
| baseline | 1.536.1507 | 151 | **2** | 44 | 197 |
| after bump | 1.543.1513 | 151 | **2** | 44 | 197 |

The failing-test sets were diffed programmatically, not by eye: **identical**. Same spec, same two line numbers, both runs.

The subtlety that makes a single run misleading — the *exact* line that times out moves, while the failure itself does not:

| attempt | times out at |
|---|---|
| baseline, attempt 1 | `:136` `await expect(combobox).toContainText(optionName)` |
| bumped, attempt 1 | `:132` `await option.waitFor({state: 'visible'})` |
| both retries | `:130` `await combobox.click({force: true})` |

All three are the same sequence in `pickOption`. The locator `getByTestId('openOrganization').getByRole('combobox')` **always resolves** to a real MUI `role=combobox` element; the call log then shows "attempting click action → scrolling into view if needed" and hangs. What never happens is the listbox popup becoming visible.

So: **the dropdown reliably does not open.** The stopping point is timing-dependent, which reads as flake from one run; two runs settle it.

## Not caused by the conway bump

Worth stating explicitly, since this surfaced during a bump: the baseline run is at the *old* pin, on the branch point before any bump commit existed, so the breakage predates it. Nothing in the failing surface touches geometry, model loading, or wasm — it is a MUI Select interaction. And no test changed state in either direction between the two pins.

## Why nobody noticed

[...]

PR #617, opened 2023-02-19
VYZN: (feature) hide/unhide and temporary isolate IFC elements

Ibrahim5aad:
This PR provides the following features:

1. Hiding and unhiding individual elements (including spatial aggregations/containers like IfcBuilding, IfcBuildingStorey, etc) (by the navigation tree, or keyboard shortcut: "H" when there is a selection you want to hide)
2. Unhide all hidden elements: keyboard shortcut "U"
3. Toggle reveal hidden elements: keyboard shortcut "R"
4. Toggle temporary Isolation mode: isolate selected elements in scene: keyboard shortcut "I"

This PR closes #384.

## Overview

https://user-images.githubusercontent.com/50090593/219949789-6e633e97-15a7-4516-a9cb-2dd22ae8f2bf.mp4

Issue #610, opened 2023-02-17
Bug: Fix the position of the page on mobile


OlegMoshkovich:
Steps to reproduce:
- on Mobile hold both the operations group and search bar
- slide up
- the whole page slides up
The position of the page should be fixed.

![image](https://user-images.githubusercontent.com/3433606/219651021-b5e95d87-ec90-4767-bcd2-c1121318ce79.png)

pablo-mayrgundter:
Related to #601 ?

OlegMoshkovich:
No, something else. 
The whole page moves up. It is rather strange.

pablo-mayrgundter:
Hm, not seeing it on android.. tho i can resize the page like that

pablo-mayrgundter:
heya, is this still happening for you?  i can't resize on android anymore, both chrome + ff work fine.

PR #1772, opened 2026-08-26
Pin conway 1.1565.608-g18366f01 — the spring fix and the Orbiter batch

pablo-mayrgundter:
Bumps `@bldrs-ai/conway` `1.588.1550` → `1.1565.608-g18366f01`. Diff is the pin and its lockfile entry, nothing else: 2 files, +5/−5.

This PR began as a smoke instance and became the pin bump once the release was blessed. **Both corpora are blessed** against this exact version — test-models#59 (`1754e15b`) and test-models-private#95 (`a57c78c6`), from conway run [33128910045](https://github.com/bldrs-ai/conway/actions/runs/33128910045).

**Previews on `80d31ab4`:** [prod](https://deploy-preview-1772--bldrs-share-prod.netlify.app) · [dev](https://deploy-preview-1772--bldrs-share-dev.netlify.app)

## The spring renders

An earlier smoke showed Orbiter's `ThumbTightenScrewAssembly` spring as a fan of flat sheets. Investigating it turned up more than a regression: **solid 970 had never rendered correctly in any release** — collapsed and invisible through `1.588.1550`, exploded and visible from `1.594.1554`. There was no good version to revert to.

| version | bbox | area | max edge | edges > 3.0 |
|---|---|---|---|---|
| `1.511.1441` … `1.588.1550` | **1.0 × 11.9 × 1.9** | 5.82 | 2.03 | 0 |
| `1.599.1551` (geom#185) | 6.1 × 11.9 × 5.3 | 63.29 | 5.97 | 230 |
| `1.594.1554` (geom#186) | 6.9 × 11.9 × 6.9 | 448.20 | 7.19 | 404 |
| **`1.1565.608` (this pin)** | 6.9 × 11.9 × 6.9 | **373.45** | **0.80** | **0** |
| source CAD | 6.9 × 11.9 × 6.9 | **374.71** | — | 0 |

Measured against the source CAD: OD **6.9100**, free length **11.8979**, coil radius **3.0019**, **7.00** turns. Area target derived by 60×1400 quadrature of the actual NURBS, cross-checked against Pappus to 0.2%; landing 0.3% under it is the expected sign, since the grid is inscribed. Signed volume 82.75, below `A × L ≈ 84` as an inscribed tessellation must be. Unpaired half-edges 19/659 → **0/35,454**.

## Four defects, in conway-geom#188

1. **Seam continuity.** The surface is u-closed — control row 0 identical to row 9 at distance exactly 0.0 across all 84 columns — and its loop traverses `EDGE_CURVE #29690` in *both senses*, so `u = +π` and `u = −π` are both exact preimages and **no residual can separate them**. Ten jumps of exactly one u-period, every one converged.
2. **No full-coverage path for b-spline.** Sphere, cylinder, cone and torus all had one; b-spline didn't, so earcut received a chart whose two long edges are the same 3D curve and emitted chords across the coil. The new path is gated on a seven-clause contract derived from what the routine consumes, read off the file's topology rather than inferred from measured area.
3. **A non-admissible bound in the trim bisection.** It ordered its search queue by `max(f(a), f(b))`, which prefers intervals that are uniformly low over intervals that *contain* the minimum. On a 7-turn helix the interval bracketing the exact answer ranked second and was never expanded — **the spring's trim ran v ∈ [6.25, 37.73] of [0, 43.98], silently missing its first and last turn** for as long as that code has existed.
4. **The descent treated a closed surface's seam as a wall.** `solveFromSeed` clamped every trial step into the domain; on a closed axis a step off one end is a *crossing*, not an exit. All 21 Armijo trials took zero-length steps and the solve returned its own seed at residual 0.1226 against a 1.19e-05 target.

## Also in this pin

| conway | |
|---|---|
| **#593** | `readReal`'s radix scaling clamped — an out-of-bounds read returned `undefined`, so `primary /= undefined` produced NaN silently. 48,593 NaN vertices → 0 |
| **#600** | trimmed B-spline extraction; 33,545 collapsed trims → 0 |
| **#604** | the screw threads — NURBS inverse-solve continuity seed. 118,867 duplicate triangles → 59 |
| **#609** | a retracing seam loop on a sphere counts as full coverage |
| **#605** | grid-placed products keep their preview retry |
| **#610** | the version scheme (`MAJOR.COMMIT.ISSUE-g<hash>`) — CI only |
| **#617** | adaptive residency for the STEP aggregate pager |

## What the bless established

The rc diff removed 39 digest rows, which looked like lost geometry. It isn't. The AP214 digest writes two row kinds, split by the `Void` column: **mesh** rows (hashed post-weld through `Reify()`) and **curve-cache** rows. All 39 removed rows are curve-cache rows — **not one mesh row is removed anywhere in either corpus**.

A curve only enters `model.curves` via the whole-curve-trim *recovery* path, so a curve row is a **defect marker, not a geometry record**. Verified across 7 models × 2 versions: `curve rows == count of whole-curve-trim warnings`, exactly. The hash they shared decodes to `"v nan nan nan" × 25` — 25 NaN points, which is why one hash spanned four unrelated models and two entity types, and why those points were poisoning face boundaries.

Post-weld, both corpora reproducing byte-exactly from the published builds:

| | supercap2/3 | Orbiter |
|---|---|---|
| errors | 1 → **0** | 1 → **0** |
| degenerate triangles | 356 → 356 | **124,506 → 6,226** |
| restored face | `#2733` +138 tris, bbox unchanged | `#50626` empty → 2,208 tris |

## What to look at

**The spring** — a 7-turn coil, not a fan.

**Orbiter's fasteners, 964 / 966 / 971 / 977.** Unaffected by the spring work; threads should read cleanly.

**`Right_Hand.step`** — expect movement: 6 of 39 digests, with area, bbox, largest triangle and shard count **identical** on all six and no orientation flips. Its 0.59% pixel delta is diffuse at low amplitude (216 of 2,423 changed pixels exceed 32/255) — retessellation shading, not shape. Silhouette unchanged.

**`index.ifc`** should be pixel-unchanged.

## Known and filed, not fixed here

- **966** gains two reverse-winding triangles and **978** four duplicated half-edges — both on faces already carrying pre-existing breakage, both **0 pixels** of 810,000.
- **19720** (Right_Hand) gains 132 degenerate triangles. Inert — zero area, no pixels — but waste.
- **Solids 955, 956** change hash without demonstrated improvement (65 of 28,830 vertices move by ≤ 3.61e-6). 954 and 975 *are* demonstrated: degenerate triangles 7,740 → 8 and 620 → 0.
- **conway#626** — the whole-curve-trim fallback still fires on two NIST models, which *gain* curve rows (4→8, 10→14). They recover cleanly as 65-point non-NaN B-splines and both models' meshes improve, so not a blocker; but this bless makes it the new floor.
- **conway#627** — the rc perf delta's noise floor exceeds its signal. Two attempts of the *same job at identical code* differ by 13.66% median while the reported regression is 9.40% median; every digest is byte-identical between them. The timing columns can't gate a release until both pins run in one job.
- **conway#619** (the grid can't refine u without cross-face coordination with the cap faces — the spring escapes this only because its ring is a circle of uniform curvature), **conway#621** (closure inferred from control data rather than evaluated endpoint curves; zero corpus reach today), **conway#625** (`tesselate` livelocks on an exactly-representable uv lattice).
- **conway#608 stays open.** Face `#50977` was one of its 23; the other 22 are untouched.

## Pre-flight

Draft gating skips `build-prod`, coverage and Playwright, so all three were run locally before this was flipped:

| gate | result |
|---|---|
| `yarn build-prod` | PASS — 14/14 static pages, 88.66s |
| `yarn test-ci` | PASS — 244 suites, 2454 passed / 5 skipped, 13 snapshots; coverage above thresholds |
| `yarn test-flows` | PASS — 161 passed, 41 skipped, 0 failed, 0 flaky |

`main` was merged in and the merge commit amended so husky's `pre-commit` actually ran on it — `git merge` does not fire that hook.

One note on the flows leg: a first run at the config default `workers: 4` failed 4 tests in `GoogleDriveConnect.spec.ts` on 30s timeouts. Nothing was skipped or quarantined. It is worker contention, not a regression — the one passing test in that file took 27.4s of its 30s budget, the failures are in fully-mocked auth/picker paths with no geometry and no network, CI runs `--workers=2` (`.github/workflows/test-flows.yml:125`), and a full re-run at `--workers=2` was green with 0 flaky. `main` at `0d99af17` is green in CI. This matches the failure mode documented in #1759/#1764: the viewer's rAF loop pins the main thread through SwiftShader and Playwright's actionability polling turns contention into timeouts on arbitrary lines.


pablo-mayrgundter:
@codex review

Small diff, but it has two failure modes that a generic pass will miss, so please attack these specifically rather than the two lines:

**1. Did the conway pin actually survive the build?** `tools/updateVersion.mjs` stamps a local version into the tracked `package.json` on every build (Share#1747), and the repo's standard remedy is `git checkout -- package.json`. **That remedy is wrong on this branch** — `package.json` also carries this PR's entire reason for existing. A checkout would silently revert the pin and leave `yarn.lock` resolving a version `package.json` no longer requests, which would still install correctly from cache locally and only diverge later. The `version` field was restored by hand instead. Please verify from the diff that:
- the only `package.json` change is `@bldrs-ai/conway` `1.588.1550` → `1.1565.608-g18366f01`
- there is no `version` field change
- `yarn.lock` resolves that same version, and its integrity entry is consistent

**2. Did the gate actually run on the merge commit?** `main` was merged in, and per this repo's `CLAUDE.md`, husky's `pre-commit` fires on `git commit` and **not** on `git merge` — so a merge commit lands ungated. It was amended (`git commit --amend --no-edit`) to force the hook. If you can see anything in the history suggesting the amend didn't take, or that the hook was bypassed, say so.

**Context for the change itself:** the pin is conway `1.1565.608-g18366f01`, which is blessed against both corpora — bldrs-ai/test-models#59 and bldrs-ai/test-models-private#95, from conway run 33128910045. Its geometry changes were reviewed in conway#618 and bldrs-ai/conway-geom#188 and gated by the full rc regression, so the geometry is not what needs your attention here; the packaging is.

**Not asking you to re-litigate:** the two follow-ups filed rather than fixed (conway#626, conway#627) and the known-and-filed list in the description are deliberate scope decisions, not oversights.

---

PR #1784, opened 2026-08-27
Preserve query string on /share redirect (fixes gclid loss)

pablo-mayrgundter:
## Problem

`https://bldrs.ai/?gclid=TEST123` drops the query string on load.

`ShareRoutes.jsx`'s `Forward` component redirects `/share` → `/share/v/p` with a
bare `navigate(dest)`, which does not carry `location.search`. Everywhere else in
the codebase redirects go through `navWith`, whose defaults preserve search and
hash — this is a single missed call site. `BaseRoutes.jsx` (step before) preserves
it correctly, and `navToDefault` (step after) tries to, but by then
`location.search` is already empty.

## Impact

GA4's `gtag/js` loads asynchronously and reads `window.location` live at send
time, after React's effect chain has already rewritten the URL. With `gclid`
gone and a `google.com` referrer, paid sessions are filed as `google / organic`.

Observed: ~872 Google Ads clicks over Aug 17–26 produced ~75 `google / cpc`
sessions in GA4, while `google / organic` rose from ~5.7 to ~110 sessions/day
with a geography and device mix matching the ad campaigns exactly (100% desktop;
campaigns run mobile −100%). Google Ads recorded zero conversions because
nothing tied back to a `gclid`. Search Console independently reports ~0 organic
clicks for the same period.

## Changes

- `src/ShareRoutes.jsx` — `Forward` now redirects through `navWith`, passing the
  router's own `search` and `hash` explicitly.

  Passing them explicitly is load-bearing, not defensive. `navWith`'s default
  options (`src/utils/navigate.js:103-106`) are evaluated against the
  module-scope global `location` — `window.location` — not the router's
  `useLocation()`. `BrowserRouter` keeps the two in sync so production behaves
  the same either way, but `MemoryRouter` never touches `window.location`, so
  under test the default falls back to jsdom's empty location and drops the
  search one level *below* where the original bug dropped it. A first pass at
  this fix used bare `navWith(navigate, dest)` and failed its own regression
  test for exactly that reason.

- `src/ShareRoutes.test.jsx` — unit-level regression guard on the forward.

- `src/routes/landingQuery.spec.ts` — real-browser E2E, desktop and mobile via
  `describeMobileAndDesktop`. The unit test asserts on react-router's abstract
  location, but the bug is specifically about `window.location`: gtag reads the
  global live at send time, so a router-level assertion can pass while GA4 still
  sees no `gclid`. This spec drives the full
  `/` → `/share` → `/share/v/p` → `/share/v/p/index.ifc` chain in Chromium and
  asserts on `window.location.search`, covering `gclid`, both `utm_*` params and
  an unknown param. Verified falsifiable: reverting `Forward` to the original
  bare `navigate(dest)` fails both form factors with `gclid` `null`.

- `public/index.html` — move the gtag stub/config block from `<body>` into
  `<head>`. Separate concern: Search Console's tracking-snippet ownership
  verification requires the snippet in `<head>` (PR #1773 worked around this
  with a meta tag). This does not by itself fix the attribution bug. It does not
  affect the MSW hermeticity the neighbouring AdSense comment describes — that
  applies to a script issuing its own network request, whereas the gtag *stub*
  makes none (it only defines `dataLayer`/`gtag` and queues a `config`); the
  network-triggering `googletagmanager.com` script is injected later from
  `src/index/ga.js#setupGa`.

## Review

codex was out of usage quota on this PR, so this went through a sub-agent round
per [agent-workflow.md §Review](design/new/agent-workflow.md). Findings and
dispositions: https://github.com/bldrs-ai/Share/pull/1784#issuecomment-5448817224

## Verification after deploy

1. `https://bldrs.ai/?gclid=TEST123` should survive to
   `/share/v/p/index.ifc?gclid=TEST123`.
2. GA4 Realtime should show the session as `google / cpc`, not `google / organic`.
3. Within 24–48h, `google / organic` should fall back toward its pre-Aug-17
   baseline and `google / cpc` should rise toward the Ads click count.

## Follow-up

`src/utils/navigate.js:23` uses `location.query`, which is not a DOM property and
is therefore always `''`. Harmless today only because `navWith`'s separate
`options.search` does the real work, but it sits directly next to this fix and a
future "cleanup" that turned it into a real value would double-append the query.
Tracked separately.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,93 +1,93 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="utf-8"/>
-    <meta name="description" content="Build every thing together: Collab and Share 3d models online"/>
-    <meta name="keywords" content="ifc, ifc viewer, cad viewer, share ifc, share model, open source"/>
-    <meta name="viewport" content="width=device-width, initial-scale=1"/>
-    <meta name="theme-color" content="#000000"/>
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0">
-    <link rel="icon" href="/favicon.ico"/>
-    <title>bldrs.ai</title>
-    <meta name="google-site-verification" content="soxvqCycG6656Bmiml345OQ0BPfPnfzHBx_CJ3dSEeQ" />
-    <!-- Start Single Page Apps for GitHub Pages -->
-    <script type="application/javascript">
-      // Single Page Apps for GitHub Pages
-      // MIT License
-      // https://github.com/rafgraph/spa-github-pages
-      (function (l) {
-        if (l.search[1] === '/') {
-          var decoded = l.search.slice(1).split('&').map(function (s) {
-            return s.replace(/~and~/g, '&')
-          }).join('?')
-          var finalPath = l.pathname.slice(0, -1) + decoded + l.hash
-          window.history.replaceState(null, null, finalPath)
-        }
-      }(window.location))
-    </script>
-    <!-- End Single Page Apps for GitHub Pages -->
-    <style>
-      :root {
-        color-scheme: light dark;
-        background-color: Canvas;
-        color: CanvasText;
-      }
-    </style>
-    <link rel="stylesheet" href="/index.css">
-  </head>
-  <body>
-    <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="root"></div>
-    <script type="application/javascript">
-      const loader = document.createElement('script')
-      const pathPrefix = window.location.pathname.startsWith('/Share') ? '/Share' : ''
-      loader.setAttribute('type', 'module')
-      loader.setAttribute('src', pathPrefix + '/index.js')
-      document.body.appendChild(loader)
-    </script>
-    <!-- gtag stub only: buffers events into dataLayer everywhere (gtagEvent
-         calls + E2E assertions rely on it). The external gtag/js loader is
-         injected by src/index/ga.js, and only on prod hosts outside
-         automation — keeps preview/dev/CI traffic out of prod analytics.
-         The config ID below must match ga.js#GA_MEASUREMENT_ID. -->
-    <script type="application/javascript">
-      window.dataLayer = window.dataLayer || [];
-      function gtag(){dataLayer.push(arguments);}
-      gtag('js', new Date());
-      // Carried IN the config call rather than a `set` queued before it.
-      // config is what triggers page_view / session_start / first_visit — the
-      // events that make landing page and new-vs-returning queryable per user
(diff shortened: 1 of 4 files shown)


pablo-mayrgundter:
## CI fix + review round

codex is out of usage quota on this PR, so per [agent-workflow.md §Review](https://github.com/bldrs-ai/Share/blob/main/design/new/agent-workflow.md) this was a sub-agent round instead.

### CI: `build` was red, now green

The new `ShareRoutes.test.jsx` failed its own two assertions (`Expected: "?gclid=TEST123" / Received: ""`).

Cause: `navWith`'s default options (`src/utils/navigate.js:103-106`) are evaluated against the module-scope global `location` — `window.location` under jsdom — not the router's `useLocation()`. `MemoryRouter` never touches `window.location`, so the default fell back to jsdom's empty location and dropped the search one level *below* where the original bug dropped it. `BrowserRouter` keeps the two in sync, so production was unaffected either way; only the test could see it.

`Forward` now passes the router's own `search`/`hash` explicitly (46ff2c8). `build` is green on that commit.

### Added: real-browser E2E, desktop + mobile

`src/routes/landingQuery.spec.ts` (b392763). The unit test asserts on react-router's abstract location, but the bug is specifically about `window.location`: gtag/js loads asynchronously and reads the global live at send time, so a router-level assertion can pass while GA4 still sees no `gclid`. The spec drives the full `/` → `/share` → `/share/v/p` → `/share/v/p/index.ifc` chain in Chromium and asserts on `window.location.search`, covering `gclid`, both `utm_*` params, and an unknown param.

Verified falsifiable: reverting `Forward` to the original bare `navigate(dest)` fails both form factors with `gclid` `null`.

### Reviewed and cleared

- **`public/index.html` reorder** — structurally valid, and it does *not* break the MSW hermeticity the neighbouring AdSense comment describes. That comment is about a script that makes its own network request; the gtag **stub** makes none (it only defines `dataLayer`/`gtag` and queues a `config`). The network-triggering `googletagmanager.com` script is injected later from `src/index/ga.js#setupGa`. Jest never parses `index.html`, and `realModelOpen.spec.ts` only needs the stub to have run by assertion time — moving it earlier can only help.
- **Forwarding `location.hash`** — inert for this path, not harmful. `handleRoute` (`src/routes/routes.ts:38-41`) returns `null` with no file segment, so `navToDefault` overwrites the hash with its hardcoded camera hash regardless.

### Follow-up, not this PR

`src/utils/navigate.js:23` — `` `${appPrefix}/v/p/index.ifc${location.query || ''}` ``. `Location.query` isn't a DOM property, so it's always `''`. Pre-existing and harmless today only because `navWith`'s separate `options.search` does the real work — but it sits right next to this fix, and a future "cleanup" that turns it into a real value would produce a double-appended query. Worth a separate issue.

---

Issue #1776, opened 2026-08-27
STEP models second load shows (or stores) incorrect NavTree


pablo-mayrgundter:
Repro:

Open https://bldrs.ai/share/v/gh/bldrs-ai/test-models/main/step/nist/as1-oc-214.stp/5/6211#n:;p:

<img width="1470" height="773" alt="Image" src="https://github.com/user-attachments/assets/5be2ee6b-1d0c-4bfe-94fa-8626ad0e467a"/>

Reload page, NavTree loses all info (have to click exapand it, but just repeats top-level name a couple times).  Properties too

<img width="1465" height="784" alt="Image" src="https://github.com/user-attachments/assets/b6ec8f39-8a26-453b-b2e6-bc5d1e00af82"/>

No error from cached load:
<img width="1452" height="883" alt="Image" src="https://github.com/user-attachments/assets/f0c2ee09-18ce-478f-99dc-30f96159832c"/>

Bisected down to https://github.com/bldrs-ai/Share/pull/1753 (1752 is fine)

<img width="1464" height="880" alt="Image" src="https://github.com/user-attachments/assets/97e68aba-9456-4176-ad77-ed5ec8ae8d4f"/>

---

**Root cause.** Conway's store-backed open (`OpenModelStream`, the M1b path that arrived with #1753's engine bump) reserves the model handle *before* it sniffs the format, and `fromStore` is IFC-only. Share offered it every file where `isIfc` was true — which includes STEP. So every STEP load burned handle 0 on a model that never opened, and the buffered `OpenModelStreamed` retry parsed as **handle 1**.

The bisect to #1753 was right, and the reason the first load looked perfect is that the model's own closures carry the correct handle. What broke is everything addressing the scene-level id `0` — including the idle GLB writer, which reads `model.modelID`, a property `CadView.jsx#loadModel` pins to `0` the moment `load()` resolves. Conway's `properties.*` shim optional-chains through `getPassthrough(modelID)`, so the spatial-tree and element-properties captures resolved `undefined` instead of throwing: the writer dropped both `BLDRS_*` extensions and still reported a successful write. The reload served that artifact, found no `BLDRS_spatial_tree`, and fell back to walking the GLB scene graph — which is the repeated-root tree in the screenshot.

That also explains the "no error from cached load": there was nothing to report. The corruption happened on the *previous* load, silently.

[...]

PR #1788, opened 2026-08-28
Run the GLB cache-hit E2E specs on mobile as well as desktop

### LINKED ISSUE #1787: Cache-hit GLB E2E specs run desktop-only — add the mobile leg
CLAUDE.md's standing rule for UI work: a happy-path E2E on desktop **and** mobile, via `describeMobileAndDesktop` (`src/tests/e2e/formFactor.ts`) — "if it's not tested, it doesn't work".

The three GLB cache-hit specs restored to life in #1783 are plain `test.describe`, so they only ever run at the config's desktop viewport:

- `src/Components/NavTree/NavTree.cacheHit.spec.ts` (2 tests: IFC and STEP)
- `src/Components/Properties/Properties.cacheHit.spec.ts` (1 test)
- `src/Containers/sceneHighlightPermalink.spec.ts` (2 tests; the second is the cache-hit leg)

Why this coverage in particular. These are the specs that guard #1776 — a STEP model whose *second* (cache-hit) load rendered an empty NavTree. It shipped because all of these were `test.fixme`, so 244 Jest suites and both Playwright jobs passed over it. The cache round-trip itself (writer → OPFS → reader) is plausibly form-factor independent, but the surfaces these specs assert through are not: NavTree renders through `react-window`, whose mounted window is a function of viewport height, and `expectRealHierarchy` counts distinct `data-node-label` values against a threshold. A mobile-only regression in those panels is currently invisible.

Scope: wrap the three specs in `describeMobileAndDesktop` and make the mobile leg pass honestly — no skips, no `fixme`, no assertion weakened past the point where it can fail.

pablo-mayrgundter:
Closes #1787.

The three GLB cache-hit specs restored in #1783 were plain `test.describe`, so they only ever ran at the config's desktop viewport. They are the specs that guard #1776 — the STEP model whose *second* (cache-hit) load rendered an empty NavTree — so the missing mobile half is not a formality. This wraps all three in `describeMobileAndDesktop`.

## What changed

- `src/Components/NavTree/NavTree.cacheHit.spec.ts` (IFC + STEP)
- `src/Components/Properties/Properties.cacheHit.spec.ts`
- `src/Containers/sceneHighlightPermalink.spec.ts`

10 tests, 10 green: `yarn test-flows <the three specs> --retries=0`.

Nothing beyond the wrapper was needed to make them pass. No threshold lowered, no branch on `ff.isMobile`, no scroll or extra expansion — the mobile leg runs the identical body. Two comment corrections are the only other edits.

## What the mobile leg turned up

**The round trip is form-factor blind, and now that is measured rather than assumed.** The cache-hit permalink test prints its scene state, and mobile reports exactly what desktop does: `{"selectedElements":["2867"],"mergedSubsets":1,"batchedSelSetTotal":0,"batchedMeshes":0,"meshes":16,"misaligned":0}`.

**`expectRealHierarchy`'s stated reason was wrong, and the mobile run is what made it checkable.** The doc said rows are the wrong measure because the row count "saturates at the viewport (~25-30 at the config's 1280x800)". Both fixtures flatten to **11 rows total**, so nothing saturates at 1280x800 and nothing saturates at 390x844 either — all 11 mount at both sizes. Virtualization is a reason the assertion *stays* correct if a fixture grows; it is not the reason it is correct today. The load-bearing reason is the other one the comment gave: the #1776 fallback emits many rows carrying one label. Comment rewritten to say that, and to record the two viewports the suite now runs at.

**`MIN_DISTINCT_LABELS = 5` is a ceiling, not a floor.** `index.ifc` and `index.step` both flatten to exactly `Bldrs / Build / Every / Thing / Together` — 5 distinct labels is the *most* either fixture can offer. There was no headroom to trade away for the smaller viewport, and raising it would fail on desktop too. Noted at the constant so the next reader does not go looking for slack that is not there.

## Proving the mobile assertions can fail

Each of these was run, watched go red, and reverted.

| Break | Result |
|---|---|
| `MIN_DISTINCT_LABELS` 5 → 99 | All 4 NavTree tests red. Message: `expanded tree had 5 distinct label(s) over 11 row(s): ["Bldrs","Build","Every","Thing","Together",…]` — byte-identical on `[desktop]` and `[mobile]`. This run is where the ceiling above was measured. |
| `expandTree(navTreePanel)` removed | All 4 red at `1 distinct label(s) over 1 row(s): ["Bldrs"]`, mobile included — so expansion is load-bearing at phone height, and the collapsed root does not satisfy the threshold by accident. |
| Properties `Name` `'Together'` → `'TogetherXX'` | `[mobile]` red, reporting the real row text `"NameTogether"` — the panel genuinely rendered through `TabbedPanels` on the phone. |
| Permalink `toHaveCount(1)` → `toHaveCount(2)` | `[mobile]` red, `Received: 1` — one NavTree row really is selected on mobile, not zero. |
| Permalink `hit.meshes > 0` → `> 1000` | `[mobile]` red, `Received: 16` — the cache-hit picking tables are present on mobile. |

Final state re-run after restoring all five: 10 passed.

## Note on the eslint line

An earlier revision of this PR carried `root: true` in `.eslintrc.cjs`. Three PRs each hit the same unrunnable commit gate in an agent worktree and each patched it locally, so it has been **split out to #1792** (merged) and removed here. Main now carries it, and this branch has merged main.

## Checks

- `yarn test-flows` on all three specs, both form factors, `--retries=0`: 10 passed.
- Husky pre-commit gate on every commit here, including the merge commit (forced with `--amend`, since `pre-commit` does not fire on `git merge`): eslint + typecheck + full Jest suite green.
- CI on `5b18cff`: `build` ✅, `playwright-run` ✅, `playwright-webifc-run` ✅.
- `package.json` version stamp from the Playwright build reverted before committing (#1747).

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,5 +1,14 @@
 /* eslint-disable no-magic-numbers */
 module.exports = {
+  // Stop the config cascade at the repo root. Without this ESLint keeps
+  // walking up, and an agent worktree under `.claude/worktrees/` (this repo's
+  // own convention) sits INSIDE another checkout of this repo: the ancestor
+  // .eslintrc.cjs then loads a second copy of every plugin ("couldn't
+  // determine the plugin ... uniquely") and, because the ignore base moves up
+  // with it, every file under the worktree looks like it lives in a dotfolder
+  // and is "ignored by default". Both make `yarn precommit` unrunnable there.
+  // In a normal checkout nothing sits above the root, so this is a no-op.
+  root: true,
   env: {
     browser: true,
     es2021: true,
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,14 +1,5 @@
 /* eslint-disable no-magic-numbers */
 module.exports = {
-  // Stop the config cascade at the repo root. Without this ESLint keeps
-  // walking up, and an agent worktree under `.claude/worktrees/` (this repo's
-  // own convention) sits INSIDE another checkout of this repo: the ancestor
-  // .eslintrc.cjs then loads a second copy of every plugin ("couldn't
-  // determine the plugin ... uniquely") and, because the ignore base moves up
-  // with it, every file under the worktree looks like it lives in a dotfolder
-  // and is "ignored by default". Both make `yarn precommit` unrunnable there.
-  // In a normal checkout nothing sits above the root, so this is a no-op.
-  root: true,
   env: {
     browser: true,
     es2021: true,
--- a/src/Components/NavTree/NavTree.cacheHit.spec.ts
+++ b/src/Components/NavTree/NavTree.cacheHit.spec.ts
@@ -6,9 +6,10 @@ import {
 } from '../../tests/e2e/utils'
 import {waitForModelReady} from '../../tests/e2e/models'
 import {captureGlbLogs, resetGlbLogs, waitForGlbLog} from '../../tests/e2e/glbLogs'
+import {describeMobileAndDesktop} from '../../tests/e2e/formFactor'
 
 
-const {beforeEach, describe} = test
+const {beforeEach} = test
 
 
 // Upper bound on toggles clicked. One click per iteration (see below), so this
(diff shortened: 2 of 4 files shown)


pablo-mayrgundter:
@codex review

Wraps three GLB cache-hit specs in `describeMobileAndDesktop` so they run at both viewports. They are the specs that guard #1776, so the mobile half is coverage, not ceremony. The eslint line this PR originally carried has been split out to #1792 and removed here.

Please attack these specifically:

1. **The threshold is a ceiling, not a floor.** `expectRealHierarchy` asserts `>= 5` distinct `data-node-label` values, and both fixtures (`index.ifc`, `index.step`) flatten to exactly `Bldrs / Build / Every / Thing / Together` — 5 is the *most* either can offer. So the assertion passes only at the exact maximum. Is that a meaningful guard or a coincidence waiting to break? Consider what happens when a fixture changes by one node in either direction.

2. **The two form factors run a byte-identical body.** No `ff.isMobile` branch, no scroll, no extra expansion. That is the claim — that the writer → OPFS → reader round trip and the panels it feeds are form-factor blind. Is there a mobile-specific path it therefore never exercises? `TabbedPanels`, the mobile drawer, and anything that changes which control opens a panel are the places I would look.

3. **Assertions that cannot fail.** This is the recurring defect on this work stream — eight found so far across #1777/#1783/#1789, each passing for a reason other than its comment claimed. The PR body lists five mutations run against the mobile leg; please check whether they cover the right things, and whether any *unmutated* assertion is satisfied by construction at the smaller viewport.

4. **A correction I made to my own earlier comment, which I would like checked.** #1783 justified counting distinct labels rather than rows because "the row count saturates at the viewport (~25-30 at 1280x800)". That is wrong: both fixtures produce 11 rows total, so nothing saturates at either viewport and all 11 mount at both. I rewrote the comment to say the load-bearing reason is the other one (the #1776 fallback emits many rows carrying one label). If the virtualization reasoning still matters for a reason I have missed, say so.

---

PR #1792, opened 2026-08-28
Stop the ESLint config cascade at the repo root

pablo-mayrgundter:
Split out of #1788 / #1789 / #1790, which each carried this same line. Landing it once here and dropping it from those three keeps each of them to its own subject and removes a three-way conflict.

## The bug

`.eslintrc.cjs` never set `root: true`, so ESLint walks up the filesystem and merges **every ancestor `.eslintrc` it finds** into this project's config. Anything in a parent directory silently becomes part of how this repo lints.

Demonstrated directly rather than inferred. With a config one level above the checkout declaring a plugin this repo cannot resolve:

```
### WITHOUT root:true (main today)
$ yarn eslint src/Filetype.js
The plugin "eslint-plugin-this-plugin-does-not-exist" was referenced from the
config file in "../.eslintrc.cjs".

### WITH root:true (this PR)
$ yarn eslint src/Filetype.js
$          # clean
```

The failure mode is what makes it worth fixing: eslint **hard-stops** rather than degrading, so it takes `yarn precommit` and the husky pre-commit hook with it. The commit gate doesn't get weaker — it becomes unrunnable.

## How it surfaced

Normally there is no ancestor config, so this is a no-op — which is why it went unnoticed. It stops being a no-op the moment the checkout is nested inside another one.

Three agent sessions running in `git worktree` checkouts under `.claude/worktrees/` each hit an unrunnable commit gate, and each patched this line locally and independently. That's the whole reason it's three-way duplicated across #1788, #1789 and #1790.

It is not only an agent concern, though: any contributor with a stray `.eslintrc` anywhere above their checkout — a home directory, a shared parent for several repos — gets the same hard stop, and the error message points at a file outside the repo.

## Also: gitignore `.claude/worktrees/`

Those checkouts are transient and machine-local. The branch inside one is pushed like any other, but the working copy is not part of the repo, and while untracked it reports as pending changes in every session that dispatches an agent.

## What this does *not* fix

The agents also reported that files inside a worktree read as `File ignored by default`. **`root: true` does not fix that** — I checked, and the warning persists either way. It's a separate ESLint behaviour about paths under dot-directories, so it isn't claimed here and isn't addressed.

## Checks

- The before/after above, run both ways.
- Husky pre-commit gate on this commit: eslint clean, typecheck clean, **246 suites / 2477 passed**.
- No behaviour change for a flat checkout or for CI — there is no ancestor config in either, which is exactly why this is safe to land on its own.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NuTsqJu1rqoz8f4FSQ9vbE)_

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,5 +1,18 @@
 /* eslint-disable no-magic-numbers */
 module.exports = {
+  // Stop the config cascade at the repo root. Without this ESLint keeps
+  // walking up the filesystem and merges every ancestor `.eslintrc` it finds,
+  // so anything in a parent directory silently becomes part of this project's
+  // lint config — and a plugin named there that this repo cannot resolve makes
+  // eslint fail outright rather than degrade, taking `yarn precommit` and the
+  // husky hook down with it.
+  //
+  // Normally there is no ancestor config and this is a no-op, which is why it
+  // went unnoticed. It stops being a no-op the moment the checkout is nested
+  // inside another one — a `git worktree` under `.claude/worktrees/` for an
+  // agent session is how it surfaced, three times independently, each time as
+  // an unrunnable commit gate.
+  root: true,
   env: {
     browser: true,
     es2021: true,
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,9 @@ deno.lock
 # no LFS bandwidth; the generated thumbnails in src/assets/thumbnails ARE
 # committed, these source models are not.
 tools/thumbnails/cache/
+
+# Sandboxes for agent sessions (`git worktree` checkouts under .claude/).
+# Transient and machine-local: the branch inside one is pushed like any other,
+# but the working copy itself is never part of the repo. Untracked they show up
+# as pending changes in every session that dispatches an agent.
+.claude/worktrees/


pablo-mayrgundter:
@codex review

Two lines of config, but they change how lint resolves for everyone, so worth a look at the edges rather than the diff.

1. **Is `root: true` correct *here*, at `.eslintrc.cjs` in the repo root?** The repo also has nested configs — check whether any of them relied on cascading up to this one, and whether stopping the cascade at this file changes the effective config for any linted path. The claim is that a flat checkout has no ancestor config so this is a pure no-op; if there is a path where it isn't, that is the finding.

2. **Does anything in the toolchain lint from a different cwd or root?** `yarn precommit`, the husky hook, the `build` job, and any editor integration each invoke eslint their own way. `root: true` changes which config files are collected; if one of those invocations depends on collection continuing past this file, it breaks quietly.

3. **The `.gitignore` entry.** `.claude/worktrees/` — check it cannot match anything a contributor would want tracked, and that it does not interact badly with the existing `.claude/` entries above it (`settings.local.json`, `scheduled_tasks.lock`).

Two things I deliberately did **not** claim, in case either looks like an omission: `root: true` does *not* fix the `File ignored by default` warning the agents also reported (verified — it persists either way), and I have not reproduced the "couldn't determine the plugin uniquely" duplicate-plugin error myself. My repro shows the more general cascade failure instead: an ancestor config's unresolvable plugin hard-stops eslint. If you think the duplicate-plugin story is load-bearing for justifying this, say so and I will reproduce it properly rather than leave it resting on three agent reports.

---

PR #1799, opened 2026-08-30
Bump @bldrs-ai/conway to 1.1575.649-gdeabdf16 (SHARE-1NK engine fix)

pablo-mayrgundter:
Engine pin bump carrying bldrs-ai/conway#654 (fixes bldrs-ai/conway#649 / Sentry [SHARE-1NK](https://bldrs-ai.sentry.io/issues/7699115915/)): the demand pump now evicts at the **start** of a pump call — everything call N delivers stays resident until call N+1 begins, so the copy window `onMeshBatch` relies on is safe — and `GetGeometry` degrades to the dummy not-found path instead of cloning a freed embind handle.

With this pin, #1798's per-placement skip hardening becomes belt-and-suspenders rather than the only line of defense, and large-model loads under `GEOMETRY_BUDGET_MB` (the SHARE-1NK cohort: 1–2 GB Revit IFCs) stop losing geometry. This is the last piece of the SHARE-1NK chain; the Sentry issue gets resolved once this deploys.

**Diff:** one dependency line in `package.json` + the matching `yarn.lock` entries.

**Validated against the new engine locally:** `build-prod` ✓, `test-ci` ✓ (258 suites / 2639 tests, coverage thresholds enforced), `test-flows src/loader/conwayDirect.spec.ts` 3/3 ✓ — the flows build bundles the new conway wasm, so the E2E exercised `1.1575.649-gdeabdf16` end-to-end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WFte6WW6bqxxL7VPAMTr1g

---
_Generated by [Claude Code](https://claude.ai/code/session_01WFte6WW6bqxxL7VPAMTr1g)_

Issue #654, opened 2023-03-10
SD Exploratory Work


oo-bldrs:
- [x] Stand up local GPU instance and document hardware specifications in wiki
- [x] Deploy latest version of stable diffusion
- [x] Build experimental API to generate image derivatives
- [x] Measure performance (success and failures) of image derivatives at different resolutions and inference steps
- [ ] Estimate costs and throughput
- [ ] Present results

PR #1800, opened 2026-08-30
bldrs-ai/conway#638: stop retaining the pumped FlatMesh stream where something else owns it

pablo-mayrgundter:
Part of bldrs-ai/conway#638 (and so of bldrs-ai/conway#635). **This is the Share half only** — the conway half is bldrs-ai/conway#657 (open draft), the pin bump to a conway carrying it is a separate step, and the joint three-spine measurement is still owed. So this deliberately carries no closing keyword.

Built on [#638's design comment](https://github.com/bldrs-ai/conway/issues/638#issuecomment-5455957015), not the issue body: the body says conway's `meshMap` replay backs Share's fallback, and it does not — Share's fallback fires only on an empty pump, which means conway opened the model non-deferred, where `streamAllMeshes` re-walks live natives. Re-traced against the pinned engine and confirmed.

> **Pin note.** This branch has merged `main`, so it now sits on conway **1.1575.649-gdeabdf16** (conway main with #654's eviction-timing fix, still **without** #657). Every pin-sensitive claim below was re-verified against that engine, and one of them changed — see "What the recapture improvement actually is".

## What this establishes

`parseIfcWithConway` accumulated every pumped `FlatMesh` into `captured` unconditionally. On the streaming path that array is dead weight — `onMeshBatch` has already assembled each batch into the durable `BatchedMesh` by the time the next one lands — and it is one of **three pointer spines over one shared 475 MB graph** (conway holds `meshMap` and `vectorFlatMesh`; this is Share's).

- **`captured` now accumulates only where nothing else takes delivery.**
- **The degraded end-of-load builds re-extract at the moment of failure** through a new `recapture()` accessor, instead of the parse holding the stream against the chance of a fallback that almost never runs.
- **The two readers that read the array rather than its contents move to a counter.** The permanent `[conwayDirect] demand pump: batches=… meshes=…` line (Share#1744) reports meshes actually delivered, and the empty-pump sentinel tests that counter. This is not cosmetic: reading `captured.length` after this change would mistake a healthy 150-mesh streaming load for a no-op pump and re-extract the whole model. The line gains `retained=yes|no`.
- **`STREAMING_CONSUMER: true` is declared on the deferred open**, inert on the pinned engine and activating conway#657's no-retention contract once the pin moves. Re-grepped on 1.1575.649: still absent, so still inert.

## What `recapture()` actually returns — equivalent, not identical

1. **Grouping.** On a deferred model the pinned `StreamAllMeshes` serves per-entity **full** FlatMeshes out of conway's `meshMap`, not the per-batch **delta** FlatMeshes the pump delivered. Same placement set, different bundling — benign for both readers here, since the merged and batched builds iterate placements and are indifferent to how they arrive, but a future consumer keyed on batch identity would care.
2. **Completeness under a budget.** `GEOMETRY_BUDGET_MB` is set on this path, and once conway has evicted anything its whole-model serve filters out placements whose natives were freed (`livePlacements`). So on a model large enough to evict, `recapture()` returns a strict **subset** of what was pumped.

### What the recapture improvement actually is (revised on the new pin)

An earlier revision of this PR claimed the filtered stream saved a degraded build from an **embind abort**, because the old pin's `getGeometry` cloned a freed native unguarded (the Sentry SHARE-1NK shape). **That claim is retired — conway#654, which arrived with this branch's `main` merge, fixed it engine-side.** On 1.1575.649 `getGeometry` probes `isNativeDeleted` and wraps the clone in a try/catch, degrading an evicted placement to a **dummy `IfcGeometry`** with a `[GetGeometry]` error rather than aborting. Share's `flatMeshToBufferGeometry` already skips a zero vertex/index geometry and counts it in `skippedPlacedGeometries`.

So on this pin **both routes render the same model** — the geometry is gone either way, because the native was freed. What the filtered re-extraction buys is **accounting, not pixels**: conway drops the placement before delivery and emits one aggregate warning naming instance and entity counts, where the retained route logs one `[GetGeometry]` error per evicted placement and surfaces the loss only as a `skippedPlacedGeometries` bump. Worth having, and small. It is no longer crash avoidance, and the residual on the windowed path is correspondingly milder than previously stated — noisier accounting, not a reachable abort.

## The re-extraction mechanism, and where it is not available

`recapture()` is the identity where the stream was retained, and a **memoised `StreamAllMeshes` capture** where it was dropped. Memoised because `buildBatchedConwayModel` and `buildConwayIfcModel` run in sequence and a second whole-model walk on a live model doubles every cached triangle count — the defect `IfcItemsMap.js` documents from the consumer side.

**Retention is kept in two states, and the second is the honest limitation of this PR.**

1. **No `onMeshBatch`** — `captured` *is* the delivery, not a copy of it.
2. **A windowed (store-backed) deferred open.** Re-verified on 1.1575.649: `streamAllMeshes` on a deferred model drains through the **synchronous** `extractGeometryBatch`, which throws `ExtractGeometryBatch is synchronous and cannot page a windowed source` outright when `model[0].isSourceExternal` (`compiled/src/compat/web-ifc/ifc_api_proxy_ifc.js:1509`, reached from the drain loop at `:2632` — line numbers refreshed, #654 moved them). `ExtractGeometryBatchAsync` cannot substitute: after a full drain its cursor is exhausted and there is no public rewind, and it remains the only async geometry entry point on the pin.

**This constraint is not lifted by conway#657.** Its `recaptureWholeModel_` hangs off the same `streamAllMeshes` drain loop, which that PR leaves calling the sync `extractGeometryBatch` — so its "a late whole-model ask still works" claim does not hold for windowed sources on either version.

**Nor is there a route around it.** The three whole-model entry points that do *not* refuse a windowed deferred model are all worse: `loadAllGeometry` and `streamAllMeshesWithTypes` have no deferred branch and seed coordination from `model[5]`, which a deferred open never writes, so every instance lands in an identity frame — silently mis-framed geometry, worse than a refusal; `getFlatMesh` is per-entity and needs an ID enumeration the loader does not have. This is recorded in the code so the next reader does not "discover" `LoadAllGeometry` as the escape hatch.

**Consequence:** a GitHub/OPFS-backed load takes the windowed open by default (`Loader.js` hands `parse` the `File` itself), so today this frees the stream on **buffered opens only** — not the large models that motivated #638. Post-pin-bump a windowed model is no worse than today, just no better.

## A pre-existing defect this change documents but does not fix

The empty-pump sentinel is reached whenever `pumpedMeshes === 0`, and that does **not** imply conway fell back to a classic open: the pump loop exits on `remaining === 0 && extracted === 0` whatever the reason, so a genuinely **deferred** model with nothing to extract — a properties-only IFC, or one whose every product failed geometry — lands there too. On a windowed source that model's one-shot `StreamAllMeshes` takes the deferred branch and refuses to page, throwing out of the load.

This is **byte-equivalent to `main`**: the sentinel it replaced (`captured.length === 0`) selected exactly the same models and called exactly the same method. A test pins the current rejection so an engine-side fix has a red-to-green. Not fixed here; being filed separately.

## What this only hopes

- **The bytes.** No memory number is claimed. Dropping one spine frees ~4.4 MB of the measured 475 MB, which is the *predicted* size of one 562,351-entry spine at 8 B — a confirmation of the model, not a result. The real figure needs all three spines dropped in one experiment on a Share build against a conway carrying #657, and is owed by the joint measurement.
- **That the flag does anything yet.** On this pin it is a dropped unknown key.
- **That the windowed half lands.** It does not, here.

## Falsifiability evidence

Four targeted reverts, each observed red.

**Revert A — restore the unconditional `captured.push(...batch)`.** 3 failed / 44 passed. Two tests fail on `expect(result.captured).toHaveLength(0)` — received 150 and 20.

**Revert B — point the boundary log and the sentinel back at `captured.length`.** 4 failed / 43 passed, and these are the two silent regressions named in #638's verification section:
```
counter fidelity:  Expected substring: "meshes=150"
                   Received string:    "…demand pump: batches=3 meshes=0 onMeshBatch=yes … retained=no"
sentinel misfire:  StreamAllMeshes — Expected number of calls: 0
                                     Received number of calls: 1
```
The second is a *healthy* 150-mesh streaming load being mistaken for an empty pump and re-extracted whole.

**Revert C — point the degraded builds back at `captured`.** 5 of 6 fail in `ShareIfcLoader.degraded.test.js`: every degraded build receives `[]` instead of the 3-mesh stream. That is the render-nothing regression. The healthy-path test still passes, so it pins the fallback specifically.

**Revert D — never increment `pumpedMeshes`.** 6 failed / 42 passed, confirming the counter assertions on the *retained* branch are falsifiable too (where `captured.length` would coincidentally agree).

One test is deliberately **not** falsifiable against this diff: "surfaces the engine refusal when a DEFERRED windowed model pumps nothing" characterises pre-existing behaviour, and exists so the engine-side fix has a red-to-green. It is non-vacuous — it also asserts the sentinel was actually reached (`StreamAllMeshes` called once, warning emitted).

## Numbers

- **Baseline on the untouched tree at `620af58`:** `yarn lint` clean; `yarn test` **258 suites / 2644 tests** — 2639 passed, 5 skipped, exit 0.
- **After, on the merged tree at conway 1.1575.649:** `yarn lint` clean; `yarn test` **259 suites / 2662 tests** — 2657 passed, 5 skipped, 13 snapshots, exit 0. `+1 suite, +18 tests`, nothing pre-existing red. The husky gate ran this on the merged tree.
- **Pre-flip validation** (run before the merge, on 1.1565.608): `yarn build-prod` exit 0; `yarn test-ci` exit 0 with coverage thresholds enforced; Playwright `conwayDirect.spec.ts` + `indexStepLogo.spec.ts` + `loadTiming.spec.ts` — **13 passed**, including a real GitHub-sourced IFC load through the demand pump that streamed and assembled normally (`Geometry: 0.250s`, `vertices=5689 triangles=3326`). Gated CI re-runs these against the new pin on this push.
- `package.json`'s build version stamp was restored each time; the tree is clean.

## Expectations legitimately changed (rubric #9)

- `conwayDirectIfcLoader.test.js` "opens deferred and pumps batches to completion" asserted `captured` had 150 with an `onMeshBatch` present. That is now 0 by design; the assertion was inverted and re-commented.
- `CadView.test.jsx`'s module double for `conwayDirectIfcLoader` returned `{modelID, captured}` and now returns `recapture` too. **This was a real caught defect, not a mechanical update:** the stale double made `ShareIfcLoader#parse` throw `recapture is not a function`, failing two CadView tests. It reproduced in isolation and passed on the untouched tree, so causation was confirmed before it was fixed. Every production return site carries `recapture`; no defensive `?? captured` was added, because that would convert exactly this class of bug into a silently empty model.
- The `GEOMETRY_BUDGET_MB` docstring predicted #654's copy-window fix as pending. It has landed, so the note now records it as closed on this pin rather than leaving a reader waiting for it. Pre-existing text, corrected because the merge made it false.

## Tests

`src/viewer/ifc/conwayDirectIfcLoader.test.js` — 12 new: the flag on the deferred open and its absence on the classic one; counter fidelity with zero retention *and* on the retained branch; retention where nothing takes delivery; retention on a windowed open whose `StreamAllMeshes` is stubbed with the real engine message (asserted never called); the drop when a windowed open falls back to buffered; on-demand re-extraction and its memoisation; the sentinel served off the counter; a genuinely empty pump still recovering with an `onMeshBatch` present; the deferred-windowed empty pump's current rejection; and `recapture` as the identity on the classic path.

`src/viewer/ifc/ShareIfcLoader.degraded.test.js` — 6 new, one per degraded path (`builder === null`, `!hasContent()`, `finalize()` throwing), the `batchedMesh` fallback, both fallbacks in sequence, and the healthy path asserting the accessor is **never** called — because re-extraction on every load would just trade retention for a whole-model re-walk.

## Review focus (the load-bearing claims to attack)

- **Fallback correctness on windowed sources.** The central claim: a windowed deferred model cannot serve a post-pump whole-model re-read on this pin *or* after conway#657, so retention there is necessary rather than timid. Evidence is `ifc_api_proxy_ifc.js:1509` and the unchanged drain loop in #657's diff. **Attack:** is there a path I missed — a rewind, an async whole-model entry, a `getFlatMesh` route that does not reach the sync pump?
- **Is `windowedSource` inferred correctly?** It is set from Share's own successful `OpenModelStream(store, …)` rather than asked of the engine, because `sourceIsExternal` is a proxy getter not exposed through the `IfcApi` shim. I argue nothing later flips it — `spillModelSource` runs from the GLB writer's `finally`, long after `parse` returns. Break that ordering claim and the retention decision is wrong for the window that matters.
- **Counter fidelity of the boundary log.** `pumpedMeshes` is incremented in the same branch that used to push, so it should equal the old `captured.length` exactly. Worth reading for an early `continue` or a batch counted twice.
- **Empty-pump sentinel preservation.** The sentinel moved from `captured.length === 0` to `pumpedMeshes === 0` and its branch now returns early with unconditional retention. The claim is that this is byte-equivalent in model selection, including for the deferred-windowed case documented above.
- **The revised improvement claim.** I now assert the filtered re-extraction buys accounting rather than pixels on this pin, because #654 made `getGeometry` degrade instead of abort and `flatMeshToBufferGeometry` already skips the dummy. If a dummy geometry can reach a consumer that does *not* skip it, that claim is too weak and the improvement is larger than stated.
- **The diagnostics.** `ifcItemsMapParity` and the instancing analysis now cost a real whole-model re-read on a dropped stream. I judged that better than reporting zeros that look like a finding, and gated the instancing one behind a thunk so a closed gate pays nothing.
- **`#1798`/`#1799`'s semantics are untouched.** `incrementalBatchedBuilder.js` is not in this diff at all; the staged boundary reads, `failedThisBatch` vs `badGeometry`, and `seenPlacements.add` ordering are unchanged, and the merge of `main` was clean (`package.json` + `yarn.lock` only).

## Sequencing

conway#657 publishes; then Share's pin bump to a conway carrying it; then the joint three-spine measurement. bldrs-ai/conway#638 stays open.

### DIFF
--- a/src/Containers/CadView.test.jsx
+++ b/src/Containers/CadView.test.jsx
@@ -45,8 +45,13 @@ jest.mock('../search/SearchIndex')
 // wit-three-shape fake the rest of the CadView assertions expect).
 // Scoped to this test file because production unit tests for
 // `buildConwayIfcModel` need the real implementation.
+// `recapture` is part of the parse's return contract, not an optional
+// extra: since conway#638 the streaming path keeps no copy of the mesh
+// stream, so it — not `captured` — is what the degraded end-of-load builds
+// read. A double that omits it makes `ShareIfcLoader#parse` throw
+// "recapture is not a function" on the fallback path.
 jest.mock('../viewer/ifc/conwayDirectIfcLoader', () => ({
-  parseIfcWithConway: () => ({modelID: 0, captured: []}),
+  parseIfcWithConway: () => ({modelID: 0, captured: [], recapture: () => []}),
   decorateConwayDirectIfcModel: () => {},
 }))
 jest.mock('../viewer/ifc/buildConwayIfcModel', () => ({
--- a/src/viewer/ifc/IfcItemsMap.js
+++ b/src/viewer/ifc/IfcItemsMap.js
@@ -283,6 +283,15 @@ export function itemsMapFromOrderedRanges(ranges, opts = {}) {
  * (Phase 3) uses `itemsMapFromConwayStream` directly because there
  * is no prior walk to collide with.
  *
+ * conway#657 closes this at the source for models opened with
+ * `STREAMING_CONSUMER` (which `conwayDirectIfcLoader` now declares on every
+ * deferred open): there is no cache for the walk to re-push into, so
+ * conway's own re-walk clears and rebuilds instead of appending and the
+ * second call returns what the first did. It stays true on the pinned
+ * engine and on every non-streaming open, so this entry point is not going
+ * away — but a doubling seen after the pin bump is a bug rather than the
+ * documented behaviour.
+ *
  * @param {object|Array} flatMeshes FlatMesh source (size()/get(i) or Array)
  * @param {object} api Conway-compatible IfcAPI (needs GetGeometry)
  * @param {number} modelID
--- a/src/viewer/ifc/ShareIfcLoader.degraded.test.js
+++ b/src/viewer/ifc/ShareIfcLoader.degraded.test.js
@@ -1,0 +1,298 @@
+// conway#638: the demand pump's FlatMesh stream is no longer retained once
+// something else takes delivery, so `ShareIfcLoader#parse`'s DEGRADED
+// end-of-load builds read `recapture()` — a whole-model accessor that
+// re-extracts at the moment of failure — instead of the parse's `captured`
+// array. These pin the wiring: on every degraded path the fallback build
+// must still receive the whole stream, and on the healthy path it must not
+// be asked for at all.
+//
+// The retention decision itself (which opens drop, which keep, and why a
+// windowed source cannot) lives one layer down and is pinned in
+// `conwayDirectIfcLoader.test.js`. Here `parseIfcWithConway` is mocked so
+// the two halves of its return value can disagree, which is exactly the
+// state a reader of the wrong one would get wrong.
+
+import ShareIfcLoader from './ShareIfcLoader'
+import {assembleBatchedModel, buildBatchedConwayModel} from './buildBatchedConwayModel'
+import {buildConwayIfcModel} from './buildConwayIfcModel'
+import {IncrementalBatchedBuilder} from './incrementalBatchedBuilder'
+import {parseIfcWithConway} from './conwayDirectIfcLoader'
+import {flatMeshToInstancedModel} from './flatMeshToInstancedModel'
+import {isFeatureEnabled} from '../../FeatureFlags'
+
+
+jest.mock('./conwayDirectIfcLoader', () => ({
+  parseIfcWithConway: jest.fn(),
+  decorateConwayDirectIfcModel: jest.fn(),
+}))
+jest.mock('./buildBatchedConwayModel', () => ({
+  buildBatchedConwayModel: jest.fn(),
+  assembleBatchedModel: jest.fn(),
+}))
+jest.mock('./buildConwayIfcModel', () => ({buildConwayIfcModel: jest.fn()}))
+jest.mock('./incrementalBatchedBuilder', () => ({IncrementalBatchedBuilder: jest.fn()}))
+jest.mock('./flatMeshToBufferGeometry', () => ({flatMeshToBufferGeometry: jest.fn()}))
+jest.mock('./flatMeshToInstancedModel', () => ({flatMeshToInstancedModel: jest.fn()}))
+jest.mock('./parsePreviewMesh', () => ({payloadToPreviewMesh: jest.fn(() => null)}))
+jest.mock('./ifcItemsMapParity', () => ({runIfcItemsMapParityCheck: jest.fn()}))
+jest.mock('./conwayDirectLog', () => ({
+  conwayDirectInfo: jest.fn(),
+  conwayDirectError: jest.fn(),
+}))
+jest.mock('../../FeatureFlags', () => ({isFeatureEnabled: jest.fn()}))
+jest.mock('../../utils/location', () => ({hasParams: jest.fn(() => false)}))
+
+// The session owns preview lifecycle and camera follow, none of which this
+// file exercises — but `previewGroup` being non-null is what makes
+// `ShareIfcLoader#parse` build an `onMeshBatch` at all, and that callback's
+// existence is the whole precondition for the stream being dropped.
+jest.mock('../ProgressiveLoadSession', () => jest.fn().mockImplementation(() => ({
+  previewGroup: {add: jest.fn(), remove: jest.fn(), children: []},
+  report: jest.fn(),
+  beginAssembly: jest.fn(),
+  notifyBounds: jest.fn(),
+  addPreviewMesh: jest.fn(),
+  setSummary: jest.fn(),
+  finish: jest.fn(),
+  abort: jest.fn(),
+})))
+
+
+/** The stream the pump delivered — what every degraded build must receive. */
+const PUMPED = [
+  {expressID: 101, geometries: {size: () => 1}},
+  {expressID: 102, geometries: {size: () => 1}},
+  {expressID: 103, geometries: {size: () => 1}},
+]
+
+const BUILD_STATS = {
+  vertexCount: 9,
+  triangleCount: 3,
+  instanceCount: 3,
+  parentCount: 1,
+  materialCount: 1,
+  skippedFlatMeshes: 0,
+  skippedPlacedGeometries: 0,
+}
+
+
+/**
+ * A model object shaped enough for the tail of `parse` (matrix stamp,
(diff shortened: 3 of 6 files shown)


pablo-mayrgundter:
@codex review

Reviewer context — load-bearing claims to attack, per the PR's review-focus section: (1) the windowed-vs-buffered split — retention must remain on windowed opens (no re-extraction exists there on the pinned engine; a drop would render nothing on degraded paths), and the split must be decided by the real source mode, not a proxy for it; (2) counter fidelity — the `[conwayDirect] demand pump` boundary log and the empty-pump sentinel must report/act on real counts with `captured` no longer accumulating on the streaming path (the sentinel must not misfire on a healthy streaming load); (3) degraded-path correctness on buffered opens — `builder === null` / `!hasContent()` / finalize-throw must still produce a full model via re-extraction, and the re-extraction must not double-append or reorder against what the incremental builder already consumed.

---

PR #1801, opened 2026-08-30
Bump conway to 1.1578.666-g39d59784, activating STREAMING_CONSUMER

pablo-mayrgundter:
## What this bump carries

`@bldrs-ai/conway` `1.1575.649-gdeabdf16` → `1.1578.666-g39d59784` (npm latest; version chain `deabdf16 → 8f0c3963 → fa964db6 → 39d59784`, nothing else in between).

1. **conway#657 — `STREAMING_CONSUMER` activates for the first time on Share's pin.** #1800 declared `STREAMING_CONSUMER: true` in `conwayDirectIfcLoader.js`'s `deferSettings` while the pinned engine didn't understand the key (an inert unknown setting). This pin is the first to carry conway#657, so the no-retention contract is now live: a `STREAMING_CONSUMER` deferred open keeps no reference to the meshes the pump delivers (this loader's `captured`/`recapture` spine is the sole owner), and `loadAllGeometry`/`getFlatMesh` on such a model now route through `streamAllMeshes`'s re-walk instead of silently mis-framing from `model[5]`.
2. Two conway-geom fixes that can legitimately shift vertex/triangle counts on curved models: **#655** (closed-trim b-spline re-seed) and **#666** (wrong-sheet inverse solve recovery).

## Comments updated

Reworded the "inert until the pin bumps past conway#657" comments in `conwayDirectIfcLoader.js`, `conwayDirectIfcLoader.test.js`, and `IfcItemsMap.js` to state the pin now carries the contract, and refreshed the `ifc_api_proxy_ifc.js` line-number citations that moved in the new build (`:1509`→`:1527`, `:2632`→`:2771`).

## Verification — the flag's first contact with the real engine

- `yarn lint`: clean.
- `yarn test`: **259 suites / 2662 tests — 2657 passed, 5 skipped, 13 snapshots** — identical to the pre-bump baseline (Jest mocks the engine, so this was expected to be unaffected).
- `yarn build-prod`: succeeded.
- `yarn test-ci`: same counts as `yarn test` above, coverage thresholds met.
- `yarn test-flows --workers=1` on the three specs from #1800 (`src/loader/conwayDirect.spec.ts`, `src/Containers/indexStepLogo.spec.ts`, `src/viewer/loadTiming.spec.ts`): **13/13 passed.** `demandGeometry` is default-on, so every one of these already exercises the demand pump under `STREAMING_CONSUMER` against the real engine — no test needed a flag override.
[...]

Issue #661, opened 2023-03-16
FS: Add file navigation to Open dialog


OlegMoshkovich:
FIle System: Add Organizations, repositories and files select components to openDialog

Issue #660, opened 2023-03-16
FS: Fetch directory from GitHub


OlegMoshkovich:
File System: Fetch user's organizations, repositories and files from GitHub

PR #1803, opened 2026-08-30
conway#660: drop windowed retention through the async whole-model ask, feature-detected

pablo-mayrgundter:
Part of bldrs-ai/conway#660 (and so of bldrs-ai/conway#635). **Share half only** — the pin bump that activates it and the PSB before/after that issue's A/B comment sets up are both still owed, so this carries no closing keyword.

## What this establishes

Share#1800 dropped the pumped FlatMesh spine — Share's own third pointer over the ~475 MB PlacedGeometry graph — on **buffered** deferred opens only. Its replacement for retention is re-extraction at the moment of failure (`recapture()`), and that re-extraction was `StreamAllMeshes`, which on a deferred model drains through the **synchronous** `ExtractGeometryBatch` and refuses a windowed source outright (*"ExtractGeometryBatch is synchronous and cannot page a windowed source"*). GitHub/OPFS `File` loads take the windowed open by default, so **the large models the epic is aimed at were exactly the ones still retaining** — measured on PSB (860.7 MB) as ~zero bytes reclaimed with two of three spines dropped (conway#660's A/B comment: +2,198.0 MB inert vs +2,205.0 MB active).

conway#672 added `StreamAllMeshesAsync(modelID, callback)`, which pages the window through `ExtractGeometryBatchAsync` and then runs the *same* post-drain serve the sync entry point runs. This consumes it, feature-detected:

| engine pin | buffered deferred open | windowed deferred open | `recapture()` drives |
|---|---|---|---|
| **without** `StreamAllMeshesAsync` (incl. the current pin `1.1578.666-g39d59784`) | drops (`retained=no`) | **retains** (`retained=yes`) | `StreamAllMeshes` |
| **with** it | drops | **drops** (`retained=no`) | `StreamAllMeshes` buffered, `StreamAllMeshesAsync` windowed |

**The current pin does not export it**, verified by grep over `node_modules/@bldrs-ai/conway` (zero hits; the sync `StreamAllMeshes` is there in both proxies). So the new regime is **dormant on this branch** and activates on the next routine bump — the land-then-activate pattern of #1801. Every old-pin behaviour is pinned unchanged by tests that run against a stub *without* the async ask, and revert **B** below (force the detect true) is what shows those pins are load-bearing rather than decorative.

## This is not purely additive on the new pin — one deliberate behaviour change, twice

Dormancy means nothing changes *today*. It does not mean the bump changes nothing, and two windowed cases behave differently once the ask is there. Both are intended.

**1. A fully-evicted windowed model now fails the degraded build loudly instead of rendering whatever survived.** Old windowed path: `captured` was retained, so a degraded build got the pump's own array back and rendered it — including placements whose natives a `GEOMETRY_BUDGET_MB` eviction had already freed, which `flatMeshToBufferGeometry` skips for zero vertex/index size. Partial or empty, silently, counted only as a `skippedPlacedGeometries` bump. New windowed path: the ask re-walks the geometry store, and conway#672's contract is *"the same stream the pump delivered, or an exact account of what is missing"* — partial loss serves the survivors with a warning naming the exact unresolved count, and **total loss throws**. So `parse` ends the load (null + `onError` + `ifcLastError`, the documented contract; OOM alone re-throws) rather than presenting an empty scene as a successful load. That is conway's "never a silent empty model" arriving in Share. It only bites when a degraded build is *already* running — the incremental assembly failed — on a model big enough for the budget to have evicted everything the re-walk needs, so it converts a bad-and-silent outcome into a bad-and-visible one. Pinned by `fails the load loudly when the whole-model ask rejects` and, at the loader level, by test 8 below.

**2. A properties-only windowed IFC loads instead of failing** — the sentinel change, below.

## The recapture seam: a Promise on every path

`recapture()` used to return the array synchronously. On a windowed source the re-extraction pages bytes off a `File`/OPFS store, so it *cannot* be synchronous, and the accessor now returns a `Promise` on **every** path — retained, buffered-sync, windowed-async.

A union return ("array on the old pin, Promise on the new one") would have worked by accident, because `await` normalises either. It was rejected: a missing `await` would then be invisible on the pin the developer is running and a silent wrong-type feed on the other — the builders iterate placements, so a Promise reaches them as an empty model, not an error. Uniform means a dropped `await` fails everywhere, immediately.

All four call sites, each verified to be in an async flow:

| call site | `ShareIfcLoader.js` | change |
|---|---|---|
| batchedMesh degraded build | `buildBatchedConwayModel(recapture(), …)` | `await recapture()` |
| merged degraded build | `buildConwayIfcModel(recapture(), …)` | `await recapture()` |
| `ifcItemsMapParity` diagnostic | `runIfcItemsMapParityCheck(…, recapture())` | `await recapture()` |
| instancing analysis | `logInstancedModelStats(…, recapture, …)` | the thunk stays a thunk (the gate must still be able to turn the whole-model read away); the function became `async` and awaits it inside its own try, and the caller awaits it so the load log's line order stays deterministic |

Only the second of those propagates a rejection: the first sits inside the batchedMesh branch's own try/catch, correctly — that branch's contract is "never break a load" — so a rejection there falls through to the merged build, which asks again, gets the **memoised** rejection, and carries it out. End state identical to the old synchronous form, with one drain instead of two.

The memo holds the **promise**, not the resolved array. Two readers that ask before the first ask resolves therefore still drive the engine once — an overlap only the async path can reach. conway#672 serialises overlapping asks internally too, but that *queues* a second full drain plus scene walk; this avoids asking.

## The sentinel decision (conway#661's Share-side symptom): taken

`pumpedMeshes === 0` does **not** imply conway fell back internally to a classic open — a genuinely deferred model with nothing to extract (properties-only IFC, or one whose every product failed geometry) exits the pump loop the same way. On a windowed source the sentinel's one-shot sync `StreamAllMeshes` then takes conway's deferred branch and throws the load away. #1800 pinned that as a pre-existing defect it neither introduced nor fixed.

With the async ask present the same sentinel **can** serve, so it does: one feature-detected branch, no new machinery, and it covers both sub-cases (the genuinely deferred model through the async drain; an internally-fallen-back classic one through the sync path conway delegates to internally — "safe on a resident source and on a non-deferred model", conway#672). A properties-only windowed IFC therefore loads on a pin with the ask instead of failing. Contained, so taken. **On a pin without it the throw is unchanged and still pinned** (`surfaces the engine refusal when a DEFERRED windowed model pumps nothing`), and the comment there now says which regime is which rather than describing the defect as unconditional.

Buffered opens keep the sync call on both regimes, deliberately: the async ask is wired **only** where the sync one cannot serve, so a pin bump cannot silently re-route the path that already worked onto a second implementation of the same answer. Revert **C** is that claim's discriminator.

## Observability

The permanent boundary line gained two tokens, so a production trace says which regime it measured without anyone reading a version string — which is what the PSB before/after will be read off:

```
[conwayDirect] demand pump: batches=3 meshes=150 onMeshBatch=yes onPreviewMesh=no windowed=yes asyncAsk=no retained=yes
```

`asyncAsk` is the **engine's capability**, not the entry point this load drives — nothing has been asked yet at that point, and on a buffered open `asyncAsk=yes` still drives the sync ask. The entry point actually driven is named by `recaptured N mesh(es) via StreamAllMeshes|StreamAllMeshesAsync`, which only a degraded build emits. (Review P3-3: the token was first called `wholeModelAsk=async|sync`, which read as a claim about the ask that ran.)

## What this only hopes

- **The bytes.** No memory number is claimed here. This makes the drop *reachable* on windowed opens; whether it pays out is the PSB before/after conway#660 owes, and that measurement cannot even run until the pin carries the ask.
- **That the ask serves what the pump served.** conway#672 measured it placement-for-placement on a windowed drained model (16/16, identity + full transform + colour + `occurrencePath`), and Share's side of that is a stub, not the engine. Jest mocks the engine here by construction.
- **That the new path has run in a browser.** It has not, and cannot until the bump — see the CI note below.

## Falsifiability

Each revert applied to the working tree, the suites re-run, then restored. All 63 tests in the two loader suites pass with none applied.

| revert | what it removes | result |
|---|---|---|
| **A** | the feature detect never fires (`hasAsyncWholeModelAsk = false`) — the pre-#660 world | **6 failed / 56 passed** |
| **B** | the feature detect always fires (`= true`) — no dormancy on the current pin | **2 failed / 60 passed** |
| **C** | the windowed condition on entry-point choice (`recaptureViaAsyncAsk = hasAsyncWholeModelAsk`) | **1 failed / 61 passed** |
| **D** | the retention decision as #1800 left it (`onMeshBatch === undefined \|\| windowedSource`) | **5 failed / 57 passed** |
| **E** | promise memoisation (memoise the resolved array instead) | **2 failed / 60 passed** |
| **F** | the sentinel's async branch (always the sync entry point) | **1 failed / 61 passed** |
| **G** | one `await` at a call site (`buildConwayIfcModel(recapture(), …)`) | **4 failed / 58 passed** |
| **H** | the merged degraded build's unguarded-ness (wrap it in a try/catch) | **1 failed / 6 passed** |

(A–G measured at 62 tests, before the P3 pass added the rejection test; H after.)

- **A** fails exactly the six new-regime tests and nothing else — the old-pin suite is untouched by the new path, which is the shape dormancy has to have.
- **B** is the byte-identity pin, and the important one: it fails `retains the stream on a WINDOWED open with no async ask, where re-extraction throws` and `surfaces the engine refusal when a DEFERRED windowed model pumps nothing`. Those two are today's behaviour on today's pin; if the detect were wrong in the "always on" direction, this branch would change what ships now.
- **C** fails only `keeps the SYNC ask on a buffered open even when the async one exists`.
- **D** fails the five windowed-drop tests but not the sentinel one — the sentinel branch is a separate decision, which is why F exists.
- **E** fails `memoises the async ask across the two consecutive degraded builds` (two engine drives for one model) and `lets a total-loss throw from the ask reach the load, once` (the failed drain re-driven).
- **F** fails only `recovers an empty pump on a DEFERRED windowed model through the async ask`.
- **G** fails all four degraded-build tests in `ShareIfcLoader.degraded.test.js`, which is only possible because that file's `recapture` stub was made **async** in this PR. With the old sync stub a dropped `await` passes — the stub was the thing making the seam untestable. (Verified independently by the reviewer.)
- **H** fails only `fails the load loudly when the whole-model ask rejects`, which is the point: that test's claim is specifically that the SECOND await is unguarded, since the first one's rejection is swallowed by design.

## Tests

`conwayDirectIfcLoader.test.js` — a new nested describe, `windowed retention with the async whole-model ask (conway#660)`, 8 tests. Its stubs attach a real async `StreamAllMeshesAsync` and leave the sync `StreamAllMeshes` **throwing the engine's actual refusal message**, so routing to the wrong entry point fails loudly rather than passing by accident.

1. drops the stream on a WINDOWED open once the engine can serve it back (`retained=no`, `asyncAsk=yes`, and the ask is not paid unless asked)
2. serves a degraded build through the async ask, never the sync refusal (bound to the windowed handle, not the buffered fallback's)
3. memoises the async ask across the two consecutive degraded builds — including the concurrent-overlap case
4. keeps the SYNC ask on a buffered open even when the async one exists (and pins `asyncAsk=yes` beside it, so the token's engine-level meaning is asserted rather than assumed)
5. still retains a windowed stream when nothing else takes delivery (no `onMeshBatch` ⇒ `captured` IS the delivery, regardless of the engine)
6. recovers an empty pump on a DEFERRED windowed model through the async ask (the sentinel decision; its old-pin twin still pins the throw)
7. serves the budget subset the ask answers with, without inventing a throw (conway#672's 3-of-16-under-a-2-KiB-budget semantics)
8. lets a total-loss throw from the ask reach the load, once

`ShareIfcLoader.degraded.test.js` — `fails the load loudly when the whole-model ask rejects` (review P3-2): the loader-level half of test 8. `recapture` rejects, `batchedMesh` off so the unguarded await runs, and the assertions are `parse` → null, `onError` called with the error, `ifcLastError` stashed, `addIfcModel` never called. Its stub also went async (see revert G).

Existing tests: the six `recapture()` reads became `await recapture()`; the windowed-retention and sentinel tests gained `asyncAsk` / `windowed` assertions and comments naming which regime they are. `CadView.test.jsx`'s recapture double went async too (review P3-1) — the last synchronous one repo-wide. Its comment records the measured limit of that change: it does **not** turn that file into a dropped-`await` detector, because its `buildConwayIfcModel` double ignores its argument entirely.

## Numbers

- **Baseline, measured on the untouched tree at `b02aab6`:** `yarn lint` clean; `yarn test` **259 suites / 2662 tests** (2657 passed, 5 skipped), 98 s.
- **This head:** `yarn lint` clean; `yarn test` **259 suites / 2671 tests** (2666 passed, 5 skipped) — **+9** (8 new-regime + 1 rejection pin). The husky pre-commit gate ran the same set on both commits.
- **Pre-flip validation** (the three things the draft gate does not cover), run on `ecbfb97`: `yarn build-prod` ✅; `yarn test-ci` ✅ with coverage thresholds enforced (statements 71.08%, branches 65.84%, functions 72.85%, lines 70.39%); `yarn test-flows src/loader/conwayDirect.spec.ts src/Containers/indexStepLogo.spec.ts src/viewer/loadTiming.spec.ts --workers=1` ✅ **13 passed**. `package.json`'s version stamp was restored by hand afterwards (#1747). The P3 commit is test-only plus one log token, and re-ran the full jest gate.
- **Those Playwright runs exercise the CURRENT pin**, i.e. the *dormant* regime — real windowed loads (GitHub → OPFS) taking the retaining path exactly as they do on main. They are evidence that this change did not disturb what ships today; they are **not** evidence the new path works in a browser. Nothing can be until the pin bump, and that is the first thing to check after it.

## Review focus

- **The feature-detect seam** — `hasAsyncWholeModelAsk` / `recaptureViaAsyncAsk` in `parseIfcWithConway`. Is the split right that the async ask is used *only* on windowed sources, leaving buffered opens byte-identical on both pins?
- **The async recapture at each call site.** Four sites, one made async to reach its thunk. Is there a fifth reader of `recapture()` anywhere, and is awaiting `logInstancedModelStats` (rather than firing it) the right call for log ordering?
- **The sentinel decision** — taking conway#661's fix in the same branch that drops retention. Contained enough, or should it have been a separate PR against #661?
- **Old-pin byte-identity.** Revert B is the argument; is it argued at the right level, or does something observable still change on `1.1578.666-g39d59784`?
- **The behaviour change on the new pin** (the section above): trading a silent partial/empty render for a loud failure inside an already-degraded build.

### DIFF
--- a/src/Containers/CadView.test.jsx
+++ b/src/Containers/CadView.test.jsx
@@ -50,8 +50,24 @@ jest.mock('../search/SearchIndex')
 // stream, so it — not `captured` — is what the degraded end-of-load builds
 // read. A double that omits it makes `ShareIfcLoader#parse` throw
 // "recapture is not a function" on the fallback path.
+//
+// ASYNC, matching the real seam (conway#660): `recapture()` returns a
+// Promise on every path, because a windowed re-extraction pages bytes off
+// the byte store and cannot be synchronous. The sync form passed only
+// because `await` normalises a plain array.
+//
+// Stated precisely, because the obvious stronger claim is false: making
+// this double async does NOT give this file a dropped-`await` detector.
+// Measured — with the `await` removed from the merged degraded build, this
+// whole file stays green either way, because the `buildConwayIfcModel`
+// double below ignores its argument entirely and hands back the harness
+// singleton. What guards that call site is
+// `ShareIfcLoader.degraded.test.js`, whose builder doubles assert on the
+// meshes they receive. This double is here to stop a fixture from
+// contradicting the contract it stands in for, not to catch a regression.
 jest.mock('../viewer/ifc/conwayDirectIfcLoader', () => ({
-  parseIfcWithConway: () => ({modelID: 0, captured: [], recapture: () => []}),
+  // eslint-disable-next-line require-await
+  parseIfcWithConway: () => ({modelID: 0, captured: [], recapture: async () => []}),
   decorateConwayDirectIfcModel: () => {},
 }))
 jest.mock('../viewer/ifc/buildConwayIfcModel', () => ({
--- a/src/viewer/ifc/ShareIfcLoader.degraded.test.js
+++ b/src/viewer/ifc/ShareIfcLoader.degraded.test.js
@@ -145,7 +145,14 @@ describe('viewer/ifc/ShareIfcLoader degraded end-of-load builds (conway#638)', (
     // does. Without that the loader's `builder` would stay null for every
     // case, and the three degraded paths this file separates would collapse
     // into one.
-    recapture = jest.fn(() => PUMPED)
+    // ASYNC on purpose, matching the real seam (conway#660): on a windowed
+    // source the re-extraction pages the byte window through
+    // `StreamAllMeshesAsync`, so `recapture()` returns a Promise on every
+    // path. A sync stub here would let a call site that dropped its `await`
+    // pass — the builders would receive an array either way — which is
+    // exactly the regression this file exists to catch.
+    // eslint-disable-next-line require-await
+    recapture = jest.fn(async () => PUMPED)
     parseIfcWithConway.mockImplementation(
       // eslint-disable-next-line require-await
       async (buffer, api, settings, onProgress, onMeshBatch) => {
@@ -290,9 +297,13 @@ describe('viewer/ifc/ShareIfcLoader degraded end-of-load builds (conway#638)', (
     expect(buildBatchedConwayModel.mock.calls[0][0]).toEqual(PUMPED)
     expect(buildConwayIfcModel.mock.calls[0][0]).toEqual(PUMPED)
     // Every array handed out is the one object the accessor memoises, so
-    // the two builds cannot be reading different re-extractions.
+    // the two builds cannot be reading different re-extractions. Each
+    // recorded result is the accessor's Promise, so resolve before
+    // comparing — `toBe` on the Promises themselves would pass on the
+    // memoised seam and equally on a broken one that re-drove the engine
+    // and happened to return the same array.
     for (const result of recapture.mock.results) {
-      expect(result.value).toBe(PUMPED)
+      expect(await result.value).toBe(PUMPED)
     }
   })
 })
--- a/src/viewer/ifc/ShareIfcLoader.degraded.test.js
+++ b/src/viewer/ifc/ShareIfcLoader.degraded.test.js
@@ -278,6 +278,46 @@ describe('viewer/ifc/ShareIfcLoader degraded end-of-load builds (conway#638)', (
     expect(buildBatchedConwayModel).not.toHaveBeenCalled()
   })
 
+  it('fails the load loudly when the whole-model ask rejects', async () => {
+    // conway#660/#672: on a windowed source `recapture()` drives
+    // `StreamAllMeshesAsync`, which THROWS rather than serving an empty
+    // model when a `GEOMETRY_BUDGET_MB` eviction freed everything the
+    // re-walk needs. That rejection has to end the load, not be absorbed
+    // into a blank scene reported as success — the failure mode that costs
+    // a user their file.
+    //
+    // The chain is worth naming because only half of it propagates. The
+    // FIRST degraded `await recapture()` sits inside the batchedMesh
+    // branch's try/catch, so a rejection there is swallowed into
+    // fall-through (correctly — that branch's contract is "never break a
+    // load"); it is the SECOND, unguarded await feeding the merged build
+    // that carries the failure out. This runs with `batchedMesh` off, which
+    // is the shipped default, so it exercises that unguarded await
+    // directly.
+    //
+    // "Loudly" is `ShareIfcLoader#parse`'s documented error contract, not a
+    // rejection: it catches, stashes `ifcLastError`, calls `onError` and
+    // returns null (re-throwing only on OOM, so `Loader.js#readModel` can
+    // show the tailored message). What must never happen is a model
+    // reaching the scene.
+    IncrementalBatchedBuilder.mockImplementation(() => {
+      throw new Error('builder construction failed')
+    })
+    const evicted = new Error(
+      'StreamAllMeshesAsync: this model was opened with STREAMING_CONSUMER, ' +
+      'and the re-walk resolved none of the model')
+    recapture.mockRejectedValue(evicted)
+    const onError = jest.fn()
+    const model = await loader.parse(new ArrayBuffer(4), jest.fn(), onError)
+    expect(model).toBeNull()
+    expect(onError).toHaveBeenCalledWith(evicted)
+    expect(loader.ifcLastError).toBe(evicted)
+    // The load is over, and nothing was added to the scene — no partial or
+    // empty model presented as a successful one.
+    expect(ifc.addIfcModel).not.toHaveBeenCalled()
+    expect(buildConwayIfcModel).not.toHaveBeenCalled()
+  })
+
   it('serves both fallbacks off the accessor when they run in sequence', async () => {
     // `buildBatchedConwayModel` throwing falls through to
     // `buildConwayIfcModel`, so both degraded builds run in one load. Each
--- a/src/viewer/ifc/ShareIfcLoader.js
+++ b/src/viewer/ifc/ShareIfcLoader.js
@@ -55,18 +55,23 @@ import debug, {DEBUG, WARN, isLogEnabled} from '../../utils/debug'
  *
  * @param {object} ifcAPI Conway IfcAPI bound to the model
  * @param {number} modelID
- * @param {Function} getCaptured returns the parse's whole FlatMesh stream.
- *   A thunk rather than the array because on the streaming path
- *   (conway#638) producing it costs a whole-model re-extraction, and that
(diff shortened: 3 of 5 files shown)


pablo-mayrgundter:
@codex review

Reviewer context — the Share half of conway#660 (feature-detected `StreamAllMeshesAsync` consumption; dormant on the current pin). Load-bearing claims to attack: (1) **old-pin byte-identity** — with the detect false, every path must behave exactly as `main` does today (windowed retains, buffered drops, sentinel throws on a deferred-windowed empty pump); (2) **the recapture seam** — `recapture()` now returns a Promise on every path, memoising the promise: check every call site awaits it, that a rejected ask staying memoised is the right behavior (vs. retry-on-next-ask), and that no call site can observe the pre-#1803 synchronous timing in a way that matters; (3) **the sentinel's new async branch** — on a windowed model with the ask present, the empty-pump recovery serves through `StreamAllMeshesAsync`: verify the recovered stream feeds the same downstream builds as the sync recovery does, and that a REJECTED async recovery degrades the way the old throw did rather than a new silent-empty; (4) **the windowed-only routing** — buffered opens must keep the sync capture on both pins (the entry-point choice condition); (5) `retained=` log fidelity across the four regime×source combinations.

---

PR #1823, opened 2026-08-31
Re-bake batched subsets from three's own ranges, not a retained copy

### LINKED ISSUE #1810: Drop the duplicate source-geometry copy (`mesh.instanceGeometry`) after batch assembly — measured 171.5 MB on a 231 MB model
Byte-lever 1 from [conway#679's attribution](https://github.com/bldrs-ai/conway/issues/679#issuecomment-5474163735) — the largest Share-owned block in the settled footprint, measured exactly.

## What holds it

The geometry payload is stored **2.62×** after a windowed conway-direct load: three's BatchedMesh buffers hold the assembled copy (264.0 MB incl. slack — Share#1809 handles the slack), and a **full duplicate** of the source geometry survives as `mesh.instanceGeometry` — built by `localGeometry()`, held by `geometryCache` during assembly, and sliced onto the durable mesh at `finalize()` so `batchedSubset` can re-bake selection subsets from it. On `sp-231MB.ifc` that duplicate is **171.5 MB**, 24% of the entire external-ArrayBuffer bucket, retained for the life of the model to serve occasional selection re-bakes.

## The change

Re-bake subsets from three's own `_geometryInfo` ranges over the batch buffers instead of from the retained duplicate, then stop attaching `instanceGeometry` at finalize. The data is already resident in the batch — the subset path needs offsets, not a second copy.

With Share#1809, the two levers together take the settled heap **1,505.6 → 1,241.6 MB (−17.5%)** on the 231 MB model and geometry amplification **2.62× → 1.08×**.

## Cautions from the report

- Enumerate every reader of `instanceGeometry` before dropping it (the attribution found `batchedSubset`; there may be diagnostics).
- Blast-radius per the repo rule: selection/isolate/hide E2E behavior must stay byte-identical; the re-bake path needs its own test that fails when subsets are built from stale ranges.
- The settled external-bucket metric is noisy run-to-run (716–1,453 MB depending on settle timing); verify the win with the scene-traversal reading (stable, 264.0/171.5 split) and a post-GC snapshot, not the raw bucket number.

Part of the bldrs-ai/conway#635 follow-up family (byte-lever 1 of conway#679).

pablo-mayrgundter:
Part of bldrs-ai/conway#635; byte-lever 1 of bldrs-ai/conway#679

Fixes #1810

The batched model kept a second, complete copy of its own geometry. Each
builder pushed the local-space `BufferGeometry` it fed `addGeometry` into a
per-instance table, `finalize()` sliced that table onto the durable mesh as
`mesh.instanceGeometry`, and it stayed there for the life of the model so an
occasional isolate could re-bake a subset from it. `addGeometry` had already
copied every one of those bytes into the batch buffers and kept no reference
to the source (`BatchedMesh.setGeometryAt`, three r0.184
`BatchedMesh.js:743-792`), so the table bought nothing but a duplicate:
**171.5 MB on sp-231MB.ifc, 24% of the whole external-ArrayBuffer bucket**,
measured by scene traversal in the conway#679 attribution.

## Every reader of `instanceGeometry`, enumerated

Repo-wide grep, production and tests. The attribution named `batchedSubset`;
there were four readers, and re-pointing all four is what this PR is.

| reader | what it took from the table | where it reads now |
|---|---|---|
| `src/viewer/ifc/batchedSubset.js` | per-instance `position` / `normal` / `index` to bake isolation subsets | rebuilt from the batch range, one copy per selected shape, released when the bake returns |
| `src/viewer/ifc/batchedToMergedMesh.js` | same, to bake the merged GLB cache artifact | one memoising reader per collection pass |
| `src/loader/glbBatchedExport.js` | the geometry OBJECT, both for its arrays and as the dedup key (`geometry.uuid`) for shared accessors | same reader; identity restored by keying it on `instanceGeometryIds` |
| `src/viewer/residency/ResidencyController.js` | `boundingSphere`, `position.count`, and object identity for per-shape use counts | `getBoundingSphereAt` + `getGeometryRangeAt` — it never needed vertex data, only two numbers per shape |

Writers removed: `incrementalBatchedBuilder` (the `state.instanceGeometry`
accumulator and the `finalize()` slice), `flatMeshToBatchedModel` and
`instancedGlbToBatchedModel` (the `BatchHandle` field), and
`buildBatchedConwayModel.decorateBatchMeshes` (the stamp onto the mesh).
Test-only readers updated in place; the two hand-rolled doubles in
`glbBatchedExport.test.js` / `glbBatchedRoundTrip.test.js` became real
`BatchedMesh`es, because a plain object with an `instanceGeometry` array no
longer exercises anything.

`instanceGeometryIds` is a different table — `batchId → geometryExpressID`,
a `Uint32Array` — and stays.

## The re-bake

`src/viewer/ifc/batchedInstanceGeometry.js` resolves an instance through
three's **public** accessors, `getGeometryIdAt` and `getGeometryRangeAt`
(`BatchedMesh.js:1217` and `:1237` in the pinned r0.184 — nothing reaches
into `_geometryInfo`), and rebuilds the shape from the batch's own buffers:
slice `position` / `normal` over `[vertexStart, vertexStart + vertexCount)`,
and rebuild the index over `[indexStart, indexStart + indexCount)` minus
`vertexStart`, because three stores every batch index offset by the shape's
vertex start (`BatchedMesh.js:778`). That subtraction is the whole
correctness story, and it is what the byte-identity tests pin.

Two properties the consumers depended on and the module preserves:

- **Identity.** Both builders added one `BufferGeometry` per
  `geometryExpressID` and put that same object in every instance's slot,
  including across the opaque/transparent split, so the GLB writer's
  accessor dedup and the residency controller's use counts grouped by
  reference. `makeInstanceGeometryReader()` memoises per source shape, keyed
  on `instanceGeometryIds`, and reproduces exactly that grouping — for one
  pass, then releases it.
- **Freshness.** Ranges are read on every call and never snapshotted:
  `finalize()`'s capacity trim reallocates the batch buffers (Share#1809),
  and three's `optimize()` compacts shape offsets outright.

Dropping the table alone would have freed nothing. The incremental builder's
`geometryCache` is the *other* owner of those geometries, and the
attribution found it still reachable from `ShareIfcLoader.parse`'s closure
(104,991 entries) because conway's proxy pins the open `settings` object. So
`finalize()` now releases it beside `seenPlacements`, once the batches are
stamped and trimmed.

## Falsifiability

Every new assertion was run against a reverted mechanism.

| revert | test that goes red |
|---|---|
| drop the `- vertexStart` index rebase | `reads back each instance's source geometry byte for byte`, `bakes a subset identical to the pre-#1810 retained-table path`, `reflects the CURRENT ranges…` (3 failures) |
| cache the range per instance instead of re-reading it | `follows a shape whose range MOVES under it` |
| key the reader per mesh+geometryId instead of per source shape | `hands one geometry object per source shape to a reader, across batches` |

The parity test is the one the issue asks for: it builds one model, bakes a
subset the pre-#1810 way from a retained table (the old algorithm, kept in
the test) and the new way from the batch, and compares `position`, `normal`,
`expressID` and `index` array-for-array.

## Measurement

`sp-231MB.ifc`, windowed conway-direct load
(`?feature=conwayDirectIfc`), same box, same procedure before and after:
chromium with `--enable-precise-memory-info --js-flags=--expose-gc`, tab
backgrounded (a foreground tab rasterises through SwiftShader here and runs
~10x slower for identical allocation), wait for `store.model`, settle 30 s,
`gc()` x3, then one scene traversal that sums `attributes[*].array.byteLength`
+ index bytes, deduped by backing store — the stable instrument from the
attribution, not the run-to-run-noisy external-ArrayBuffer bucket. Both runs
loaded the same model: 5,601,739 vertices, 139,264 instance slots, identical
batch capacity (5,388,081 / 10,970,532 and 213,658 / 385,044 verts/indices).
The 256 MiB canary moved the heap counter 248.4 MB / 251.5 MB, so both reads
are live rather than latched.

| settled, post-GC | before | after | delta |
|---|---:|---:|---:|
| **`mesh.instanceGeometry` copies** | **171.5 MB** (104,991 geometries) | **0.0 MB** (0) | **-171.5 MB** |
| batch vertex attributes | 128.2 MB | 128.2 MB | 0 |
| batch indices | 43.3 MB | 43.3 MB | 0 |
| batch per-instance tables | 13.8 MB | 13.8 MB | 0 |
| wasm linear memory | 514.8 MB | 514.8 MB | 0 |
| `usedJSHeapSize` | 2,288.5 MB | 1,983.9 MB | **-304.5 MB (-13.3%)** |

171.5 MB and 104,991 geometries reproduce the attribution's figures exactly.

Two things worth naming rather than rounding off:

- **The heap fell by more than the geometry bytes** — 304.5 MB against
  171.5 MB of ArrayBuffers. The difference is the V8 side of the same
  release: 104,991 `BufferGeometry` objects, their three `BufferAttribute`
  wrappers each, and the `geometryCache` entry + `Box3` per shape, all of
  which the traversal above does not count. The attribution's own V8
  breakdown put `localGeometry` at 60.5 MB as the largest single allocation
  site; ~1.3 KB of object graph per shape accounts for the rest. So the
  scene-traversal number is the conservative one.
- **Geometry amplification is now 1.00x**, not the report's predicted 1.08x:
  the batch holds 171.5 MB to deliver 171.5 MB. The report measured 2.62x
  against a build without Share#1809, whose trim has since removed the 92.5
  MB of doubling slack — visible above as capacity == used in both runs.
  This lever removes the other duplicate, so 343.0 MB -> 171.5 MB stored for
  the same payload.

## Blast radius

- Selection / preselection never touched this path (they recolour in place
  via `batchedHighlight`); isolate / hide bake through `batchedSubset`, and
  the subset geometry is byte-identical by the parity test above.
- The two GLB writers produce the same artifact: same geometry values, same
  accessor dedup (identity preserved), same instance tables.
- `ResidencyController` is the one place with a behavioural note rather than
  a byte-identity claim: three's per-geometry sphere is computed over the
  shape's index range, while `BufferGeometry.computeBoundingSphere` walks
  every position. They agree for every shape whose vertices are all
  referenced — which is every shape Conway emits — and differ (tighter, more
  correct) only for a shape with orphan vertices. It feeds eviction
  ordering, not geometry.
- A batch with no index buffer is now refused up front by
  `hasBatchedGeometry` rather than per instance by the old `geom.index`
  check; same outcome (nothing to bake), one branch earlier.

## Validation

Baseline first, on `origin/main` at e2c38e0 before any edit: `yarn lint`
clean, `yarn test` 264 suites / 2802 tests (5 skipped).

- `yarn lint` (eslint + tsc): clean.
- `yarn test`: **265 suites / 2808 tests**, 5 skipped, 0 failures — +1 suite
  and +6 tests, all in `batchedInstanceGeometry.test.js`. Run again by the
  husky pre-commit hook on the commit itself.
- Playwright, against a `SHARE_CONFIG=playwright` build of this branch:
  `IfcIsolator.spec.ts` + `HideFeat.spec.ts` + `conwayDirect.spec.ts` — 12
  passed; `navTreeOccurrenceSelection.spec.ts` + `batchedGlbCache.spec.ts` +
  `Properties.cacheHit.spec.ts` — passed; `indexStepLogo.spec.ts` +
  `loadTiming.spec.ts` — 10 passed; `sceneHighlightPermalink.spec.ts` — 6
  passed.
- One confounder, named: in the 4-worker run, the three **desktop**
  `sceneHighlightPermalink` tests timed out waiting for `data-model-ready`
  (their mobile twins passed in the same run). Re-run alone with
  `--workers=1`, all 6 pass. This box has no GPU, so four concurrent
  SwiftShader contexts miss the spec's 15 s ready timeout; the failure is
  contention, not the diff. `yarn build-prod` / `yarn test-ci` /
  the full flows run are for the pre-flip pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ScPZ6RarXvYmSdz6BUXF78)_

### DIFF
--- a/design/new/model-display-controls.md
+++ b/design/new/model-display-controls.md
@@ -210,7 +210,8 @@ group, not a separate story.
 
 ## 5. Residency for every format (ask 4)
 
-`ResidencyController` requires `isBatchedMesh` + `instanceGeometry` +
+`ResidencyController` requires `isBatchedMesh` + readable batch buffers
+(`getGeometryRangeAt` / `getBoundingSphereAt`, see `batchedInstanceGeometry`) +
 `setVisibleAt`, so it renders nothing outside `?feature=batchedMesh` IFC/STEP.
 `ResidencyControl` self-gates on `instanceCount > 0`, which is the right shape
 — it just needs more ways to count.
--- a/design/new/viewer-replacement.md
+++ b/design/new/viewer-replacement.md
@@ -785,7 +785,11 @@ Three settled conclusions:
   - *Isolate / hide (`src/viewer/ifc/batchedSubset.js`).* `IfcIsolator`
     drives the batched model unchanged through a `createSubset` /
     `removeSubset` surface that re-bakes the kept instances (`getMatrixAt` +
-    the retained `instanceGeometry`) into world-aligned subset Meshes
+    the shape read back out of the batch's own buffers,
+    `src/viewer/ifc/batchedInstanceGeometry.js` — the retained
+    `instanceGeometry` table it used to bake from was a full duplicate of the
+    batch, 171.5 MB on a 231 MB model, dropped in Share#1810) into
+    world-aligned subset Meshes
     carrying a synthetic per-vertex `expressID` (so the isolated subset
     stays pickable). `visualElementsIds` is unioned from `instanceParents`.
   - *BVH picking.* `three-mesh-bvh` accelerated raycast validated on
--- a/design/roadmap.md
+++ b/design/roadmap.md
@@ -287,7 +287,7 @@ multi-worker; etc.*
   narrowing on the batched path; `EXT_mesh_gpu_instancing` batched-native GLB cache
   (§3b.v) so the round-trip stays instanced and the artifact shrinks.
 - Open: **residency backends** — the B2 slider is gated on the batched path
-  (`isBatchedMesh` + `instanceGeometry`), so it's invisible for GLB, OBJ and
+  (`isBatchedMesh` + readable batch buffers), so it's invisible for GLB, OBJ and
   friends. Extract a backend interface and add a scene-graph backend
   (`Object3D.visible`) plus a merged backend (index compaction over
   `IfcInstanceMap`'s per-instance triangle ranges) so the cached GLB re-load of
--- a/src/Components/Residency/ResidencyControl.test.jsx
+++ b/src/Components/Residency/ResidencyControl.test.jsx
@@ -35,9 +35,9 @@ const grey = () => ({x: DEFAULT_COLOR.x, y: DEFAULT_COLOR.y, z: DEFAULT_COLOR.z,
 
 /**
  * Batched-model double carrying the tables the color section reads. No
- * `instanceGeometry` / `setVisibleAt`, so `ResidencyController` finds nothing
- * to evict and the residency section stays hidden — which is exactly the case
- * that proves the color section gates independently.
+ * batch-geometry surface / `setVisibleAt`, so `ResidencyController` finds
+ * nothing to evict and the residency section stays hidden — which is exactly
+ * the case that proves the color section gates independently.
  *
  * @param {Array<object>} sourceColors the file's own colors
  * @return {object} model double
@@ -65,10 +65,21 @@ function colorOnlyModel(sourceColors = [grey(), grey(), grey()]) {
  */
 function residencyModel() {
   const model = colorOnlyModel()
-  model.instanceGeometry = [1, 2, 3].map((radius) => ({
-    boundingSphere: new Sphere(new Vector3(), radius),
-    getAttribute: () => ({count: radius * 100}),
-  }))
+  const radii = [1, 2, 3]
+  // The batch surface `ResidencyController` reads since Share#1810: one
+  // geometry per instance, sized and bounded through three's own per-geometry
+  // accessors instead of a retained `instanceGeometry` table.
+  model.geometry = {attributes: {position: {}}, index: {}}
+  model.getGeometryIdAt = (index) => index
+  model.getGeometryRangeAt = (geometryId, target = {}) => {
+    target.vertexStart = 0
+    target.vertexCount = radii[geometryId] * 100
+    target.indexStart = 0
+    target.indexCount = radii[geometryId] * 100
+    return target
+  }
+  model.getBoundingSphereAt = (geometryId, target) =>
+    target.copy(new Sphere(new Vector3(), radii[geometryId]))
   model.getMatrixAt = (index, matrix) => matrix.identity()
   model.setVisibleAt = jest.fn()
   return model
--- a/src/loader/glbBatchedExport.js
+++ b/src/loader/glbBatchedExport.js
@@ -1,4 +1,8 @@
 import {Color, Matrix4, Quaternion, Vector3} from 'three'
+import {
+  hasBatchedGeometry,
+  makeInstanceGeometryReader,
+} from '../viewer/ifc/batchedInstanceGeometry'
 import {eachBatch} from '../viewer/ifc/batchedModel'
 import {glbVerbose} from './glbLog'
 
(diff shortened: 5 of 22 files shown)


pablo-mayrgundter:
@codex review

Review focus — the load-bearing claims to attack: (1) **reader completeness** — four `instanceGeometry` consumers were found and re-pointed (`batchedSubset`, `batchedToMergedMesh`, `glbBatchedExport`, `ResidencyController`); is anything else reading vertex data, identity, or bounds off the retained table, including diagnostics and the display/residency paths? (2) **the range re-bake's index rebase** (`- vertexStart`) — correct for every attribute layout the batch can hold, and against ranges that MOVE when three compacts/reuses slots (the `follows a shape whose range moves under it` test's claim)? (3) **identity semantics** — `glbBatchedExport`'s accessor dedup used the geometry *object* as its key; the replacement keys on `instanceGeometryIds` — is that identity stable across the same source shape appearing in both opaque and transparent batches? (4) **the second-owner clear** — `finalize()` now clears the builder's `geometryCache`; does anything (degraded rebuild, late diagnostics, GLB export ordering) legitimately read it after finalize? (5) **public-accessor reliance** — `getGeometryIdAt`/`getGeometryRangeAt` per r0.184; any behavioral gap vs the old table (e.g. reserved vs used counts in ranges)?

---

PR #1829, opened 2026-08-31
Loader: evict truncated part-21 sources and empty GLB cache artifacts

pablo-mayrgundter:
## Summary

GitHub OPFS HIT keys off remote sha + file existence, not local completeness. Two sticky poison cases:

1. **Truncated part-21 source.** A write killed after `truncate(0)` leaves a prefix under the correct sha. Conway finalizes that prefix as a complete index and every forward ref into the missing tail is \`not in the index\` (Arty \`STYLED_ITEM #36800\` → \`MANIFOLD_SOLID_BREP #1031384\`). Evict 0-byte files and part-21 files whose last 256 bytes lack \`END-ISO-10303-21\`, then re-fetch. GitHub-only — same shape as the LFS-pointer guard — because a re-fetch can repair it. Uploads are left alone (that copy is the only copy).

2. **Empty GLB artifact.** A failed extract can still return COMPLETE with 0 meshes, and the writer used to cache that scene. The next load is then a HIT of nothing. The writer now refuses to cache a scene with no POSITION vertices; the reader peeks the container JSON and evicts any artifact with no positioned primitives.

Existing empty/truncated OPFS junk is dropped on the next load of that URL.

Not related to conway#704 type-inherited materials (IFC-only). Arty is AP214.

## Test plan

- [x] \`looksLikeTruncatedPart21\` / Blob helper: complete file, prefix-only Arty-shaped DATA, non-part-21, 0-byte
- [x] \`sceneHasRenderableGeometry\` / \`cachedGlbHasRenderableGeometry\`: empty meshes, no POSITION, positioned primitive
- [x] Writer skips a 0-vertex traversed scene
- [x] Cover: truncated GitHub OPFS hit is evicted and re-fetched
- [x] Husky pre-commit (eslint + typecheck + 2879 Jest tests)
- [ ] Reload Arty from GitHub on a tab that still has the poisoned OPFS entries; expect a miss, a re-parse, and a render

@codex review

### DIFF
--- a/src/loader/Loader.cover.test.js
+++ b/src/loader/Loader.cover.test.js
@@ -557,6 +557,25 @@ describe('load() error/edge paths with OPFS enabled', () => {
   })
 
 
+  it('does not evict a GitHub OBJ whose header mentions ISO-10303-21', async () => {
+    const obj = new MockFile('# exported from ISO-10303-21 STEP\nv 0 0 0\n')
+    dereferenceAndProxyDownloadContents.mockResolvedValue([
+      'https://raw.githubusercontent.com/owner/repo/main/model.obj',
+      'abc123sha',
+      true,
+      false,
+    ])
+    doesFileExistInOPFS.mockResolvedValue(true)
+    downloadModel.mockResolvedValue(obj)
+
+    await load('https://github.com/owner/repo/blob/main/model.obj',
+      viewer, onProgress, true, setOpfsFile, '')
+
+    expect(deleteFileFromOPFS).not.toHaveBeenCalled()
+    expect(axios.get).not.toHaveBeenCalled()
+  })
+
+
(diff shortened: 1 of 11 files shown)


pablo-mayrgundter:
@codex review

pablo-mayrgundter:
@codex review — P1: health check now walks container chunks as subarray views (no second unpack/copy). P2: footer marker is END-ISO-10303-21;

pablo-mayrgundter:
@codex review — P2s: POSITION accessor count must be > 0; part-21 footer must be the trailing record (whitespace only after it).

pablo-mayrgundter:
@codex review — last P2: part-21 truncation sniff is gated on the IFC/STEP loader; 0-byte files of any type still evict. Stopping after this round.

PR #1845, opened 2026-09-14
share-140: key GLB hydration and picking off the artifact, not the cache (#1844)

### LINKED ISSUE #1844: share-140: an exported Bldrs GLB opened in Share renders but element hover, click and nav-tree selection do nothing
From the #1837 deploy-preview smoke: export `index.ifc` uncompressed with Bldrs metadata on, open the `.glb` in Share (`/share/v/new/<uuid>.glb`). The model renders and the nav tree shows the IFC hierarchy, but mouse-over doesn't highlight elements, clicking geometry selects nothing, and selecting a nav-tree row doesn't highlight the scene.

## Why

The export IS the batched-native cache artifact (`EXT_mesh_gpu_instancing` + `BLDRS_instance_tables` / `BLDRS_spatial_tree` / `BLDRS_element_properties`). When Share reads that artifact back on a **cache hit**, `Loader.js#load` runs the hydration that makes it a first-class model: `hydrateBatchedModelFromInstancedGlb` (InstancedMeshes → the decorated BatchedMesh with instance tables), then `restoreCacheHitPicking` (per-mesh `IfcInstanceMap`, BVH build), then `inferModelCapabilities` / `attachElementSubsets`. Every one of those is gated on `cameFromGlbCache`, i.e. on *where the bytes came from*, not on *what the file is*. A user-opened file takes the plain-GLB path: the `BLDRS_*` reader plugins still park their payloads on `userData` (which is why the nav tree appears — that hydration keys off the payload's presence), but the batched hydration and the picking restore never run, so there are no instance tables to pick against.

## Fix

Key the hydration and picking gates off the artifact itself — "this GLB carries `BLDRS_instance_tables` (batched) or per-vertex `_EXPRESSID` (merged)" — with `cameFromGlbCache` kept only where it is about the *source* (skipping the store path, the reader log lines). Concretely in `Loader.js`: the `hydrateBatchedModelFromInstancedGlb` condition (line ~717), `restoreCacheHitPicking(model, …)` (line ~814) and the BVH gate inside it (line ~1117). Fail-soft as today: a GLB without the payloads stays a plain GLB.

A `.glb` from a *different* Share build (older schema) is the case to think about: the payload readers already version-check, so a mismatch should degrade to plain, never throw.

## Tests

- Jest: `Loader.restoreCacheHitPicking.test.js` / `glbBatchedRoundTrip.test.js` — a batched artifact loaded with `cameFromGlbCache=false` still hydrates and gets `instancePicking` / `createSubset`.
- `describeMobileAndDesktop` E2E: export `index.ifc` (None), open the download through the Open dialog's file chooser, wait for the model, click a nav-tree element → the URL gains the element path and the scene highlights it.

Follow-up to #1837 (which now also makes compressed exports readable — the decoders were flag-gated on the read side). Depends on nothing; #1843 (portable export) should re-check this path for the de-instanced shape.

### LINKED ISSUE #1846: a merged GLB artifact whose scene root is a bare mesh node loses its per-vertex expressID
Found while building a merged-layout fixture for #1844 (PR bldrs-ai/Share#1845). Pre-existing, not introduced there, and not reachable from the default batched layout — filing so it isn't rediscovered a third time.

## What happens

When a GLB's scene root has no geometry of its own, `readModel` hoists `children[0].geometry` onto the root. `convertToShareModel`'s decoration walk then visits **that same geometry object twice** — once as the root, once as the child:

1. the root visit promotes `_expressid` → `expressID` and **deletes `_expressid`**;
2. the child visit finds no `_expressid`, concludes the mesh has no element ids, and **overwrites `attributes.expressID` with the synthetic single-element `Int8Array(1)` placeholder**.

Net: a merged artifact that genuinely carries per-vertex element ids comes back with a placeholder, so `inferModelCapabilities` never flips `expressIdPicking` and the model is unpickable.

## Why it hasn't bitten

- The default layout is **batched** (`glbBatched` default-on), which has no per-vertex ids at all — it picks through `BLDRS_instance_tables`.
- The real merged writer emits a Group-rooted scene, so the root has a container node above the mesh and the hoist never fires.

It is reachable by a hand-assembled or third-party GLB with a single mesh at the scene root, and by any future writer that emits that shape.

## Fix sketch

Make the decoration walk idempotent per geometry — track visited `geometry.uuid`s, or don't delete `_expressid` until the walk completes, or skip the synthetic placeholder when the geometry already has a real `expressID` attribute with `count > 1`. The third is the smallest and matches the `count > 1` convention `isBldrsGlbArtifact` and `inferModelCapabilities` already use.

## Test

A merged GLB whose scene root is a bare mesh node (no container) round-trips through `load()` with `expressID.count === vertexCount`, not `1`. `src/loader/glbArtifact.fixture.js#mergedGlbBytes` (added in bldrs-ai/Share#1845) builds the container-node shape deliberately to avoid this; the regression test is the same fixture **without** the container.

pablo-mayrgundter:
Fixes #1844. Fixes #1846.

Two bugs on one path: the Export tab hands the user the batched-native cache artifact, and reopening that `.glb` through the Open dialog produced a model nothing could pick.

## 1. Hydration and picking were keyed on the byte source (#1844)

The `BLDRS_*` reader plugins park their payloads on `userData` whatever the source (so the NavTree appeared), while `hydrateBatchedModelFromInstancedGlb` and the BVH build were gated on `cameFromGlbCache` — on where the **bytes** came from, not on what the **file** is.

`isBldrsGlbArtifact(model)` (new, exported) recognises the three signatures this build's writer emits: `BLDRS_instance_tables` (batched), `BLDRS_face_ids` (merged per-triangle), or per-vertex `expressID` with `count > 1`. The `> 1` matches `ShareModel.js:233`, and is provably right rather than merely conventional: `convertToShareModel` gates promotion on `preserved.count > 1` (`:1508`) and overwrites anything smaller, so a count-1 `expressID` at predicate time *is* the synthetic placeholder.

- **Hydration gate** → `isGlbBatchedActive() && model.userData.bldrsInstanceTables`. Only `BldrsInstanceTablesReader` — a GLTFLoader plugin — ever writes that key, so its presence already proves a GLTFLoader parse and no scoping term is needed.
- **Picking/BVH gate** → `isGlbParse && isBldrsGlbArtifact(model)`. Here the scoping term *is* needed and the asymmetry is the interesting part: per-vertex `expressID` is exactly what a live Conway-direct parse emits, so without `isGlbParse` this would double-build every IFC's BVH.

Fail-soft by construction, not by a new guard: a payload whose schema version the reader rejects is never parked on `userData`, so an older build's artifact reads as "not an artifact" and degrades to plain.

### Per-site `cameFromGlbCache` audit

14 occurrences at base, all in `Loader.js`; 11 at head (8 code + 3 comment).

| Line | What it does | Verdict |
|---|---|---|
| 302, 379, 517 | declaration; set true on a GitHub / non-GitHub cache hit | source — keep |
| 473 | comment on the `canStreamOpen` guard | source — keep |
| 537 | guards the conway store-open path so a hit never part-21-streams the artifact | source — keep |
| 587, 618 | reset on stale-LFS-pointer eviction and on OPFS failure | source — keep |
| **717** | `hydrateBatchedModelFromInstancedGlb` gate | **artifact — re-keyed** |
| 749 | `reader: parsed GLB OK: …` + 0-mesh / 0-vertex warnings | source — **keep**: this block is a *cost*, not a capability. `glbInfo` is ungated (unlike `glbVerbose`), and `summarizeGlbScene` walks the scene computing a per-mesh `Box3`. |
| **814**, **1117** | `restoreCacheHitPicking(...)` and the BVH gate it feeds | **artifact — re-keyed** |
| 910/913 | jsdoc + param name | renamed → `isGlbArtifact` |

`inferModelCapabilities` (`:766`) and `attachElementSubsets`, which #1844 names as part of the broken chain, were **never gated** — nor is the instance-map half of `restoreCacheHitPicking`, which keys on `capabilities.instancePicking || userData.bldrsFaceIds`. And the issue's "call at ~814" and "BVH gate at ~1117" are one decision point, not two.

### Cache-hit path unchanged

The dropped conjunct was `true` on every hit: both `cameFromGlbCache = true` sites are immediately preceded by `swapToGlbLoader`, and `loader.type` is written in exactly two places in the file (`:2065`, `:2282-2283`), neither of which runs again afterwards. Review hunted for a counterexample and found one class — a batched artifact whose `INSTANCE_TABLES_VERSION` the reader rejects while `BLDRS_GLB_SCHEMA_VERSION` still matches would now skip the BVH. Perf-only, and unreachable while the OPFS slot is partitioned by `schemaVer`. Noted in the jsdoc.

## 2. A user-opened *merged* export lost its per-vertex expressID (#1846)

Found in review. **An earlier revision of this description claimed the merged case was "a hover-latency problem, not a correctness one" — that was wrong**, and fixing it is why this PR grew a second commit.

1. The merged writer exports a **bare `Mesh`** (`batchedToMergedMesh.js:247,259`).
2. `GLTFExporter.parse` on a non-`Scene` wraps it in an `AuxScene` and pushes the mesh directly (`GLTFExporter.js:2645-2656`) — **no container node**.
3. Export hands the user chunk 0 verbatim, a bare GLB.
4. On reopen `readModel`'s hoist fires (`:1892-1908`), and `recursiveDecorate` visits that one geometry object twice: the root visit promotes `_expressid` → `expressID` and deletes `_expressid`; the child visit finds none and **overwrites `expressID` with the synthetic `Int8Array(1)`**.
5. `perVertexTrusted` is true for a bare GLB, so the face_ids order cross-check reads against a count-1 attribute, fails, warns, and nulls the map; the fallback then builds it from the clobbered ids.

Reachable whenever the batched export declines (shear matrices / missing attributes — the documented fail-soft at `glbExport.js:278-286`) or under `?feature=disableGlbBatched`. Cache-hit is immune: `parseBldrsGlbContainer` wraps chunks in a Group of Scenes, so the hoist never fires.

**Fix** (`Loader.js:1538`): one `else if` arm — a geometry already carrying `expressID` with `count > 1` sets `hasPerVertex` and is left alone rather than falling to the placeholder branch. Stated as the rule *real per-vertex identity outranks the placeholder*, so it is order-independent; either visit may be the promoting one. Rejected alternatives, with reasons in the commit: *don't hoist* (`model.geometry` has consumers beyond this walk), and *a visited-geometry `Set`* (needs `hasPerVertex` propagation, and changes behaviour for two meshes legitimately sharing one geometry).

### Blast radius — the part that mattered

This arm sits in `recursiveDecorate`, which is on the load path for every non-IFC model. Two independent checks:

**Enumeration.** `convertToShareModel` runs only on the `!isIfc` branch (`:755`), and `isIfc` covers `ifc`/`step`/`stp` (`:1984`) — live IFC and STEP parses never reach the walk at all. Every producer of an `expressID` `BufferAttribute` in `src/` is one of three non-test sites (`flatMeshToBufferGeometry.js:330`, `batchedSubset.js:167`, `batchedToMergedMesh.js:240`); none feeds this walk. So the only way the arm's condition is true at entry is the promotion earlier in the same walk.

**Instrumentation.** Replacing the arm's body with a `throw` and running the full suite:

```
F1-ARM-ENTERED name=Model type=Mesh depth=1
● … › restores per-mesh instance maps and BVHs on the merged layout
● … › keeps per-vertex ids on a Mesh-rooted merged artifact with no face_ids (#1846)
Tests: 2 failed, 5 skipped, 3171 passed, 3178 total
```

Exactly the two merged-artifact tests enter it, at `depth=1`. The other 3176 do not — including the cache-hit GLB suites, the batched user-opened artifact, conway-direct IFC and STEP, and OBJ/STL/USD/FBX/splat.

## Verification

| | base `b89f1ff` | head `0ae37df` |
|---|---|---|
| tools Jest | 8 suites / 27 tests | unchanged |
| src Jest | 289 suites, 3163 passed + 5 skipped, 13 snapshots | **290** suites, **3173** passed + 5 skipped, 13 snapshots |
| eslint `src netlify tools --max-warnings 0` | clean | clean |
| `yarn typecheck` | clean | clean |
| `yarn test-ci` (coverage) | — | 73.59 stmts / 68.99 br / 74.89 fn / 73.03 lines, thresholds passed |
| `yarn build-prod` | — | succeeds |
| `yarn test-flows …/exportGlb.spec.ts` | 10 tests | **12** tests, all pass (2.9 min) |

`build-prod`, `test-ci` and Playwright were run locally — the three things the husky hook doesn't cover. `package.json` restored with a standalone `git checkout --` after each build; no version stamp in any commit.

### Tests pin the change

**#1844**, against a revert of `Loader.js`:

```
● hydrates the batched-native layout to a pickable BatchedMesh
  > 138 | expect(model.isBatchedMesh).toBe(true)          Expected: true  Received: undefined
● restores per-mesh instance maps and BVHs on the merged layout
  > 174 | expect(mesh.geometry.boundsTree).toBeDefined()  Received: undefined
● leaves a GLB with no Bldrs payload plain, and does not throw
  TypeError: (0 , _Loader.isBldrsGlbArtifact) is not a function
Tests: 3 failed, 1 passed, 4 total
```

Plus 5 red in `Loader.restoreCacheHitPicking.test.js`. What *passed* pre-fix is the point: `userData.bldrsInstanceTables` was truthy and the merged model's `instanceMap` was attached — the "renders, tree is there, nothing selectable" symptom.

E2E, rebuilt from the reverted tree — every earlier assertion passed (download, reopen, full `Bldrs → Build → Every → Thing → Together` tree, row highlight, URL element path); only the **scene** highlight was zero:

```
1) [chromium] [desktop] › an exported .glb reopens as a pickable model, not just geometry
2) [chromium] [mobile]  › (same)
   > 440 | expect(await sceneHighlightCount(page)).toBeGreaterThan(0)   Expected: > 0   Received: 0
```

**#1846**, with the fix removed:

```
● … › restores per-mesh instance maps and BVHs on the merged layout
● … › keeps per-vertex ids on a Mesh-rooted merged artifact with no face_ids (#1846)
    Expected: 6   Received: 1
```

Ordering of evidence: the fixture was corrected to the real Mesh-rooted shape **first**, and the pre-existing merged test went red on it before `Loader.js` was touched — the failure path reproduced end to end through the real `load()`.

**One test cannot fail pre-fix, and this PR says so rather than implying otherwise:** *"degrades an artifact from a different Share build to plain, not to an error"* passes both before and after. The gate never fired for a user-opened file, so a version mismatch had nothing to break. It is a blast-radius guard the issue asked for, not a pin.

## Review

Codex reviewed `419e03a` clean without engaging any named claim. Per [`agent-workflow.md`](https://github.com/bldrs-ai/Share/blob/main/design/new/agent-workflow.md) §Review, a substituting sub-agent round ran against six specific claims and found #1846 plus four accuracy defects — all fixed here. Round summary: [#issuecomment-5668907011](https://github.com/bldrs-ai/Share/pull/1845#issuecomment-5668907011). Claims 2, 3, 5 and 6 came back sound with the reasoning matching the code.

Also fixed from that round: the merged fixture modelled a shape the writer never emits (container node, no `BLDRS_face_ids`) while its header claimed to prevent exactly that drift — it now emits the real Mesh-rooted shape and injects face_ids via the writer's own `capturePerTriangleIds` → `buildFaceIdsExtensionData` pair, with a `withFaceIds: false` option for the pre-face_ids artifact; the `isGlbParse` comment said "two gates" where one consumes it; two stale `batchedArtifact.fixture` references; a leftover `pageerror` debug listener.

## Known gaps

- **#1847** (user-opened *compressed* merged export trusts per-vertex ids the codec may have scrambled) is filed and not fixed here — pre-existing, and that block already ran ungated before this change.
- The merged path is not exercised against three's real `GLTFExporter`: it throws under jsdom (`FileReader.readAsArrayBuffer` on a Blob), the documented reason the fixture hand-assembles bytes. The Mesh-rooted claim rests on reading `GLTFExporter.js`'s `AuxScene` wrap plus `batchedToMergedMesh.js:247,259`. The Playwright reopen-and-pick test does exercise a real browser round trip.
- `withFaceIds: false` covers the pre-face_ids artifact by construction; not verified against a real archived artifact from an older build (none in-repo).
- Unrelated, pre-existing, not fixed: `yarn test-flows-build-and-serve` bare is broken on `main` too — `test-flows-serve` is `http-server docs -c-1 -p` with no port value, which current Node rejects. The port arrives positionally from `tools/playwright.config.js`'s `webServer`, so `yarn test-flows …` works.

https://claude.ai/code/session_01WxRzBHBAx9zt7ckRuqubNd


pablo-mayrgundter:
@codex review

Hazards worth aiming at, since the risky part of this change is an invariant the diff doesn't state:

1. **"The cache-hit path is unchanged."** This is argued by construction in the description, not asserted by any test. The claim is that on a hit, `isGlbParse && isBldrsGlbArtifact(model)` is always `true` wherever the old literal `true` was, and that dropping `cameFromGlbCache` from the hydration conjunction is a no-op because the other conjuncts already imply it. Attack the edge: is there a real cache-hit artifact carrying *none* of the three signatures where behaviour now differs?

2. **`isGlbParse` as the live-IFC-parse guard.** `isBldrsGlbArtifact` returns true for a live IFC parse too (its meshes carry per-vertex `expressID`), so `loader.type === 'glb' || 'gltf'` is the only thing keeping the BVH traversal off live models. Is `loader.type` reliably set by then on every path reaching line 814 — including the fallback paths at 587 and 618, and a `.bld` child load?

3. **`count > 1` vs the synthetic placeholder.** `convertToShareModel` stamps `Int8Array(1)` on models with no ids. Is `> 1` right for every layout, or is there a legitimate single-element artifact it now misclassifies as plain?

4. **The deliberate non-change at line 734.** The `reader: parsed GLB OK` log stays keyed on `cameFromGlbCache`, argued as a cache-pipeline diagnostic rather than a model capability. Is that the right split, or does a user-opened artifact now silently skip diagnostics someone will want?

5. **Ordering.** The predicate is evaluated at the call site, before `restoreCacheHitPicking`'s mid-function `delete model.userData.bldrsFaceIds`. Checked, but worth a second read.

---

pablo-mayrgundter:
## Review round 1 — agent-reviewed, and why

Codex ran on `419e03a` and replied "Didn't find any major issues" without engaging any of the five claims named above. Per [`design/new/agent-workflow.md`](https://github.com/bldrs-ai/Share/blob/main/design/new/agent-workflow.md) §Review — *"a clean review that never engaged with the load-bearing claim has not reviewed it"*, and *"a single clean pass is not by itself a merge signal"* — I ran a substituting sub-agent review as this round, pointed at the six specific claims rather than the diff generally. Recording that here so a later reader knows which reviewer signed off.

It found one blocking defect codex missed, plus prose that the code contradicts.

### Sound, with the PR's reasoning matching the code

- **`isGlbParse` as the live-parse guard.** `loader.type` is written in exactly two places — `findLoader`'s `loader.type = extension` (`:2065`) and `swapToGlbLoader` (`:2282-2283`) — and both `cameFromGlbCache = true` sites are immediately preceded by the swap, so the conjunct is true on every hit. `.bld` children resolve their own loader per child (`BLDLoader.js:53`, sequential `await`, no interleaving); there is no swap-then-fallback path (a failed `readModel` throws out of `load()`).
- **`count > 1`.** Stronger than the PR argues: `convertToShareModel` itself gates promotion on `preserved.count > 1` (`:1508`) and overwrites anything smaller, so by the time the predicate runs a count-1 `expressID` is *provably* the placeholder. `ShareModel.js:233` confirmed to use the identical convention.
- **Ordering.** No intervening statement at all between argument evaluation and the callee's `delete`. One production call site, once per load, O(nodes).
- **The `traverse` guard.** Unreachable, so a speculative defense by rubric #5 — but it mirrors the identical guard in `inferModelCapabilities` (`ShareModel.js:202-204`), which is the function this one is keyed to. Consistency wins; `false` is also the safe answer.
- **Cache-hit path unchanged.** The reviewer hunted for a counterexample and found one class: a batched artifact whose `INSTANCE_TABLES_VERSION` the reader rejects while `BLDRS_GLB_SCHEMA_VERSION` still matches would now skip the BVH. Currently unreachable — the OPFS slot is partitioned by `schemaVer` — and perf-only if it became reachable. Getting a jsdoc note.
- **Numbers verified independently:** 290 suites / 3172 passed + 5 skipped / 13 snapshots, exit 0; eslint and typecheck clean; tree clean afterwards (no stray `.js`, no version stamp).
- **`glbBatchedRoundTrip.test.js` "assertions unchanged" — verified true.** The one assertion that left the file (`expect(written).not.toBeNull()`) is preserved as a throw inside `batchedArtifactBytes`; same failure, same trigger. No coverage lost.

### Blocking: the merged layout is still broken, and this PR says otherwise

The description's "corrections to the issue as filed" claims a user-opened **merged** artifact already got its `IfcInstanceMap` and only its BVH was missing — *"a hover-latency problem, not a correctness one"*. The code contradicts that:

1. The merged writer exports a **bare `Mesh`** (`batchedToMergedMesh.js:247,259`).
2. `GLTFExporter.parse` on a non-`Scene` wraps it in an `AuxScene` and pushes the mesh directly (`GLTFExporter.js:2645-2656`) — **no container node**.
3. Export hands the user chunk 0 verbatim, a bare GLB.
4. On reopen `readModel`'s hoist fires (`:1892-1908`), and `recursiveDecorate` visits that one geometry object twice — root visit promotes `_expressid` → `expressID` and deletes `_expressid` (`:1505-1512`); child visit finds none and **overwrites `expressID` with the synthetic `Int8Array(1)`** (`:1514-1519`).
5. `perVertexTrusted` is true for a bare GLB, so the face_ids cross-check reads against a count-1 attribute, fails, and nulls the map; the fallback then builds it from the clobbered ids.

Reachable whenever the batched export declines (shear matrices / missing attributes, the documented fail-soft at `glbExport.js:278-286`) or under `?feature=disableGlbBatched`. Cache-hit is immune because `parseBldrsGlbContainer` wraps chunks in a Group of Scenes, so the hoist never fires.

That is #1844's symptom class, unfixed — so `Fixes #1844` would auto-close a live bug. **Fixing it here** rather than narrowing the claim, since #1844's title is exactly this symptom. #1846 (filed earlier from the same mechanism) will close with it.

Also being fixed: the merged fixture models a shape the writer never emits (a container node, and no `BLDRS_face_ids`) while its header claims to prevent exactly that drift — it was the container that made the merged test pass; the `isGlbParse` comment says "two gates" where only one consumes it; two stale `batchedArtifact.fixture` references.

One correction to my own PR body: the line-749 rationale claims re-keying that log "would break console-hygiene assertions". Unsupported — `installGlbLogCapture()` diverts the channel under Jest and every `getGlbLogs()` assertion is `some(...)`/`toContain(...)`, never absence or exact-set. The decision to keep that log cache-only stands on the other two grounds (`glbInfo` is ungated, and `summarizeGlbScene` costs a full traverse with a per-mesh `Box3`), so the code doesn't change — the prose does. Worth noting the same PR's widened hydration gate *does* newly emit one `[glb]` line for cache-flag-off users, which cuts against the spam argument.

Staying in draft until this lands.

---

pablo-mayrgundter:
@codex review

Round 2. The diff changed materially since your clean pass on `419e03a` — head is now `0ae37df`, and the description has been rewritten to match. A sub-agent round found a second bug (#1846) that the first revision of this PR had explicitly, and wrongly, claimed was not a correctness problem. Please re-read rather than diffing against your previous pass.

The new commit adds one `else if` arm to `convertToShareModel`'s `recursiveDecorate` (`Loader.js:1538`): a geometry already carrying `expressID` with `count > 1` is left alone instead of falling through to the synthetic `Int8Array(1)` placeholder.

Claims to attack, in priority order:

1. **Blast radius of the new arm.** `recursiveDecorate` runs on every non-IFC model load. The argument that this is safe rests on two things: (a) `convertToShareModel` runs only on the `!isIfc` branch (`:755`) so live IFC/STEP never reach it, and (b) instrumenting the arm with a `throw` showed exactly 2 of 3178 tests entering it. Is the enumeration complete — is there a producer of an `expressID` `BufferAttribute` outside the three named non-test sites (`flatMeshToBufferGeometry.js:330`, `batchedSubset.js:167`, `batchedToMergedMesh.js:240`)? And is there a *runtime* path the test suite doesn't cover where a geometry legitimately arrives at the walk with a real `expressID` already set?

2. **Two meshes sharing one geometry.** The fix was chosen over a visited-geometry `Set` partly because the `Set` "changes behaviour for two meshes legitimately sharing one geometry". Does the shipped arm handle that case correctly, or does it merely differ from the rejected alternative?

3. **`count > 1` as the placeholder discriminator**, now load-bearing in two places rather than one. `convertToShareModel:1508` gates promotion on the same threshold. Is there any layout where a real single-vertex-id geometry exists at this point?

4. **The fixture rewrite.** `mergedGlbBytes` now emits a Mesh-rooted scene and injects `BLDRS_face_ids` through the writer's own `capturePerTriangleIds` → `buildFaceIdsExtensionData` pair. Does it now faithfully model what the merged writer emits, or has it traded one synthetic shape for another? The claim that the merged writer is Mesh-rooted rests on reading `GLTFExporter.js`'s `AuxScene` wrap plus `batchedToMergedMesh.js:247,259` — not on an executed export, because `GLTFExporter` throws under jsdom.

5. **Still-open from round 1, unchanged:** that the cache-hit path is unchanged (argued by construction, asserted by no test), and that `isGlbParse` alone keeps live parses out of the BVH traversal.

Note one test is disclosed as unable to fail pre-fix (the older-schema degradation guard). If you find any *other* test that cannot fail, that is a finding.

---

